PR #13, opened 2021-10-17
Create brush_filechooser-help.it.txt

albanobattistella:


flxzt:
Sorry about the wasted effort, but the lastest update removed that helper text and stroke templates in general in order to simplify some things. Custom strokes will be integrated differently in the future, hopefully more intuitively.

PR #35, opened 2022-01-15
build: complete translations support

oscfdezdz:


flxzt:
Thanks for the PR. I think we can exclude the gschema. And I think the merge_file() methods are missing for the desktop and metainfo files.
I also would like to have the keys in simplified english and have an extra english translation or at least revisit and improve the keys a little bit, so I might regenerate the .pot after the merge (and add translator comments in a few places). Thats not a problem for your spanish translation right?

oscfdezdz:
> I think we can exclude the gschema. And I think the merge_file() methods are missing for the desktop and metainfo files.

It should be like this now.

> I also would like to have the keys in simplified english and have an extra english translation or at least revisit and improve the keys a little bit, so I might regenerate the .pot after the merge (and add translator comments in a few places). Thats not a problem for your spanish translation right?

No, it is not a problem, I will leave #36 open and when the changes are done in main, I will rebase it.

flxzt:
Ok done, I hope it works. First time integrating gettext.You can rebase against the new .pot

PR #36, opened 2022-01-15
po: add Spanish translation

oscfdezdz:
Depends on https://github.com/flxzt/rnote/pull/35.
Marking as a draft as it will need rebasing if that one gets merged.


flxzt:
getting parsing errors and a panic, I think the spanish translation is not valid UTF-8?

oscfdezdz:
It is probably because when generating the pot file, the encoding is set to `CHARSET`. I have been testing `bind_textdomain_codeset` function without getting it to work as it should.

flxzt:
Awesome. Thanks for the effort for the added translation support!

PR #39, opened 2022-01-17
Add pdf keyword and Office category

nbenitez:
Rnote does import and export to PDF so let's add the 'pdf' keyword so when people search for 'pdf' in flathub.org then rnote is among the results.
We also add the Office category which shows up in flathub.org as 'Productivity'.


flxzt (approved):
good idea!

Issue #37, opened 2022-01-15
Retire Lena


piegamesde:
The example screenshots in the README and in other places include an image of [Lena Forsén](https://en.wikipedia.org/wiki/Lena_Fors%C3%A9n). Lena has long retired, and [does not wish the image to be used anymore](https://pursuit.unimelb.edu.au/articles/it-s-time-to-retire-lena-from-computer-science). Please respect this and update the screenshots.

flxzt:
I was totally unaware. Of course, do you have a suggestion for a replacement?

piegamesde:
Um, I don't know ^^ I only know the standard test 3D models FWIW (Stanford bunny, Blender's Suzanne, Utah teapot, …).

I've done a bit of digging on other software projects and there appears to be no canonical replacement. The "best" I've found is https://sipi.usc.edu/database/database.php?volume=misc&image=13, of which most images should do. "Baboon" and "Peppers" appear to be the most popular. Fun fact: there's even [a paper](https://doi.org/10.1080/09500340.2016.1270881) on that subject (which, ironically, does not contain any images at all).

Since this is just an example for "this software can include raster images" and we don't have any technical requirements for testing algorithms (color features etc.), I honestly don't care. Any royalty-free image will do just fine. Personally, I'd go with "Baboon", 'cause why not:

![image](https://user-images.githubusercontent.com/14054505/149626249-13961a36-0607-4c02-b7e2-0d7f971bf829.png)

flxzt:
Lets do the mandrill, he has a magnificent stare. Do you want to take some screenshots to showcase the different features?

flxzt:
Lets use this version ![Mandrill](https://user-images.githubusercontent.com/19841886/149626679-148cf4a6-80f5-4972-ba31-842028197fda.png)

piegamesde:
Do you still have the save file used for those screenshot lying around? Otherwise, would you mind if I simply photoshop the new images in? :D

flxzt:
Ah, no I want new screenshots anyways. But I can do them myself then. They can only be updated on the store page with a new release though.

piegamesde:
For one of the desktop applications that I've written, I automated the screenshot generation so that they would always keep up with UI changes without too much manual effort. I don't know how feasible this is with Gtk, but I really recommend to automate as much of the process as possible. It's definitely been worth the initial setup in my case.

flxzt:
updated with v0.3.0!

PR #51, opened 2022-01-21
Replace Len(n)a with June the odd-eyed cat

miker2049:
June was found here:
https://commons.wikimedia.org/wiki/File:June_odd-eyed-cat.jpg

Sorry didn't initially see #37


flxzt:
will do for next release

PR #52, opened 2022-01-21
Ukrainian translation for rnote is added.

vlisivka:


flxzt:
Thanks! You can add the lang the LINGUAS

PR #59, opened 2022-01-23
Create it.po

albanobattistella:


flxzt:
there are two PRs from you, this one and #58 .  which one should I pick?

albanobattistella:
> there are two PRs from you, this one and #58 . which one should I pick?

HI,
are the same.
Thanks!
Albano

PR #72, opened 2022-01-30
i18n: update es.po

oscfdezdz:


flxzt:
I want the changelog to be left out and deleted the strings in the pot. Could you  sync against that? Thanks!

Issue #117, opened 2022-02-18
Allow opening multiple rnote windows


Mat931:
**Is your feature request related to a problem? Please describe.**
Having multiple windows open at the same time could be useful if you want to work on two documents side by side

**Describe the solution you'd like**
If a new-window desktop action is present in the .desktop file, you can open a new window by middle clicking the app icon in the Ubuntu dock or by using the context menu

PR #118, opened 2022-02-18
Add new-window desktop action

Mat931:
https://github.com/flxzt/rnote/issues/117


flxzt:
Looks like the color scheme is always resetted when opening a new window. Not sure what happens there.

PR #138, opened 2022-03-21
Added Hebrew translation

yarons:


yarons:
I would also like to suggest using a translation platform such as Weblate which should make the pushing and pulling redundant.

Thank you.

flxzt:
Hi, thanks for the translation!
And I agree. (it is actually the very first reported issue, #1).

flxzt:
@yarons I finally applied for the hosted weblate service, you can find the weblate repo here: https://hosted.weblate.org/projects/rnote/repo/

yarons:
Thanks @flxzt , you are the best!

PR #140, opened 2022-03-24
Upgrade old rnote files so they can be opened in the newest version

FriederHannenheim:
Opening this pull request to get Feedback. I will extend it to upgrade versions older than 0.3. For now I need to know if this is up to code style and if you have any suggestions.

### DIFF
--- a/rnote-engine/src/sheet/mod.rs
+++ b/rnote-engine/src/sheet/mod.rs
@@ -21,7 +21,7 @@ use self::{background::Background, format::Format};
 use gtk4::{glib, Snapshot};
 use p2d::bounding_volume::{BoundingVolume, AABB};
 use serde::{Deserialize, Serialize};
-use serde_json::json;
+use serde_json::{json, to_string_pretty};
 
 #[derive(Debug, Serialize, Deserialize)]
 #[serde(default, rename = "sheet")]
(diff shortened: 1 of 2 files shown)


flxzt:
Thanks for addressing this!
My original plan was to move the code dealing with the file format to the rnote-fileformats subcrate, and have something similar to the XoppFile struct. This would decouple the file format from the internal representation, and we could optimize for different things like the file size ( e.g. I don't think storing the slotmap with its different components as it is is very efficient). And when we have file format breaks, we could match on the version number (like you did) and implement TryFrom from the old into the new type (and chaining up those conversion so we avoid having to refactor it everytime). TryFrom could also be implemented to have conversions between .rnote <-> .xopp (and possibly more formats in the future). The subcrate could then also serve as a library for other drawing apps eventually.
The downside is that there would be quite a bit of duplicate code, but I don't think this is entirely avoidable if we want to stabilize the format.

Also: There will be quite a bit of changes for the next version, (the WIP branch `compose`). I haven't touched the store holding the strokes yet, but it definitely have some changes (possibly introducing a quadtree on top of the slotmap). I am not sure how compatible it will be to older versions. We might want to have that as a base version for compatiblilty for the future.

flxzt:
It seems I have failed to insert the correct versions in the file format so far, so matching against that for older versions won't work anyways :/ Either we convert on a best effort basis, or we start from v0.5.0 on for backwards compatibility. Since so much will change in the next release, I think we should split for the last time and from then on, make some effort to support older versions. What do you think?

FriederHannenheim:
I think that would be better too. The users have been warned and there's no need to add a lot of unnecessary code just to upgrade now.

wojpawlik:
> Either we convert on a best effort basis, or we start from v0.5.0 on for backwards compatibility.

Does this mean current files will be openable forever (not necessarily losslessly)?

I'd love to give RNote a shot to avoid Xournal++'s jank

flxzt:
@wojpawlik I will try to always provide an upgrade path, but I can't promise it. If you are planning to archive files for a longer time, I would suggest sticking to Xournal++'s format for now.

PR #163, opened 2022-05-07
Suggestions for menu strings

mbousq:
Hi Felix. 
These are just some suggestions to potentially clarify menus in the app.

**What I found**
- The use of "Sheet" everywhere. When the user has "show format option" active, the user is exposed to, in a way, multiple sheets. Therefore, options such as clear sheet, save sheet and so on could infer saving _only_ the active "sheet". In this merged request, I changed "sheet" to "document" where necessary, while keeping "sheet" for "Sheet format", "Zoom to fit sheet width" etc.

- "Expand mode" to "Layout"
- "Endless vertical" to "Continuous vertical"
- "Export .." to "Export…"

Tell me what you think.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,6 +2,13 @@
 Please ask in the [Github Discussions](https://github.com/flxzt/rnote/discussions) section if you need help with anything!
 
 # Building
+
+First clone the repository and init its submodules
+```
+git clone https://github.com/flxzt/rnote
+git submodule update --init --recursive
+```
+
 ## Building with Flatpak
 There is a flatpak manifest in `build-aux/com.github.flxzt.rnote.Devel.json`.
 
--- a/misc/building/rnote-windows-build.md
+++ b/misc/building/rnote-windows-build.md
@@ -1,0 +1,62 @@
+# Build instruction for Windows (experimental, WIP)
+
+### Prerequisites
+- install msys2 from here: https://www.msys2.org/
+- install Rust from here: https://www.rust-lang.org/
+
+in a msys2 terminal, install git and the dependencies:
+```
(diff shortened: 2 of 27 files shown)


flxzt:
Very good suggestions, I fully agree! Let me update the naming in the source too, and then the pot can be regenerated from it.

mbousq:
By the way, I forgot another thing.  We have "Brush" in the header bar and "Pen sounds" in the menu. Not sure which terminology to keep between pencil or brush, or if you're planing on adding a brush sound :p

flxzt:
Oh true. I guess "brush sounds" would be better

mbousq:
And last one, Dot Radii, I don't quit understand the two values. The first one seems to relate to stroke thinckness and the second one is the radius ?

mbousq:
ah Ok, got it. vertical and horizontal radius

flxzt:
since you didn't understand it right away, this probably should also be updated with a clearer description..

mbousq:
Maybe horizontal and vertical flattening ?

https://en.wikipedia.org/wiki/Flattening

flxzt:
Not sure if that is more clear without context. My thought was to a more detailed explanation in the subtext.

Just so you are not confused when the strings are updated: I'll keep `pen sounds` for now, that'll leave me the option to extend them to other pen types in the future.

flxzt:
What about "The radii of the individual dots in the textured stroke"

mbousq:
Good idea !

flxzt:
Ok great! Can you check the new generated .pot, if I missed something? I want to do a new fix release soon

mbousq:
All good, at least as far as I can see :+1:

flxzt:
Ok. Thanks!

Issue #17, opened 2021-11-19
Feature request: Import images from the clipboard


giannissc:
Currently the only way that you can add pictures to the notebooks is through the import option. It would be nice if you could use copy+paste to add pictures to it

flxzt:
is now implemented with [bc5db70](https://github.com/flxzt/rnote/commit/bc5db70218a08a2a040325b29c100a169936ea68).

PR #181, opened 2022-05-19
feature: clipboard paste from screenshot for png, jp(e)g, bmp and tiff mime types

RickStanley:
Extending https://github.com/flxzt/rnote/issues/17 and follow up on https://github.com/flxzt/rnote/discussions/177, I implemented paste from a direct screenshot save. Not sure if it's the way to go, but it works. The branching is getting a little too big though.


flxzt:
thanks! I see they added .save_to_png_bytes() as API to textures in 4.6, that makes things easy.

> Trying the approach like:
> 
> https://github.com/flxzt/rnote/blob/65ba51827fba1e8b36120b78e191d9fc0f9b95a0/rnote-ui/src/appwindow/appwindowactions.rs#L1256
> , I got "No compatible transfer format found", so figured it was some upstream limitation, and maybe it's going to be fixed in the upcoming release? https://gtk-rs.org/gtk4-rs/git/docs/gdk4/struct.Clipboard.html

I am not sure I understand you, can you elaborate?

flxzt:
Ok 😄 did you test "image/jpeg"? It might be possible that this works too

RickStanley:
I does work! Just tested it, can confirm for:
- image/jp(e)g;
- image/png;
- image/tiff;
- image/bmp.

Maybe these too: image/avif text/ico image/icon image/ico application/ico image/x-ico image/x-icon image/x-MS-bmp image/x-bmp, but I can't say for sure.

Anyway, I think that the "time has come" to refactor these if/else branches to a `match`. I was going to do this, but I was/am not comfortable enough to fiddle with Rust just yet.

RickStanley:
So, I took a crack at it, I came up with this:

Create a image mime type enum for supported enums:
```rust
enum ImageMimeType {
    BMP,
    JPEG,
    JPG,
    PATH,
    PNG,
    SVG,
    TIFF,
}

impl ImageMimeType {
    const VALUES: [&'static str; 7] = [
        Self::BMP.value(),
        Self::JPEG.value(),
        Self::JPG.value(),
        Self::PATH.value(),
        Self::PNG.value(),
        Self::SVG.value(),
        Self::TIFF.value(),
    ];

    const IMAGES_MIME_TYPE: [ImageMimeType; 7] = [
            ImageMimeType::BMP,
            ImageMimeType::JPEG,
            ImageMimeType::JPG,
            ImageMimeType::PATH,
            ImageMimeType::PNG,
            ImageMimeType::SVG,
            ImageMimeType::TIFF,
    ];

    const fn value(&self) -> &str {
        match *self {
            ImageMimeType::BMP => "image/bmp",
            ImageMimeType::JPEG => "image/jpeg",
            ImageMimeType::JPG => "image/jpg",
            ImageMimeType::PATH => "image/uri-list",
            ImageMimeType::PNG => "image/png",
            ImageMimeType::SVG => "image/svg+xml",
            ImageMimeType::TIFF => "image/tiff",
        }
    }

}
```

Get the clipboard content's `mime_types` and find the first that matches the array of supported mime types:
```rust
if let Some(mime) = content_formats.mime_types().iter().find(|&mime| ImageMimeType::VALUES.contains(&mime.as_str())) {
  // ...
}
```

Get equivalent enum to that mime type (which in turn is in `&GString`):
```rust
if let Some(mime_type_enum) = ImageMimeType::IMAGES_MIME_TYPE.iter().find(|&available_mime| mime == available_mime.value()) {
  // ...
}
```

Since we already know that we've got ourselves a mime type, spawn:
```rust
glib::MainContext::default().spawn_local(clone!(@strong appwindow => async move {
  // ...
}
```

Finally `match` against the mime type and execute the function accordingly:
```rust
match mime_type_enum {
      ImageMimeType::SVG => {
        // handle this
      }
      ImageMimeType::PATH => {
        // handle this
      }
      ImageMimeType::PNG |
      ImageMimeType::JPEG |
      ImageMimeType::JPG |
      ImageMimeType::BMP |
      ImageMimeType::TIFF => {
        // handle this
     }
};
```

flxzt:
If we do that, I would suggest using the [mime crate](https://docs.rs/mime/latest/mime/) and refactor all mime-types in the import, export, dialogs, etc. with it so that we can define supported types as a static array somewhere and match with it throughout the codebase. There are subtleties dealing with mime-types, and if we do make the effort, we should make it so that it is as robust as possible. (and as easy to maintain as possible as well)

But not sure if this is actually necessary? It is a lot of work and I don't currently see it benefiting the codebase that much tbh.
I think with this PR it is enough adding another `if else` for jpeg.

RickStanley:
Sure, I can add an "or" along with `image/png` for jp(e)e

flxzt:
Great! Just to be clear, I wouldn't at all be opposed to a separate PR implementing this, if you want to do it. But it definitely has a much bigger scope than this quick addition

flxzt:
is it ready? lgtm

RickStanley:
It is :+1:

PR #225, opened 2022-07-17
Update BUILDING.md

TeiVDK:
Added "cd rnote" because not everyone may know you need to be inside of the rnote folder to execute the second command. Also without it simply copying and pasting the lines results in errors.


flxzt:
could indeed be confusing indeed. Thanks!

TeiVDK:
No problem!

PR #227, opened 2022-07-27
Stroke layers

### LINKED ISSUE #106: Higlighter option
There should be a dedicated highlighter option that keeps the same opacity no matter how fast or slow you move your pen/mouse. Currently using translucent brush leaves weird circles that can be very close or very far from each other making line opacity very inconsistent and ugly.

image

### LINKED ISSUE #183: [Feature request]: A highlighter - ability to highlight under a handwriting
The ability to _highlight_ handwritten notes with a highlighter tool. i.e. a marker, which is wider, and the colours of which are lighter variants of the handwriting itself.

**Example(s):**

highlight_handwritten_notes

handwriting_examples

highlighting_example

_**Edit**_: This is essentially the same as #106.

flxzt:
This implements stroke layers.

There are several dedicated layers: documents (pdfs), images (bitmap / vector), highlighter and user layers. They get rendered in this order.
The marker brush style now saves into the highlighter layer which is drawn under the "regular" brush layer, making it possible to mark text without obstructing it. Fixes #106, fixes #183 when merged.

When selecting strokes, they are also no longer resurfaced to the top, but now stay in their layer.

The brush now saves the individual options for each style and the brush width switches and is remembered when changing between the styles.

### DIFF
--- a/rnote-engine/src/document/mod.rs
+++ b/rnote-engine/src/document/mod.rs
@@ -185,8 +185,7 @@ impl Document {
         let padding_horizontal = self.format.width * 2.0;
         let padding_vertical = self.format.height * 2.0;
 
-        let mut keys = store.stroke_keys_as_rendered();
-        keys.append(&mut store.selection_keys_as_rendered());
+        let keys = store.stroke_keys_as_rendered();
 
         let new_bounds = if let Some(new_bounds) = store.bounds_for_strokes(&keys) {
             new_bounds.extend_by(na::vector![padding_horizontal, padding_vertical])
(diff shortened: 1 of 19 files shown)


flxzt:
For highlighting there is the dedicated "marker" pen style, which always draws underneath regular handwriting

EDLLT:
> For highlighting there is the dedicated "marker" pen style, which always draws underneath regular handwriting

Yes, that is correct. Perhaps you didn't get what I meant

When drawing a line which is slightly transparent as a Marker behind a normal stroke, weird circles appear similar to #106  instead of it being a constant line

PR #235, opened 2022-08-01
Automate installing the build on Windows

iwoithe:
**Description**
This PR automates installing the build on Windows so that there is limited manual work to be done. This is needed to simplify the script which will create the Windows installer.

**Additional Notes**
Not sure on the best method of automating the renaming of ``rnote.exe``. This will need to be worked out at somepoint.


flxzt:
looks good, thanks!

flxzt:
regarding renaming the exe, maybe meson is able to handle it. I need to check

PR #249, opened 2022-09-17
update link to pizarra

categulario:
I'm moving domains, what a painful process .-.


flxzt:
thanks for thinking about that link :)

PR #308, opened 2022-10-19
adding arrow struct

TornaxO7:
Thlis PR adds the arrow.


TornaxO7:
@flxzt I need some help: Currently I'm trying to understand the [`Composer`](https://github.com/flxzt/rnote/blob/main/rnote-compose/src/style/mod.rs#L54) implementation of `Line`. Isn't that [here](https://github.com/flxzt/rnote/blob/main/rnote-compose/src/style/mod.rs#L57\) a recursive call?

TornaxO7:
@flxzt (ping)

flxzt:
No, the second call to `composed_bounds()` is receiving the options, which also implement the `Composer` trait. This is done to avoid having a trait object, to be able to access all options fields when needed by just using `match` (Maybe you can guess by now that I don't like type erasure in Rust very much :P )

TornaxO7:
> No, the second call to composed_bounds() is receiving the options, which are another implementation of the Composer trait for the shapes. This is done to avoid having a trait object, to be able to access all options fields when needed by just using match on the style enum (Maybe you can guess by now that I don't like type erasure in Rust very much :P )

Eh, may I ask where the second implementation of the `Composer` is? I can't really find it :thinking:

flxzt:
for example for the smooth style [here](https://github.com/flxzt/rnote/blob/main/rnote-compose/src/style/smooth/mod.rs)

TornaxO7:
> for example for the smooth style [here](https://github.com/flxzt/rnote/blob/main/rnote-compose/src/style/smooth/mod.rs#L183)

oh ok, now I understand the structure of the `style` directory, thanks!

TornaxO7:
May I ask you, where you create/find the svg-files for the different shape builders like the svg image of the [`line`](https://github.com/flxzt/rnote/blob/main/rnote-ui/data/icons/scalable/actions/shape-line-symbolic.svg)?

flxzt:
@TornaxO7
These shapes are stored in a `shapestroke`, which like all other strokes implements the "StrokeBehaviour" trait. This trait has two required methods, `gen_images()` for rendering into a bitmap image which is displayed in the app and `gen_svg()` to generate a svg for export.
Now these methods pass in a `RenderContext` by `Piet`, the rendering library. This library has different backends, most useful for Rnote are the `cairo` and `svg` backend.

Since the StrokeBehaviour trait requires the `DrawBehaviour` trait which defines drawing to the generic `impl RenderContext`, for most strokes we can avoid code duplication by just implementing drawbehaviour and generate bitmap images and SVGs by using these two different backends in `gen_images()` and `gen_svg()`.

TornaxO7:
Eh, wait, I didn't mean with `where` where they are loaded, I rather mean *where* you created/downloaded those SVG xD Because I need one for the arrow

flxzt:
@TornaxO7 Oh wait I totally misread your question.^^ I am creating most icons myself in inkscape. I can do that too for the arrow

TornaxO7:
> 
@TornaxO7 Oh wait I totally misread your question.^^ I am creating most icons myself in inkscape. I can do that too for the arrow

Ok, that's the answer which I was looking for xD

TornaxO7:
I'm not really experienced with `inkscape`, so a quick draw for an arrow would be neat :sweat_smile:

TornaxO7:
But good to know, then I can make a dummy icon for the time being

PR #303, opened 2022-10-18
split init function from workspacebrowser

TornaxO7:
This will clean up the init function without losing readability

PR #311, opened 2022-10-22
splitting up init function from workspacebrowser

TornaxO7:
Continue of: https://github.com/flxzt/rnote/pull/303

### DIFF
--- a/rnote-ui/src/workspacebrowser/mod.rs
+++ b/rnote-ui/src/workspacebrowser/mod.rs
@@ -350,16 +350,16 @@ fn setup_workspace_listbox(wb: &WorkspaceBrowser, appwindow: &RnoteAppWindow) {
 
     }));
 
-    {
-        let appwindow_c = appwindow.clone();
-        workspace_listbox.bind_model(Some(&wb.imp().workspace_list), move |obj| {
+    workspace_listbox.bind_model(
+        Some(&wb.imp().workspace_list),
+        clone!(@strong appwindow => move |obj| {
             let entry = obj.to_owned().downcast::<WorkspaceListEntry>().unwrap();
             let workspace_row = WorkspaceRow::new(entry);
-            workspace_row.init(&appwindow_c);
+            workspace_row.init(&appwindow);
 
             workspace_row.upcast::<Widget>()
-        });
-    }
+        }),
+    );
 }
 
 fn setup_prefix_listbox(wb: &WorkspaceBrowser, appwindow: &RnoteAppWindow) {
--- a/rnote-ui/src/workspacebrowser/mod.rs
+++ b/rnote-ui/src/workspacebrowser/mod.rs
@@ -167,18 +167,9 @@ impl WorkspaceBrowser {
         setup_add_workspace_button(self, appwindow);
         setup_edit_workspace_button(self, appwindow);
 
-        {
-            let remove_workspace_button = self.imp().remove_workspace_button.get();
-            let edit_workspace_button = self.imp().edit_workspace_button.get();
-            setup_workspacelist(
-                self,
-                appwindow,
-                remove_workspace_button,
-                edit_workspace_button,
-            );
-        }
-
+        setup_workspacelist(self, appwindow);
         setup_workspace_listbox(self, appwindow);
+
         setup_prefix_listbox(self, appwindow);
         setup_file_rows(self, appwindow);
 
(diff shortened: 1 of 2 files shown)


rnote-ui/src/workspacebrowser/mod.rs
```diff
@@ -163,64 +163,224 @@ impl WorkspaceBrowser {
...
+        {
+            let remove_workspace_button = self.imp().remove_workspace_button.get();
+            let edit_workspace_button = self.imp().edit_workspace_button.get();
+            setup_workspacelist(
+                self,
+                appwindow,
+                remove_workspace_button,
+                edit_workspace_button,
+            );
+        }
```
TornaxO7, line 179: Tbh, I'm a bit unsure if we should remove those `let` lines and move them inside to `setup_workspacelist` or not.

On the one hand you can clearly see here, that those two widgets are needed here but on the other hand: Is it necessary to see in the init function that the workspacelist needs those two widgets? What do you think?

flxzt: I would just move them into `setup_workspacelist()`. It is very cheap to call `imp()` and `.get()`, so it can be done in multiple places wherever one needs a reference to the widgets

TornaxO7: Ayeaye captain!

TornaxO7: done

rnote-ui/src/workspacebrowser/mod.rs
```diff
@@ -417,146 +561,20 @@ impl WorkspaceBrowser {
...
-    }
-
-    fn setup_dir_actions(&self, _appwindow: &RnoteAppWindow) {
-        self.insert_action_group("workspace_action", Some(&self.imp().workspace_actions));
-
-        self.imp()
-            .workspace_actions
-            .add_action(&workspace_action::create_dir(self));
-    }
+    // setup workspace rows
```
flxzt, line 569: this block is another functionality, it does not belong to the file row's setup

TornaxO7: whoops

rnote-ui/src/workspacebrowser/mod.rs
```diff
@@ -417,146 +561,20 @@ impl WorkspaceBrowser {
...
+    let appwindow_c = appwindow.clone();
```
TornaxO7, line 570: May I ask why we need a `.clone()` and not a `@clone(@weak appwindo ...)` for this one?

flxzt: I don't really remember

TornaxO7: Hm... ok, would you mind if I try to put it in a `@clone(@weak appwindo ...)`?

flxzt: i wouldn't!

TornaxO7: alright, this should be fixed now

flxzt:
Looks good to me, if you're finished I would merge it.

rnote-ui/src/workspacebrowser/mod.rs
```diff
@@ -163,64 +163,220 @@ impl WorkspaceBrowser {
...
+                    wb.set_current_workspace_dir(parent_dir.to_path_buf());
+                }
             }
         }));
+}
 
-        // Setup file rows
-        {
-            let primary_list_factory = SignalListItemFactory::new();
+fn setup_file_rows(wb: &WorkspaceBrowser, appwindow: &RnoteAppWindow) {
```
TornaxO7, line 376: I'm a bit unsure about this function... it still has a lot of lines and it contains different logic steps, like setup the FileFilter, create the sorter and prepare some other widgets. I'd like to split this up, but I'm unsure how. On the one hand I'd like to move this function into another file since the extra functions would be only used for this function, but on the other hand it might be a little bit too much, to create only *one* extra file for this. What do you think? (I'm a little bit tempted to do the same as my previous PR: Creating a new directory and then the different files xD)

flxzt: I think its okay, I really don't think there is any benefit to split it into smaller parts. The setup here is interconnectect and are understood together. It reduces readability when you have to jump around across files to read it.

Also: why obsess over some relatively trivial and cosmetic code changes for something that already works?

TornaxO7: >  It reduces readability when you have to jump around across files to read it.

Hm... this can be normally avoided by using keybindings and I think that most editors and IDE's support this feature, but I do get your point. Splitting the code too much up, can result to unreadable code instead of more readable.

> Also: why obsess over some relatively trivial and cosmetic code changes for something that already works?

I'd say: Readability and Maintainability :D (but yeah, this is probably a little bit *too* much "makeup" for the code.

TornaxO7:
Well, I think that I split up the main parts of the init function, so this can be merged now. What do you think?

flxzt:
Looks good! Thank you

PR #325, opened 2022-10-27
Update BUILDING.md, Outdated dependencies on Fedora 36.

ben-hansske:
Fedora 36 has gtk4 version 3.6.2 which is not recent enough to build rnote. Fedora 37 works AFAIK.


flxzt:
Thanks right, thanks!

PR #339, opened 2022-11-08
Split shaperpage init function

TornaxO7:
The reason, why I splitted up the init function of the shaperpage.rs is the same as for the [workspacebrowser.rs](https://github.com/flxzt/rnote/pull/311).

## Structure
As you can see from the file-structure, I move `shaperpage.rs` into a directory. The reasons for that are:
1. You can easier read, that it setups the given widget:
```rust
    pub fn init(&self, appwindow: &RnoteAppWindow) {
        shaperstyle_widget::setup(self, appwindow);
        shapeconfiguration_widget::setup(self, appwindow);
        border_width_widget::setup(self, appwindow);
        constraints_widget::setup(self, appwindow);
        stroke_colorpicker_widget::setup(self, appwindow);
        fill_colorpicker_widget::setup(self, appwindow);
        shaperbuilder_widget::setup(self, appwindow);
    }
```
IMO it's better than
```rust
    pub fn init(&self, appwindow: &RnoteAppWindow) {
        setup_shaperstyle_widget(self, appwindow);
        setup_shapeconfiguration_widget(self, appwindow);
        setup_border_width_widget(self, appwindow);
        setup_constraints_widget(self, appwindow);
        setup_stroke_colorpicker_widget(self, appwindow);
        setup_fill_colorpicker_widget(self, appwindow);
        setup_shaperbuilder_widget(self, appwindow);
    }
```
2. I saw that `shaperpage.rs` also has a `refresh_ui` which makes it possible to add the following convention for the most widgets if I read the code correctly: Each widget can be interact with `setup` and `refresh` functions, but I'm still not really sure, if I understood the `refresh_ui` fucntion completely.
3. Some widgets were too complex (had a lot of widgets on their own) that it wouldn't be worth it to split each widget into one single setup function, since even those would be worth it to split up, but having setup functions for the widgets of the widgets for the shaperpage would make the code more confusing than readable.
4. The `shaperpage.rs` logic is reduced to a smaller `init()` and `refresh_ui` function :)
5. The general idea is: If you want to add a new widget to the shaperpage: Create a new file and set it up!

What's your opinion to this?


flxzt:
Thanks for doing this work!

Since the original file is only 600 LOC even with the coordinate system PR I still am not too enthusiastic about splitting it into so many small files. If a composite widget grows too large I would rather split it up into multiple smaller self contained composite widgets with their own `.ui` file and everything, but here with many smaller default widgets which all modify a separate configuration value I don't think it is possible.

Splitting the setup functions would okay with me though, if you really want to do that. :)

TornaxO7:
> Since the original file is only ca. 600 LOC even with the coordinate system PR I still am not too enthusiastic about splitting it into so many small files. If a composite widget grows too large I would rather split it up into multiple smaller self contained composite widgets with their own .ui file and everything, but here with many smaller default widgets which all modify a separate configuration value I don't think it is possible.

I understand your arguments here and I also think that this would be probably the better solution.

> Splitting the setup functions (in the same file) would okay with me though, if you really want to do that. :)

I do want to do that, but I'm afraid that this wouldn't really improve the code-overview. Let's take a look into the [`constrains_widget.rs`](https://github.com/flxzt/rnote/pull/339/files#diff-9df9f860e90787e9fa22661b8a11a2486803c0bb14c628fa9cecf27ca3f301ea) for example. As you know, it sets up *four* widgets in one widget. If I split up the setup of the constraint widget in the same file, which should only handle the shaperpage in general imo due to its filename, then it would contain some logic about some widgets of its own widgets (if you understand what I mean here).
So I think, having separate `.ui` files would be suitable here or am I overthinking this? :thinking:

flxzt:
Thanks again, but as the maintainer I would be more lost in this structure in the future. In this case consistency in the codebase is more important here imho

PR #340, opened 2022-11-08
add three coordinate system shape builders

Kneemund:
Adds "2D coordinate system", "3D coordinate system", "2D single quadrant coordinate system".

Things to consider/discuss:
 1. Each shape builder currently emits 2 line shapes per axis, making it possible to only delete the positive/negative part.
 2. The 3D coordinate system is very simplified (the X axis is at an 45° angle), making it simple to use but not suitable for accurate representations.
 3. There is no 3D single octant coordinate system at the moment, and I don't think that's necessary either.
 4. The boundary of the shape builders is currently computed by merging the AABBs of all the lines that are involved; this could be heavily simplified by creating a single AABB using the axes as half extends, but doing so wouldn't take the shape style into account. Realistically, the `loosened` operation should take care of minor differences due to the shape style, but I went for the most accurate and "future-proof" solution for now.
 5. The 3D coordinate system builder could be turned into a two step builder (like the ellipse builder), allowing you to "define" your own X axis length.
 6. (These builders should be grouped and collapsible, but that should probably be done in a separate PR/commit.)
 7. (Rename some of the shapes? Improve some of the symbols?)


TornaxO7:
> 7. (Rename some of the shapes? Improve some of the symbols?)

Maybe a little suggestion to this one: I'd probably rename them to `CartesianCoordinateSystem2D` and `CartesianCoordinateSystem3D` since this is its [original name](https://en.wikipedia.org/wiki/Cartesian_coordinate_system).

Kneemund:
I left that out because I didn't want to make the names even longer (e.g. "2D single quadrant cartesian coordinate system"), but that does introduce a bit of ambiguity, yeah. I'm not entirely sure if these shapes should have short(er) or clear(er) names.

TornaxO7:
hm... I understand your point here, but I think that it's necessary here. Let's see what Felix says about this.

flxzt:
This is really nice, thanks a lot!

My thoughts:

> The 3D coordinate system is very simplified (the X axis is at an 45° angle), making it simple to use but not suitable for accurate representations.

I would say this is absolutely fine. Anyone who wants to create accurate plots is using latex tikz or a math plotting library, not a sketching app like rnote.

> There is no 3D single octant coordinate system at the moment, and I don't think that's necessary either

Me neither, a user can always delete some axis if they want to customize the graph

to. 4.:
I am totally okay with taking the BB's of the individual shapes to calculate the builder one. It does not really matter performance wise, since we are only talking about a maximum of 6 lines anyways.

> The 3D coordinate system builder could be turned into a two step builder (like the ellipse builder), allowing you to "define" your own X axis length.

I personally actually really like that the x-axis is the average of the other two. It feels very intuitive and should cover the common case where a graph has similiar axis lengths. But I can also see that users would want to specify the x-axis independently of the others. So either one would be okay for me

to 6.: agree. When using a grid, the names could appear in a tooltip when hovering over the icons.

regarding the naming:

the user facing strings can be renamed to `2d cartesian coordinates`, `3d cartesian coordinates` etc. in my opinion.

Intenrally maybe rename them to `CoordsCartesian2d`, `CoordsCartesian2dSingleQuadrant` `CoordsCartesian3d` ? A little shorter and the names are ordered "broad -> specific".

Kneemund:
Thanks!

> I personally actually really like that the x-axis is the average of the other two. It feels very intuitive and should cover the common case where a graph has similiar axis lengths. 

Yeah, it feels intuitive to me too.

> But I can also see that users would want to specify the x-axis independently of the others. So either option would be okay for me. If there is a second step, the x axis could be built with an arbitrary angle. So there is an advantage to it.

I wouldn't want that personally, because I'd never hit the 45° :P
Is there a situation where you *wouldn't* want an angle close to 45°? That seems fairly uncommon to me, and if someone needs it, they could draw a 2D coordinate system and add another axis using the line shape. The same thing goes for cases in which you want an arbitrary length, I guess. 

Either way, I don't really prefer one over variant the other, both have advantages and disadvantages in terms of simplicity and flexibility. I personally wouldn't add two separate builders though unless we go for a grid-style selection and have space to spare - but that could end up being confusing.

> I think the user facing string are okay as they are, internally the names can maybe be shortened with `Coordinate` -> `Coord`?

Alright, I'll do that.

flxzt:
> Is there a situation where you wouldn't want an angle close to 45°? That seems fairly uncommon to me, and if someone needs it, they could draw a 2D coordinate system and add another axis using the line shape. The same thing goes for cases in which you want an arbitrary length, I guess.

It could be solved with activating constraints, but it would be uninutive to do with a stylus *while* being in the in-between step. Like you said most users will then compose the coordinate system with an additional line themselves anyways. So simple builders trump more advanced ones here.

Otherwise looks all good to me. :) Feel free to add yourself in the `AUTHORS` file if you want.

Kneemund:
I agree.
Alright, thanks. :)

flxzt:
@kneemund just FYI and because I only noticed it now. I'll fix the slightly blurry icons by aligning the shapes inside the Svg with the pixel grid. Most of the time GTK renders icons with a 1to1 user unit -> px, so if you end up with blurry Svg icons somewhere this is the reason.

Kneemund:
@flxzt I just noticed that I didn't update anything related to i18n/localization in this PR. Is that automated? If not, do you have to (re-)generate anything or manually add the strings to `rnote.pot`?

flxzt:
No worries, and no it is not automated. I regenerate the po template every once in a while with `meson compile rnote-pot`. I then delete the strings out of the template that I don't want to be translated (like the changelog).
If you would add a gettext() call in a new source file or create a new `.ui` file you would need to add it to `POTFILES` but since this PR didn't do that nothing needs to be done.

PR #343, opened 2022-11-09
update output file path after filerow actions

Kneemund:
Currently, renaming and deleting files and directories in the workspace browser does not update the canvas output file path. This PR changes that, fixing some bugs:

- Renaming and deleting files that are opened causes automatic saving to fail, as the file does not exist at the outdated location.
- Manual saves recreate the old file that was supposed to be moved or deleted.


flxzt:
Thank you!

PR #347, opened 2022-11-13
small fixes related to unsaved changes

Kneemund:
- Set unsaved changes to `true` after deleting a file or one of its parent directories.
- Only show new file confirmation dialog if there are unsaved changes.

### DIFF
--- a/rnote-ui/src/dialogs/mod.rs
+++ b/rnote-ui/src/dialogs/mod.rs
@@ -94,6 +94,24 @@ pub fn dialog_clear_doc(appwindow: &RnoteAppWindow) {
 }
 
 pub fn dialog_new_doc(appwindow: &RnoteAppWindow) {
+    let new_doc = |appwindow: &RnoteAppWindow| {
+        appwindow.canvas().engine().borrow_mut().clear();
+
+        appwindow.canvas().return_to_origin_page();
+        appwindow.canvas().engine().borrow_mut().resize_autoexpand();
+        appwindow.canvas().update_engine_rendering();
+
+        appwindow.canvas().set_unsaved_changes(false);
+        appwindow.canvas().set_empty(true);
+
+        appwindow.app().set_input_file(None);
+        appwindow.canvas().set_output_file(None);
+    };
+
+    if !appwindow.unsaved_changes() {
+        return new_doc(&appwindow);
+    }
+
     let builder = Builder::from_resource(
         (String::from(config::APP_IDPATH) + "ui/dialogs/dialogs.ui").as_str(),
     );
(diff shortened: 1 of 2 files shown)


flxzt:
Thanks!

flxzt:
Thinking about it, we should probably install a [file monitor](https://gtk-rs.org/gtk-rs-core/stable/0.16/docs/gio/struct.FileMonitor.html) that updates this whenever we change the output file. This way changes even from outside of the app are registered

Kneemund:
True, I didn't consider that. I'd look into it, unless you already started working on it.

flxzt:
thanks, and no I didn't

Kneemund:
I can't get the file monitor to work; the `FileMonitor#connect_changed()` callback never fires. I'm either doing something wrong, or it's a flatpak issue. I don't really want to do a native build since I'd have to install it and I'm currently on my main machine. Do you know if there are any flatpak permissions I need to change/grant?

flxzt:
I'm not 100% sure, but I think it should resolve when you install the development flatpak bundle (if you can do that on your machine). I used to have a similar issue with the directory list which internally is using a file monitor as well, I documented the issue [here](https://github.com/flxzt/rnote/blob/main/BUILDING.md#bugs-and-workarounds).

Kneemund:
The file explorer works and updates fine when using flatpak. I also tried a native build on a different machine, but it doesn't work there either.

This is one of the things I've tried:
```rs
log::error!(
    "FILE EXISTS: {}",
    file.query_exists(None::<&gio::Cancellable>)
);

file
    .monitor(gio::FileMonitorFlags::all(), None::<&gio::Cancellable>)
    .unwrap()
    .connect_changed(
        glib::clone!(@weak self as app_window => move |_monitor, _file, other_file, _event| {
            app_window.canvas().set_output_file(other_file.cloned());
            log::error!("SOMETHING CHANGED");
        }),
    );

log::error!("CONNECTED MONITOR");
```

The output is:
```
FILE EXISTS: true
CONNECTED MONITOR
```
`SOMETHING CHANGED` never appears when renaming the file through the file browser or using external tools (Nautilus).
The file path is right too, I've printed that before. I guess that my code just isn't correct, I can't see what's wrong though. I don't have much experience with GIO, so it might be something obvious.

flxzt:
Ah, I remember: For file monitors you need to keep the monitor around, so something like this works as long as you don't drop `monitor`: 

```rust

    let monitor = file.monitor_file(gio::FileMonitorFlags::all(), None::<&gio::Cancellable>).unwrap();

    monitor.connect_changed(move |_monitor, _file, other_file, _event| {
            println!("File changed");
        });

```

Kneemund:
That works, thanks. 
However, `other_file` (the `Option<gio::File>` containing the new/updated information) is `None` when moving the file out of the current directory (...this being possible isn't mentioned in the [documentation](https://docs.gtk.org/gio/signal.FileMonitor.changed.html) either). Assuming that this is the intended behavior, we can "only" use this for renaming and have to unset the output file if it is moved.

PR #358, opened 2022-11-16
adjust setter_button CSS, closes #305

### LINKED ISSUE #305: Color Selection is slightly shifted [ui]
**Describe the bug**
The position of the colors is slightly wrong

**Screenshots**
The black color field has a different indent than the other fields. It's shifted more to the right
image

This seems to happen with all colors. The yellow field on the bottom ever so slightly overlaps with the green one.
image

 - OS: Fedora
 - Installation source Flatpak
 - Desktop Environment Gnome

Kneemund:
The PR improves the contrast of the border around color buttons by mixing the default border color with the current foreground color. This fixes the issue mentioned in the first part of #305 - if you look closely, you can see that there is a border around the first button, but it is nearly identical to the current background color, making it look like an indentation issue.

dark theme             |  light theme
:-------------------------:|:-------------------------:
  |  

The PR also adds `1px` of margin to the bottom of each button. This improves/fixes (?) the second issue mentioned in #305.

0px | 1px | 2px
:-------------------------:|:-------------------------:|:-------------------------:
 |   | 

EDIT: The PR also fixes the border radius and brightness transitions when switching between or hovering over/clicking on colors.

---

TLDR:
- Closes #305.
- Fixes transition when hovering over & switching colors.

### DIFF
--- a/rnote-ui/data/ui/style.css
+++ b/rnote-ui/data/ui/style.css
@@ -41,13 +41,14 @@
 
 .setter_button {
     padding: 0 0 0 0;
+    margin-bottom: 1px;
     background-blend-mode: screen;
     background-image:
         linear-gradient(45deg, #11111188 25%, transparent 25%, transparent 75%, #11111188 75%, #11111188),
         linear-gradient(45deg, #11111188 25%, transparent 25%, transparent 75%, #11111188 75%, #11111188);
     background-size: 20px 20px;
     background-position: 0px 0px, 8px 8px;
-    border-color: @borders;
+    border-color: mix(@borders, @theme_fg_color, 0.25);
     border-style: solid;
     border-width: 1px;
     filter: brightness(100%);
--- a/rnote-ui/data/ui/style.css
+++ b/rnote-ui/data/ui/style.css
@@ -48,11 +48,9 @@
         linear-gradient(45deg, #11111188 25%, transparent 25%, transparent 75%, #11111188 75%, #11111188);
     background-size: 20px 20px;
     background-position: 0px 0px, 8px 8px;
-    border-color: mix(@borders, @theme_fg_color, 0.25);
     border-style: solid;
     border-width: 1px;
     filter: brightness(100%);
-    transition: border-radius 0.15s ease-out, filter 0.15s ease-out;
 }
 
 .setter_button:hover {
--- a/rnote-ui/src/colorpicker/colorsetter.rs
+++ b/rnote-ui/src/colorpicker/colorsetter.rs
@@ -180,7 +180,8 @@ mod imp {
 .setter_button {{
     margin{0}: 6px;
     background-color: rgba({3}, {4}, {5}, {6:.3});
-    transition: margin{0} 0.15s ease-out;
+    border-color: alpha(darker(rgba({3}, {4}, {5}, {6:.3})), 0.3);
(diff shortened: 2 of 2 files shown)


flxzt:
I think adding the 1px margin could even be enough, a black color still looks a bit offset then but any grey shade as the border color looks unpleasant to my eyes.

Kneemund:
Yeah, that does look better. I also noticed that the transition CSS property is set (and thus overwritten) in the same file (L184). This removes the transition of the border radius and hover/active brightness filter - I'd fix that by moving those transitions into `colorsetter.rs` while I'm at it.

flxzt:
You are right, I guess I never looked to closely and the border radius kiind of looks animated while the button is rectracting, but it is actually not. Nice, thanks!

flxzt:
There were some issues with button visibility when the color is transparent, so I reverted to the original border color with less transparency in 11c78e1c937ae85188bd426044d83c64e12f1b56. It is not as bad as before, but the black color looks a bit offset again. On the other hand, it is difficult to find the right border that works for all color and alpha's, and I think  button visibility is more important

Kneemund:
Hm, I'd rather adjust the previous way of calculating the color than use `@borders` again. For example, you could use an alpha component of one, which should improve visibility, and use `mix()` instead of `darker()`/`lighter()` for more fine-grained control.

`darker(rgba({3}, {4}, {5}, 1.0))`:
grafik

If the borders are too dark, then you could e.g. `mix()` with black and remove `darker()`.

flxzt:
The issue I had was when the button is set to completely transparent, and we take the the selected color, we either have: with low alpha: A barely visible border resulting that the checkerboard blends into the background (thats what I meant with visibility), **or** with high alpha: whatever color the user has set without the alpha, so basically could be any color.

High alpha also look not very good, it is like the high contrast theme and I would rather have the look 1px offset which only is visible when you really pay attention to it.

I *think* `@border` is already the fg color, see https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/named-colors.html#helper-colors . Not entirely sure what the docs is referring to with `currentColor` but looking at the css in the inspector it is either black (in light theme) or white (in dark theme) with low alpha.

Kneemund:
Fair enough.

What about `mix(mix(@theme_bg_color, rgba({3}, {4}, {5}, 1.0), {6:.3}), @theme_fg_color, 0.25)`? That's high alpha without the drawbacks, because it's manually blended with the background color *before* being brightened/darkened.

|light | dark |
|---|---|
| grafik | grafik |

flxzt:
Yeah thats good. ( with a little less fg_color, mixed with 0.2 ). I'll change it, but then its settled. :P I don't want to waste time on this pedantic little thing anymore 😄 )

But the checkerboard pattern looks bad in your screenshots. I wonder why that is, I don't have those diagonal gaps..

Bildschirmfoto vom 2022-11-22 16-11-48

flxzt:
okay, does 18px with 9x offset work too? That size would be better anyway because then the pattern aligns better to the size.
I guess it could still happen, the only other option I see is to implement custom rendering via snapshot() and move away from css for the widget completely. But at least for the next release (which I hope to do very soon) I think changing this back to either 20px or 18px is the best solution.

Kneemund:
Both 18px and 20px work for me, yes.
We'll know soon enough if there are other people who have this issue at different sizes, at least if they report it. :P

flxzt:
That was my thought too. We'll  see

PR #380, opened 2022-11-27
add file monitor to output file

Kneemund:
* This PR implements a file monitor for the current output file, as discussed in #347.
* Uses toasts to indicate moves/deletions, renames, and modifications (with a button for reloading).
  * Updates the output file and unsaved changes flag accordingly.
* Refactors toasts and adds toasts with buttons.
* It makes special treatment for file row actions redundant, which is why it partially reverts #347 and fully reverts #343.

(For more details, see discussion below.)

### DIFF
--- a/rnote-ui/src/canvas/mod.rs
+++ b/rnote-ui/src/canvas/mod.rs
@@ -3,6 +3,7 @@ mod input;
 
 // Re-exports
 pub(crate) use canvaslayout::CanvasLayout;
+use gettextrs::gettext;
 use rnote_engine::pens::PenMode;
 
 // Imports
@@ -56,6 +57,7 @@ mod imp {
 
         pub(crate) output_file: RefCell<Option<gio::File>>,
         pub(crate) output_file_monitor: RefCell<Option<gio::FileMonitor>>,
+        pub(crate) output_file_monitor_signal: RefCell<Option<glib::SignalHandlerId>>,
         pub(crate) unsaved_changes: Cell<bool>,
         pub(crate) empty: Cell<bool>,
 
@@ -143,6 +145,7 @@ mod imp {
 
                 output_file: RefCell::new(None),
                 output_file_monitor: RefCell::new(None),
+                output_file_monitor_signal: RefCell::new(None),
                 unsaved_changes: Cell::new(false),
                 empty: Cell::new(true),
 
@@ -556,6 +559,30 @@ impl RnoteCanvas {
         }
     }
 
+    pub(crate) fn block_output_file_monitor(&self) {
+        let output_file_monitor = self.imp().output_file_monitor.borrow();
+        let output_file_monitor_signal = self.imp().output_file_monitor_signal.borrow();
+
+        if let (Some(output_file_monitor), Some(output_file_monitor_signal)) = (
+            output_file_monitor.as_ref(),
+            output_file_monitor_signal.as_ref(),
+        ) {
+            output_file_monitor.block_signal(&output_file_monitor_signal);
+        }
+    }
+
+    pub(crate) fn unblock_output_file_monitor(&self) {
+        let output_file_monitor = self.imp().output_file_monitor.borrow();
+        let output_file_monitor_signal = self.imp().output_file_monitor_signal.borrow();
+
+        if let (Some(output_file_monitor), Some(output_file_monitor_signal)) = (
+            output_file_monitor.as_ref(),
+            output_file_monitor_signal.as_ref(),
+        ) {
+            output_file_monitor.unblock_signal(&output_file_monitor_signal);
+        }
+    }
+
     pub(crate) fn init(&self, appwindow: &RnoteAppWindow) {
         self.setup_input(appwindow);
 
@@ -581,6 +608,7 @@ impl RnoteCanvas {
                 let output_file = canvas.output_file();
 
                 let mut current_output_file_monitor = canvas.imp().output_file_monitor.borrow_mut();
+                let mut current_output_file_monitor_signal = canvas.imp().output_file_monitor_signal.borrow_mut();
 
                 // cancel old monitor.
                 if let Some(old_output_file_monitor) = current_output_file_monitor.take() {
@@ -591,19 +619,46 @@ impl RnoteCanvas {
                 if let Some(file) = &output_file {
                     match file.monitor_file(gio::FileMonitorFlags::WATCH_MOVES, gio::Cancellable::NONE) {
                         Ok(output_file_monitor) => {
-                            output_file_monitor.connect_changed(
-                                glib::clone!(@weak canvas => move |_monitor, _file, other_file, event| {
-                                    if event == gio::FileMonitorEvent::Renamed || event == gio::FileMonitorEvent::MovedOut {
-                                        if other_file.is_none() {
-                                            canvas.set_unsaved_changes(true);
-                                        }
+                            let output_file_monitor_signal = output_file_monitor.connect_changed(
+                                glib::clone!(@weak canvas, @weak appwindow => move |_monitor, _file, other_file, event| {
+                                    match event {
+                                        gio::FileMonitorEvent::Changed => {
+                                            appwindow.canvas().set_unsaved_changes(true);
 
-                                        canvas.set_output_file(other_file.cloned());
+                                            let reload_toast = adw::Toast::builder().title(&gettext("Opened file was modified on disk.")).priority(adw::ToastPriority::High).button_label(&gettext("Reload")).timeout(0).build();
+                                            reload_toast.connect_button_clicked(
+                                                clone!(@weak appwindow => move |_reload_toast| {
+                                                    if let Some(output_file) = appwindow.canvas().output_file() {
+                                                        if let Err(e) = appwindow.load_in_file(&output_file, None) {
+                                                            log::error!("failed to reload current output file, {}", e);
+                                                        }
+                                                    }
+                                                })
+                                            );
+
+                                            appwindow.toast_overlay().add_toast(&reload_toast);
+                                        },
+                                        gio::FileMonitorEvent::Deleted | gio::FileMonitorEvent::MovedOut => {
+                                            canvas.set_unsaved_changes(true);
+                                            canvas.set_output_file(None);
+
+                                            adw::prelude::ActionGroupExt::activate_action(&appwindow, "text-toast", Some(&gettext("Opened file was moved or deleted on disk.").to_variant()));
+                                        },
+                                        gio::FileMonitorEvent::Renamed => {
+                                            // other_file *should* never be none.
+                                            if other_file.is_none() {
+                                                canvas.set_unsaved_changes(true);
+                                            }
+
+                                            canvas.set_output_file(other_file.cloned());
+                                        },
+                                        _ => (),
                                     }
                                 }),
                             );
 
                             *current_output_file_monitor = Some(output_file_monitor);
+                            *current_output_file_monitor_signal = Some(output_file_monitor_signal);
                         },
                         Err(e) => log::error!("creating a file monitor for the new output file failed with Err {}", e)
                     }
--- a/rnote-ui/src/canvas/mod.rs
+++ b/rnote-ui/src/canvas/mod.rs
@@ -589,22 +589,23 @@ impl RnoteCanvas {
 
                 // create new monitor.
                 if let Some(file) = &output_file {
-                    if let Ok(output_file_monitor) =
-                        file.monitor_file(gio::FileMonitorFlags::WATCH_MOVES, gio::Cancellable::NONE)
-                    {
-                        output_file_monitor.connect_changed(
-                            glib::clone!(@weak canvas => move |_monitor, _file, other_file, event| {
-                                if event == gio::FileMonitorEvent::Renamed || event == gio::FileMonitorEvent::MovedOut {
-                                    if other_file.is_none() {
-                                        canvas.set_unsaved_changes(true);
+                    match file.monitor_file(gio::FileMonitorFlags::WATCH_MOVES, gio::Cancellable::NONE) {
+                        Ok(output_file_monitor) => {
+                            output_file_monitor.connect_changed(
+                                glib::clone!(@weak canvas => move |_monitor, _file, other_file, event| {
+                                    if event == gio::FileMonitorEvent::Renamed || event == gio::FileMonitorEvent::MovedOut {
+                                        if other_file.is_none() {
+                                            canvas.set_unsaved_changes(true);
+                                        }
+
+                                        canvas.set_output_file(other_file.cloned());
                                     }
+                                }),
+                            );
 
-                                    canvas.set_output_file(other_file.cloned());
-                                }
-                            }),
-                        );
-
-                        *current_output_file_monitor = Some(output_file_monitor);
+                            *current_output_file_monitor = Some(output_file_monitor);
+                        },
+                        Err(e) => log::error!("creating a file monitor for the new output file failed with Err {}", e)
                     }
                 }
 
--- a/rnote-ui/src/canvas/mod.rs
+++ b/rnote-ui/src/canvas/mod.rs
@@ -57,9 +57,9 @@ mod imp {
         pub(crate) engine: Rc<RefCell<RnoteEngine>>,
 
         pub(crate) output_file: RefCell<Option<gio::File>>,
-        pub(crate) output_file_expect_write: RefCell<bool>,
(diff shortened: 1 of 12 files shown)


rnote-ui/src/canvas/mod.rs
```diff
@@ -571,6 +573,34 @@ impl RnoteCanvas {
...
+                                        canvas.set_unsaved_changes(true);
+                                    }
+
+                                    canvas.set_output_file(other_file.cloned());
+                                }
+                            }),
+                        );
+
+                        *current_output_file_monitor = Some(output_file_monitor);
+                    }
```
flxzt, line 601: lets log the error in an `else {}`

flxzt:
I wonder why it only sometimes knows the new file. Did you investigate into this further (is this a flatpak directory permission issue ? ).

Kneemund:
I wasn't able to find out why it behaves like that yet, no. I created this PR on a machine with native builds/installations, where it worked sometimes. I'm now running a flatpak build and it didn't even work once yet (I haven't tried it very often though).

Maybe flatpak permissions make the issue worse, but it (sometimes) happens regardless of the runtime environment.

flxzt:
On my machine rename always gives the new file, moving never does with flatpak and native builds

Kneemund:
I think I found somewhat of a pattern: it only works if the target/new directory is opened in the workspace browser. At least on flatpak, I'm not sure how native builds behave (and I can't test it ATM). 

(I've tried this using `monitor(gio::FileMonitorFlags::all(), gio::Cancellable::NONE)`.)

flxzt:
You're right, the current workspace directory affects this. Only when the file is moved **into** the directory it gives the new file. Both native and as flatpak

Another thing: I think we should match on the monitor event, and handle all cases accordingly ( like deletes )

Kneemund:
> Another thing: I think we should match on the monitor event, and handle all cases accordingly ( like deletes )

Right, I assumed that `WATCH_MOVES` only emits renames, move ins and move outs. But I guess that moving to the trash folder counts as a move out, not a deletion, which is why I've never seen it while debugging. I can't think of other events (besides renaming, moving and deleting) that we have to handle though.
Also, if we figure out how to properly track moves across all folders, then we'll have to check for trash folders (and maybe some other special cases).

flxzt:
When I move the file across partitions, I always get a delete, so there is also a boundary there

Kneemund:
Alright. If the "explanation" in my previous message is correct, then I'd say that we should treat moves and deletions equally (i.e. clear output file, set unsaved changes) in order to keep things consistent, like you suggested.

flxzt:
> I guess that this is because the file list monitors the opened directory? 
My first assumption too.  I tried to install a directory monitor along with the file monitor everytime the output file changes, but that didn't seem to make a difference.

I assume more knowledge about inotify would help, I have to read up on it.

Maybe we shouldn't at all track where the file moved. We could also mimic the behaviour of `gnome-text-editor` which simply spits out a warning with a reload button when the active file has changed in any way outside of the app, I start to think this is  probably the most sensible option and likely what users expect anyway.

The only thing we then probably can handle is rename. That also gets emitted when the file is saved over itself from the save dialog.

Edit: yes, agree with your last comment

Kneemund:
It seems like the move destination issue is a limitation of inotify (but expected behavior), at least according to [StackOverflow](https://stackoverflow.com/questions/5000880/can-inotify-tell-me-where-a-monitored-file-is-moved).

We could use the "changed" event for that, but I'm not sure how we should differentiate between rnote's and external changes/file modifications. I guess the simplest options would be to either cancel the monitor before and recreate it after the write, or to use some kind of flag that's set before and removed after the write (we'd have to consider the delay though; the first option sounds more reliable).

> The only thing we then probably can handle is rename. That also gets emitted when the file is saved over itself from the save dialog. we need to carefully consider this as well

Does this cause any issues? I guess that it's redundant and can be avoided by comparing the two files, but it shouldn't do any harm, right?

flxzt:
> It seems like the move destination issue is a limitation of inotify
ah now it makes (slightly) more sense then that gio is able to give us the moved file, when the process (or glib context ?) already watches the directory. My guess is that it 

I'd prefer the first option too. Every save is done with `save_document_to_file()` in `imexport:rs:374` so when we save the signal handler id that is returned by `connect_changed()` we can block it temporarily with [monitor.block signal()](https://gtk-rs.org/gtk-rs-core/stable/0.16/docs/glib/object/trait.ObjectExt.html#tymethod.block_signal) before writing to disk and unblock it right after.

> Does this cause any issues? I guess that it's redundant and can be avoided by comparing the two files, but it shouldn't do any harm, right?

I meant: if we decide to display a warning or something. But if we do the above, this doesn't matter.

Should we follow text-editor's example and keep the output file when the file changed on disk or not? Personally I think reverting to "New Document" (maybe differentiate by displaying "unsaved document" ?) is cleaner and creates less confusion actually but I don't have a very strong opinion on that.

Kneemund:
> I meant: if we decide to display a warning or something and the changed signal fire would fire. but if we do the above, this doesn't matter.

I still don't quite understand, sorry. I thought we only want to track file changes *because* we want to display notifications/warnings, or did I get something wrong?

> Should we follow text-editor's example and keep the output file when the file changed on disk or not?

I'm not sure about this. I think we should let the user decide, since there *might* be cases where file changes are out of the user's control. Simply clearing the output file without telling the user isn't a good idea IMO.

---

The general idea is the following then, right?

Update Output: `Renamed`.

Clear Output: `Deleted`, `MovedOut`.

Display Warning/Do Something: `Changed`.

Do Nothing: `AttributeChanged`, `Created` & `MovedIn` (only apply to directories AFAIK), `ChangesDoneHint` (we don't want to wait until all changes are done), `PreUnmount` & `Unmounted` (only sent if `WATCH_MOUNTS` is set, otherwise counts as `Deleted` I think), `Moved` (legacy, for deprecated flag).

flxzt:
> BTW, do we need to manually disconnect the signal before cancelling the file monitor?

This is poorly documented, but I don't think so. Since the handler is simply a callback and should get cleaned up when the object is destroyed there shouldn't be a problem. Looking at the code, cancelled simply prevents the emission of the changed signal: https://gitlab.gnome.org/GNOME/glib/-/blob/main/gio/gfilemonitor.c#L293

For the rest: yes sounds good! I think we should also display a warning for `Deleted / Moved Out` events to let users know what is going on. We could simply use toasts for it, like we do when exports fail. Something like this:

```rust
adw::prelude::ActionGroupExt::activate_action(&appwindow, "error-toast", Some(&gettext("Current file has changed on disk").to_variant()));
```

Kneemund:
Alright, using toasts sounds good to me.

What options should we present to users when the file is changed? I'm assuming "Overwrite Changes", "Reload from Disk" and "Clear/Reset Output File", but I'm unsure about naming them (especially the last one - maybe the default/"Cancel" action? doesn't sound very intuitive though).

Kneemund:
Do we really want to display this as a toast? What would the default action be? Overwrite or clear the output file? A dialog would allow us to present all 3 options, although it could also be a bit more disruptive.

flxzt:
The default when the user does not reload would be mark it as unsaved, keep the output file, and overwrite when saving

I'd like to avoid dialogs when they are not user initiated, the Gnome [Guidelines](https://developer.gnome.org/hig/patterns/feedback/dialogs.html?highlight=dialog) recommend using toasts for those things as well

Kneemund:
Fair enough. Having "Overwrite" *and* "Clear Output File" as options is kind of redundant too; you can just choose a different save location if you want to do the latter.

The only (foreseeable) issue that remains is how we're going to deal with other Rnote instances overwriting the output file (causing a rename event, as I explained above).

BTW, something that I've noticed while looking at the dialog schema (completely unrelated):
I think "workspace" should start with a capital letter here: https://github.com/flxzt/rnote/blob/main/rnote-ui/data/ui/dialogs/dialogs.ui#L42 (all other dialog titles are fully capitalized too).

Kneemund:
Well, the issue is that the changed event isn't emitted. So the only option that I see here is "hardcoding" a check into the rename event branch, and dispatching the same toast that we'll use for changes whenever the previous file name was `.goutputstream-<hash>` (or something like that - whatever the temporary save file is called).

I've pushed the changes that I've made so far; I haven't refactored the toasts yet but most things should already work. I'm not sure if there should be a confirmation dialog for reloading the file after a change - if so, then simply using `open_file_w_dialogs` would result in misleading titles.

flxzt:
Ah I see, you mentioned this issue three times now but I somehow always missed what you were trying to tell me, sorry.

How annoying. Yeah, agree with you regarding the handling, checking that and then handling it like delete / movedout.

I don't think we need another dialog. If the user decided and clicked on reload, we should load without an additional prompt

Kneemund:
To be fair, I wasn't very clear about it either. :P
Alright.

One last issue I've noticed: when the file changes multiple times, the toasts stack indefinitely and you have to close them one-by-one. I don't think that we can fix this, because - AFAIK - we can only add toasts to the queue and the only callback we have is the button click one. Adding a timeout would improve this, but I don't think that there should be one in this case.

---

What about refactoring toasts like this? This is without any `Option`-al arguments, because I couldn't get the closure argument to work properly.

<details>
<summary> refactored toast dispatch functions </summary>

```rs
pub fn dispatch_toast_button<F: Fn(&adw::Toast) + 'static>(
    &self,
    text: &String,
    button_label: &String,
    button_callback: F,
    timeout: u32,
) {
    let text_notify_toast = adw::Toast::builder()
        .title(text.as_str())
        .priority(adw::ToastPriority::High)
        .button_label(button_label)
        .timeout(timeout)
        .build();
    text_notify_toast.connect_button_clicked(button_callback);

    self.toast_overlay().add_toast(&text_notify_toast);
}

pub fn dispatch_toast_text(&self, text: &String) {
    let text_notify_toast = adw::Toast::builder()
        .title(text.as_str())
        .priority(adw::ToastPriority::High)
        .timeout(5)
        .build();

    self.toast_overlay().add_toast(&text_notify_toast);
}

pub fn dispatch_toast_error(&self, error: &String) {
    let text_notify_toast = adw::Toast::builder()
        .title(error.as_str())
        .priority(adw::ToastPriority::High)
        .timeout(0)
        .build();

    self.toast_overlay().add_toast(&text_notify_toast);
}
```
</details>

flxzt:
> One last issue I've noticed: when the file changes multiple times, the toasts stack indefinitely and you have to close them one-by-one.

We could add another method, `dispatch_toast_w_button_singleton()` which saves the added toast and [dismisses](https://world.pages.gitlab.gnome.org/Rust/libadwaita-rs/stable/latest/docs/libadwaita/struct.Toast.html#method.dismiss) the previous one, if present. And connect the dismiss signal to take itself when it is dismissed by the user.

Kneemund:
Oh, I only looked at the `ToastOverlay` because I thought it would have those methods, not the toasts themselves.

flxzt:
I meant: In the dismiss callback which also fires when the user clicked on the "x", take the singleton toast out wherever it is stored (probably in appwindow as `RefCell<Option<Toast>>` ).

I'm not sure how you'd implement the `toast_storage` parameter, where else would we want to store it? What we could also do to be able to have multiple singletons, is a hashmap with a string key or whatever as a parameter. Not sure we'd ever need this though

Kneemund:
I don't think that we have to take it out after it was dismissed, because calling dismiss on a toast that has already been dismissed just does nothing. Clearing the singleton is only important if we want to keep the old toast instead of replacing it, but I think replacing it is better because it serves as a visual indicator that the file has been modified *again*.

Well, I'd make the parameter a mutable reference to an `Option<Toast>` and create a separate variable for holding it - in this case one beneath `output_file_monitor` and `output_file_monitor_signal`. That way, we could have multiple toast types, of which only one is queued at a time.
(If we don't take the singleton toast out like I've described above, then the function won't have to hold on to the mutable reference for long.)

A map would also work, but that's probably a bit less flexible.

flxzt:
Oh I wasn't saying calling dismiss again in the callback, just taking out the object to drop it.  

Anyway, regarding where the toasts should live, lets do it your way (maybe with an explaining  `///` doc comment above the function)

Kneemund:
Alright, yeah. I just pushed did what I imagined, everything *should* work.
We just talked past each other regarding some things. :P

rnote-ui/src/canvas/mod.rs
```diff
@@ -571,6 +602,73 @@ impl RnoteCanvas {
...
+                                                    }
+                                                }
+                                            }
+
+                                            // other_file *should* never be none.
+                                            if other_file.is_none() {
+                                                canvas.set_unsaved_changes(true);
+                                            }
+
+                                            canvas.set_output_file(other_file.cloned());
```
flxzt, line 652: I would dispatch a "regular" toast here, just to let the user know that the file got renamed.

flxzt:
Okay, I checked it out and the behaviour mostly looks good, except for one remaining issue: When a document is opened in two windows, the "modified on disk" toast still pops up in both for some reason:

[Video_2022-12-04_13-08-52.webm](https://user-images.githubusercontent.com/19841886/205489756-5d82df88-4140-43f5-b50a-65b7c59bb4bb.webm)

You'd also need to do a rebase, I refactored most of the error messages. The conflicts shouldn't be difficult to resolve though

Kneemund:
Alright. I can't reproduce the issue you showed in the video though. Do you mind logging the event type and file(s) of the file monitor event? I'm assuming that the signal blocking isn't working correctly/the signal gets unblocked before inotify knows about and reports the event. The only workaround I can think of is setting a flag on each save and discarding the next event.

Kneemund:
I added an `expect_write` flag and refactored some things, everything should work now (I hope). I still have some cleanup to do, I accidentally pushed my `config.rs` and also noticed that some of your changes to the error messages were reverted during the rebase...

flxzt:
yeah the config.rs situation is annoying, I believe I accidentally committed the native build version earlier today. I would add it to .gitignore but the cargo check CI job needs it :/

And sure you can add the indentation stuff here,  no worries (and thanks!)

The monitor also now works, I'll merge it then. Probably doing a release very soon as well

rnote-ui/src/canvas/mod.rs
```diff
@@ -554,6 +582,75 @@ impl RnoteCanvas {
...
+                        match event {
+                            gio::FileMonitorEvent::Changed => {
+                                dispatch_toast_reload_modified_file();
+                            },
+                            gio::FileMonitorEvent::Renamed => {
+                                // if previous file name was .goutputstream-<hash>, then the file has been replaced using gio.
+                                if FileType::is_goutputstream_file(&file) {
+                                    if canvas.output_file_expect_write() {
+                                        // => file has been modified due to own save, don't do anything.
+                                        canvas.set_output_file_expect_write(false);
```
flxzt, line 611: maybe it should also be cleared on the changed event or even always at the end of the changed callback? I am thinking about other platforms (maybe a future port to windows ) where the behaviour might be different. It makes me a bit anxious that clearing the flag is behind so many conditions.

Kneemund: Right. I'd personally clear it after every event then, i.e. move it below line 636.

flxzt: yeah, I'm not confident enough that this always will be cleared. The worst thing that could happen is an invalid warning when a user changes the file externally while it is being written, but I don't think thats really problematic

Kneemund: Yeah. On another note, I think the part after the OR should be removed in the snippet I sent above, right? I don't think that the output file is ever set to a `.goutputstream-` file, but I might be missing the thought that I had while I writing this.

flxzt: > We could even remove the condition here:

I think so too actually, recreating the monitor wouldn't be an issue, would it?

and also agree with always clearing it at the end of the callback

flxzt: I would remove this `if !canvas.output_file_expect_write() || !FileType::is_goutputstream_file(output_file)` entirely, because it could be an issue when saving the file for the first time. when only `!is_outputstream_file()` is removed the output monitor will never be created I thikn

Kneemund:
I moved the `expect_write` clear below the match block (as discussed above), and fixed two edge cases I noticed: the "output file is modified" toast should be dismissed if you save the document or if you set the output file to `None` (e.g. after moving/deleting it).

flxzt:
I don't think the `dismiss` call when loading in the file is an issue, its the right place for it. And another flag or something would make it even more complex and difficult to understand (and maintain).  
These kind of behaviours are inherently complex, I don't think there is a more elegant solution

I don't notice anything else either, so looks good to me. Thanks a lot for this!

Issue #263, opened 2022-09-26
Half infinite document layout


D0V4HKIIN:
**Describe the solution you'd like**
A mix of the infinite and the continuous vertical document layout where there is a border to the top and the left but not on the right nor the bottom.

ktotam1:
this app would be way more useable with this feature imo

flxzt:
not opposed to adding this but what is the downside of the infinite layout in comparison to this?

D0V4HKIIN:
I use the continuous vertical layout and sometimes what i write is a bit longer than the width of the page which is annoying because then i have to split the formula/sentence in 2 lines. I guess infinite would allow me to fix that but it's kind of annoying to use because i have to be more precise when scrolling up and down to look at older notes.

D0V4HKIIN:
I feel like my explanation's really bad, but the idea is the ssme as a text editor. You don't need it nor do you want to be able to scroll to the left or over the top of the file

flxzt:
No, makes sense! I'll review the PR soon

Issue #134, opened 2022-03-18
[GTK] Document layout submenu has broken rendering


defini-tiv:
Steps to reproduce:

1. Open 3-dot menu and go in “Expand Mode” menu.
2. Click on canvas to close the menu
3. Open Menu again and see both main and expand mode menu weirdly overlapping.

![grafik](https://user-images.githubusercontent.com/68238058/159350847-0c0d72dd-b6d0-43e6-bdf7-d00146b1d36d.png)

flxzt:
hmm, I can't replicate this one. Can you capture a screenshot?

maymage:
I can't reproduce this, @defini-tiv

flxzt:
closed due to non-reproducibility. This also looks like a theming issue, not something that Rnote can fix.

PR #386, opened 2022-11-29
continuous vertical and horizontal layout implementation (#263)

D0V4HKIIN:
Implementation of the layout i described in this issue: https://github.com/flxzt/rnote/issues/263

There is a bug that makes the layout menu overlap when selecting a layout and then reopening the menu.

Also I am unsure about the naming of the layout. Some ideas i had: 
- Semi-infinite
- Half-infinite
- Continuous
- Continuous xy
- Continuous vertical and horizontal


D0V4HKIIN:
Oh the bug i discovered was already there, I found a discussion about it in this issue here https://github.com/flxzt/rnote/issues/134.

flxzt:
Thanks for implementing this!

The implementation looks good when we think of it as a 2 dimensional version of the continuous vertical mode, but I am not sure we should implement it like this.
When writing, I think it is irritating that it always resizes horizontally after a stroke especially when zoomed out when the pages don't fit the whole screen width.  
Maybe we should do a combination, where you can add pages columns when wanted (equivalent the "fixed size" mode), and vertically it expands automatically.

D0V4HKIIN:
You are right. Maybe it should be implemented in a similar way to the infinite layout instead of the continuous vertical layout. I will give it a try so we can try out the different layouts.

D0V4HKIIN:
Okay i am very unsure about how i implemented this. I don't really understand the allocate function in canvaslayout.rs. But it seems to be working nicely.

flxzt:
I think the PR already looks pretty good!  
Imo the "semi-finite" layout is the way to go. Does it still provide the benefits regarding scrolling?

canvaslayout simply updates the scrollbars values and limits with the document dimensions and updates the rendering when the viewport changes.  
I think the only modification you'd need to make is to match on "semi-infinite" separately and set `h_lower` and `v_lower` so that it includes the document shadow value, for a margin on the upper and left sides.

D0V4HKIIN:
sorry, i haven't been able to work on this for a while now. What benefits about scrolling do you mean?

D0V4HKIIN:
So i removed the continuous xy layout and added the margins on top and to the left.
I am not able to replicate the bug where the canvas grows infinitely when scroll up or to the left.

flxzt:
Here is a screen capture of the issue, look at the scrollbars:

[Kooha-2023-01-10-13-22-15.webm](https://user-images.githubusercontent.com/19841886/211551198-ddd47f55-d4c5-46ca-8084-df6e3cd40301.webm)

The size also is not correct at first when switching to it.

D0V4HKIIN:
Okay, I have been able to replicate it now, it's happening when dragging to the top left.

D0V4HKIIN:
this introduces a new bug where zooming out when the view is in the top left of the document doesn't update the size... :/

D0V4HKIIN:
The solution feels kind of hacky but the scroll bars it behaves like it should in my testing.

D0V4HKIIN:
The size not being correct also seem to be happening with the inifinite layout.

D0V4HKIIN:
I think that fixing the size not being correct when switching to it is outside of the scope of this pr because it also happens with the infinite layout and it's not related to the new feature.

D0V4HKIIN:
I should stop writing code when I am tired, it only leads to horrible code xD

rnote-engine/src/document/mod.rs
```diff
@@ -194,6 +205,26 @@ impl Document {
...
+
+        let new_viewport_bounds =
+            &viewport.extend_right_and_bottom_by(na::vector![padding_horizontal, padding_vertical]);
+
+        let new_bounds = self.bounds().merged(new_viewport_bounds);
+
+        self.x = 0.0;
+        self.y = 0.0;
+
+        if new_viewport_bounds.maxs[0] >= self.bounds().maxs[0] {
```
flxzt, line 220: what is the reason for these conditions?

rnote-ui/src/canvas/canvaslayout.rs
```diff
@@ -87,6 +87,10 @@ mod imp {
                     (document.x + document.width) * total_zoom
                         + super::CanvasLayout::OVERSHOOT_HORIZONTAL,
                 ),
+                Layout::SemiInfinite => (
+                    (engine.document.x - Document::SHADOW_WIDTH) * total_zoom,
```
flxzt, line 91: lets use the recently added OVERSHOOT_HORIZONTAL const here

D0V4HKIIN: Yes, I merged without checking too much on the changes...
Also OVERSHOOT_HORIZONTAL shouldn't be dependent on the zoom right?

rnote-engine/src/document/mod.rs
D0V4HKIIN: The reason for these conditions is to fix the bug with the scrollbars expanding when dragging towards the top left.
It doesn't evaluate to true when the viewport is smaller than the documents bounds.
Actually I realize now that the bug should be fixed at the root and that this is more of a workaround.
I have really no clue why but for some reason the width of the document increases when the viewport has a negative offset. This leads to self.bounds() being bigger than what it should be which in turn leads to the scroll bars bug.

D0V4HKIIN: ooooh, it's the extends function that computes maxs - mins and mins is actually negative bc of the viewport...

rnote-ui/src/canvas/canvaslayout.rs
D0V4HKIIN: btw what is the difference between SHADOW_WIDTH and OVERSHOOT_HORIZONTAL?

flxzt: I wanted to separate them, so they can be tweaked independently :)

flxzt:
Alright, looks good to me. Thank you!

PR #396, opened 2022-12-08
fix windows build

flxzt:
This finally fixes the windows build and updates the build instructions.

rnote running on win 11

🎉


D0V4HKIIN:
375 files changed 0.o? most of them don't even seem to have changed. Is that because of the force pushes?

Issue #393, opened 2022-12-06
optimize CI workflow 


Kneemund:
The CI workflow can be optimized a bit. 

### Caching cargo build files and downloads.

![grafik](https://user-images.githubusercontent.com/49598842/206018009-df0f00c6-ab29-4724-b203-ba2d2dd28e5f.png)
![grafik](https://user-images.githubusercontent.com/49598842/206018112-8ae03ccf-b79a-474a-8153-9adc5b164ba4.png)

The bottom run is one without any caching, but also includes building the cache (took 2m 11s in this case).
The top run has an [exact cache hit](https://docs.github.com/en/actions/using-workflows/caching-dependencies-to-speed-up-workflows#cache-hits-and-misses) and is about twice as fast.

One cache is about 1GB ATM; GitHub allows for 5GB per cache and 10GB in total (older caches are automatically deleted if that limit is exceeded AFAIK). This could be fine-tuned if it ever becomes too big (e.g. only cache the build artifacts - downloading the dependencies is much quicker).

### Caching fedora packages.

The workflow could use a custom docker image hosted on docker hub that already contains all the needed dependencies, but maintaining that would be more difficult/fragmented. I'm not sure how you would otherwise (safely) cache `dnf` packages though.
This would shave off another 2.5 minutes. The `Dockerfile` would look something like this:

```dockerfile
FROM fedora:37

RUN dnf upgrade --refresh -y
RUN dnf install -y meson gcc gcc-c++ clang make automake cmake kernel-devel gtk4-devel libadwaita-devel poppler-glib-devel poppler-data alsa-lib-devel
```

Using e.g. `ubuntu-latest` instead of a container would also be faster, but not as fast as a custom image (and currently impossible due to ubuntu's GTK4 package being too old).

---

Not a PR yet because this needs to be discussed first, especially if the docker hub solution sounds good to you.

Kneemund:
Using a prebuilt image of the `Dockerfile` above, you get down to about 3m 2s (best possible circumstances). Downloading and verifying non-official images from DockerHub takes longer than I expected. I'm not sure if that's really worth the effort of maintaining an image.

flxzt:
Thanks for looking into it, there is definitely a lot that can be improved here. I am not very familiar with containers and CI, thats definitely an area I need to look into more. But because of that, I also would like to keep things as simple as possible.

I can't really add much here, you'd need to tell if its worth it. We could first start without the custom image, as that already brings a big improvement. We can always follow up with it later.

[...]

PR #404, opened 2022-12-16
CI: add cargo cache, use meson, rename steps

Kneemund:
* Adds caching like described in the first half of #393.
  * A cache is invalidated as soon as (the hash of) `Cargo.lock` changes. If that happens, the latest cache is restored anyway, but the cache is rebuilt at the end of the build process.
* Renames most of the steps, or adds names if they were missing.

Things to consider:
* `rustfmt` is already included in the used toolchain, and the step that's supposed to install it (`rustup component add rustfmt`) only logs that it's up to date. It could be safely removed as far as I can tell, but there would be no benefit in terms of time.


flxzt:
Should we take this as the opportunity to switch to `meson compile` & `meson install` as a job now, and use `_mesonbuild/cargo-home` as cached directory? It would be slower than `cargo check`, but we would test that the entire project with the data files builds as intended. And it looks like `check` is not much faster than actual compilation anyway

Kneemund:
Tried meson, it is indeed slower and the cache is 1.5GB instead of 1GB. 

We could also upload the rnote binary, but that takes another 37s and 424MB. Considering that it won't even be executable without a system installation and up-to-date schemas, I'm not sure if it's worth it (unless you know of a way to package it?). We should probably just go with flatpak for CD in the future, but we'll have to figure out caching first.

flxzt:
How much slower? But I think even then it is preferable to `cargo check` directly. I could integrate `cargo check` as a test in meson so we'd get the benefit of both.  
Regarding the cache size, does it matter as long as we don't hit the limit?

If we want to start uploading nightly builds I agree to stick to a flatpak package after setting that up

flxzt:
I added `app-cargo-check` and `cli-cargo-check` targets to the meson script in 0f42f1b45ac8be4b7e623266cec72f51098363f0 , so it is now possible to run only cargo check within meson.
So we could replace cargo check with something like:

```bash
meson compile app-cargo-check -C _mesonbuild`
meson compile cli-cargo-check -C _mesonbuild`
meson test -v -C _mesonbuild
```

I think we can then remove config.rs again and add it to the `.gitignore`.

Kneemund:
> How much slower? 

It took about 6min in total (with a cache hit). That *would* also check all the installation scripts though. The cache size doesn't really matter, no.

> I added app-cargo-check and cli-cargo-check targets to the meson script

The app check is working, but I don't think that the cli check is defined correctly (or I'm doing something incorrectly; meson can't find it).

> I think we can then remove config.rs again and add it to the .gitignore.

Alright, finally :)

flxzt:
Ah, you need to opt-in to enable the cli targets by setting `-Dcli=true` when configuring meson initially

Still an improvement, right? I think not too bad.

Kneemund:
Do we really need `cargo clippy` as a meson target? Can't we just run that directly, like `rustfmt`?
And do we want to run it on dependencies? If not, we can specify `--no-deps` and save quite a bit of time.

flxzt:
I was thinking about caching and I thought clippy needs to always build all dependencies for the lints. I think `--no-deps` is used to [exclude path dependencies](https://doc.rust-lang.org/nightly/clippy/usage.html#workspaces)

Kneemund:
`rustfmt` fails without `config.rs`, I forgot about that. Do we want to run it during the same job as everything else, or configure meson/add a meson target for it? Also, removing `config.rs` makes the git commit check fail if you don't have a generated (and ignored) version of that file.

flxzt:
> rustfmt fails without config  

Ah right, I forgot about that too. It would be consistent to add that to meson as well, I'll do it.

Is it possible to reuse the cache for different jobs? If yes, it would be better to separate clippy into its own job

> Also, removing config.rs makes the git commit check fail

We have the fmt check in CI, it might not be really necessary anymore and I always thought it is a bit too intrusive. Should we remove it, what do you think?  

> I also don't know why clippy doesn't work;  

I am not sure I follow, where doesn't it work? Do you think it'll work once the PR is merged?

I am also inclined to change the default workflow permissions for GITHUB_TOKEN [to read-only](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/#setting-the-default-permissions-for-the-organization-or-repository), this shouldn't break anything introduced here, right?

Kneemund:
> If yes, it would be better to separate clippy into its own job

Having multiple jobs would kind of be a waste of time, because each job has to pull the fedora image from docker and install all necessary dependencies - even the format job (because meson requires them). They would of course run in parallel, so it wouldn't be slower, it'd just drain your workflow minutes. I don't think that you can change that because each job runs in a separate VM. But that's your decision :P

Reusing the cache should be possible. I think that a cache miss would make both jobs miss and rebuild the cache though, so we might want to run them sequentially (that'd be a lot slower though, see above).

> I am not sure I follow, where doesn't it work? Do you think it'll work once the PR is merged?

Clippy fails to write code annotations for PRs, because GITHUB_TOKEN doesn't (and shouldn't) have write permissions there, and apparently that's a [known issue only GitHub can fix](https://github.com/actions-rs/clippy-check/issues/2). Instead, it posts an error, some warning messages and the results elsewhere (in the workflow overview).

> I am also inclined to change the default workflow permissions for GITHUB_TOKEN [to read-only](https://github.blog/changelog/2021-04-20-github-actions-control-permissions-for-github_token/#setting-the-default-permissions-for-the-organization-or-repository), this shouldn't break anything introduced here, right?

I'm not sure honestly, according to the issue I linked it seems that clippy does need write permissions. We could try that though.

flxzt:
> But that's your decision :P  
Yeah on second though it doesn't make much sense, the output can be inspected just as well when it is in the same job. I didn't hit any limits yet but since this PR is about optimizing the workflow I agree to keep it there then.

> Clippy fails to write code annotations for PRs  
So it doesn't work for PR's, right? I'd say an always failing job is not desirable even when the output is inspectable and I would then rather go wih the clippy target I added in meson. I don't care that much about the annotations tbh, as long as I see something its fine for me.  
And as you said, giving token write permissions for PR's shouldn't be done..

Okay, I'll change the repo token permissions to read-only. I believe if a jobs really needs it in the future it still can be overriden explicitely.

Kneemund:
Alright. Should rustfmt stay in its separate job? It's the same situation as having a separate job for clippy, except that combining it would make the "at a glance" output (e.g. at the bottom of PRs) considerably worse.

> So it always fails for PR's, right?

We can add `continue-on-error: true` to the step. That way even if it fails, it won't cause the whole job to fail. I think the error message doesn't even make the job fail, but I can test that. BTW, the clippy action falls back to printing the output to STDOUT in PRs.

flxzt:
>Alright. Should rustfmt stay in its separate job  

Probably not. Let's try it out having everything in one single job

> That way even if it fails, it won't cause the whole job to fail  

Ah okay, that would be fine then :)

Kneemund:
> Let's try it out having everything in one single job

Alright.

Do we want `continue-on-error: true` anyway? Clippy returns a non-0 exit code for errors, but I have no idea what those would look like (and if they deserve to mark the whole job as failed).

Kneemund:
Apparently my attempt to rebase this closed the PR. 😆

> We have the fmt check in CI, it might not be really necessary anymore and I always thought it is a bit too intrusive. Should we remove it, what do you think?

I don't have a strong opinion on it, but I'd personally rather see the formatting error before committing instead of during the PR checks. The hook saved me more than once already.

flxzt:
> do we want continue-on-error: true anyway

Yes I think so. For actual errors we have `cargo check`

> I don't have a strong opinion on it, but I'd personally rather see the formatting error before committing instead of during the PR checks. The hook saved me more than once already.

Okay, lets keep it then. I don't know how we could modify it to avoid failure when `config.rs` is not present, but it should be a rare occasion that someone would try to commit without building at least once anyway right?

Kneemund:
> Yes I think so. For actual errors we have cargo check

Alright. I added it.

> Okay, lets keep it then. I don't know how we could modify it to avoid failure when config.rs is not present, but it should be a rare occasion that someone would try to commit without building at least once right?

I added a check to the pre-commit hook that informs you about `config.rs` not being present (and that building the project generates it), and gives you the option to perform the commit anyway or abort it (I copied large parts of the rustfmt check below).

flxzt:
> I added a check to the pre-commit hook  

Awesome, thanks!

One more thing, then it should be good to go: lets install the `libappstream-glib` package to run the (optional) appdata file check as well.

flxzt:
Again thanks a lot for doing this! I really appreciate your help here

Kneemund:
No problem! :)

Kneemund:
Apparently the clippy action does *not* post its output to STDOUT if the "resource not accessible by integration" error occurs during non-PR runs. I'm assuming that this happens due to the missing write permissions. I guess we could  either switch to the meson target you've created, or give `GITHUB_TOKEN` write permissions again. Either way should work, I'd leave that decision to you :P
image

flxzt:
Okay, then lets stick to the clippy in meson, for simplicity and is in line with the rest

PR #405, opened 2022-12-16
fix output file monitor when saving document with changes to new file

Kneemund:
## Bug
When saving a document with unsaved changes to a new file, the "file was modified on disk" toast is shown:

https://user-images.githubusercontent.com/49598842/208193877-04e2b123-ac5c-44db-8b8c-1cbcd040b0a9.mp4

This bug is an oversight from #380. I was only able to reproduce it on native installations, not when using flatpak (likely due to differences in file system latency/inotify).

## Cause
This happens because the action described above actually emits two inotify events: the file is created first, and then written to/modified.

Bildschirmfoto vom 2022-12-16 22-07-33

## Fix
To fix this bug, the `expect_write` flag has to *also* be respected when the emitted event is an actual `Changed` event (see logs above) - not only when it's a `Renamed` event and the file name was `.goutputstream-<hash>` (i.e. regular save).

Additionally, we can't clear the `expect_write` flag after every event anymore, because `Created` is emitted before `Changed`.
We could clear it after every event except `Created`, but that's a bit risky (there could be similar edge cases). Only clearing the flag after it has been checked for in `dispatch_toast_reload_modified_file` is also risky (as discussed in #380). I went for what I think is the least risky solution: clearing it after every event that we explicitly handle.

## Things to Consider
* When to clear the `expect_write` flag (see above).

### DIFF
--- a/rnote-ui/src/canvas/mod.rs
+++ b/rnote-ui/src/canvas/mod.rs
@@ -596,6 +596,11 @@ impl RnoteCanvas {
                 output_file_monitor.connect_changed(
                     glib::clone!(@weak self as canvas, @weak appwindow => move |_monitor, file, other_file, event| {
                         let dispatch_toast_reload_modified_file = || {
+                            if canvas.output_file_expect_write() {
+                                // => file has been modified due to own save, don't do anything.
+                                return;
+                            }
+
                             appwindow.canvas().set_unsaved_changes(true);
 
                             appwindow.canvas_wrapper().dispatch_toast_w_button_singleton(&gettext("Opened file was modified on disk."), &gettext("Reload"), clone!(@weak appwindow => move |_reload_toast| {
@@ -610,15 +615,14 @@ impl RnoteCanvas {
                         match event {
                             gio::FileMonitorEvent::Changed => {
                                 dispatch_toast_reload_modified_file();
+
+                                canvas.set_output_file_expect_write(false);
                             },
                             gio::FileMonitorEvent::Renamed => {
                                 // if previous file name was .goutputstream-<hash>, then the file has been replaced using gio.
                                 if FileType::is_goutputstream_file(file) {
-                                    if !canvas.output_file_expect_write() {
-                                        // => file has been modified by external means, handle it the same as the Changed event.
-                                        dispatch_toast_reload_modified_file();
-                                    }
-                                    // else => file has been modified due to own save, don't do anything.
+                                    // => file has been modified, handle it the same as the Changed event.
+                                    dispatch_toast_reload_modified_file();
                                 } else {
                                     // => file has been renamed.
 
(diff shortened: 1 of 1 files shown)


flxzt:
Thanks! I don't like how this is getting even more complex, but I don't know how it could be handled otherwise. Maybe adding a comment in the code explaining why we can't clear the flag for every event would be nice.

What we could also do, unless I am missing another case, is clearing the flag when creating a new monitor, so we at least start in a clean state for a new file.

Not really related to this particular issue, but because we talked previously about the behaviour on windows and we now have a working build: I already expected it, it doesn't really work as intended. After a bit of testing it seems that every save simply emits:

```
ERROR rnote::canvas > event: Deleted, file: Some("C:\\Users\\Felix\\Downloads\\New Documents.rnote"), other_file: None
```

I believe, this is because `create_replace_file_future()` always calls `replace_future()` instead of ´replace_contents_future()` and it seems that the delete event is emitted before the monitor is replaced by the new one for the new file. Adding this:

```rust

if !canvas.output_file_expect_write() {
    appwindow.canvas_wrapper().dispatch_toast_text(&gettext("Opened file was moved or deleted on disk."));
}
```
Seems to fix it. I believe this should make it more robust on linux as well.

This should probably be refactored, but would likely result in new behaviour in the monitor so not in this PR. I you don't want to spend more time with it anymore, I can do it as well. ( If you could look at it as a reviewer then, I would really appreciate it :) )

Kneemund:
Yeah, I'm not sure how you'd simplify this. I'll add a comment below the event type match block.

> What we could also do, unless I am missing another case, is clearing the flag when creating a new monitor, so we at least start in a clean state for a new file.

That would mess up saving new documents, since that sets the output file, recreates the file monitor (-> resets the flag) and then creates and replaces the file (as above).

I think the straightforward solution is only clearing the flag after it blocked an event we care about. Although that doesn't sound very robust... But it's the only option without logical flaws that I know of (at least on linux, and assuming that we don't rethink the whole flag thing).

I'll add a return guard that checks for the flag in every match arm for now. That should cover the bug on windows you've mentioned and potentially more edge cases.

> Changing it to replace_contents_future() when we only want to overwrite the file would likely result in new behaviour in the monitor so lets do it in another PR. I could do it as well for a change. ( If you could look at it as a reviewer then, I would really appreciate it :) )

I wouldn't mind, but I'm not exactly sure where to look for all of this yet, so feel free to do it. And yeah, sure :)

flxzt:
> That would mess up saving new documents  
Ah right, I am losing the overview here.

Okay, looks good and on windows it is fixed as well. It loses the file as soon as it is written to by another window because of the `delete` event but I hope when I refactor to use `replace_contents_future()` that this will get fixed.

flxzt:
I tested `replace_contents()` but it doesn't make a difference here. The [docs](https://docs.gtk.org/gio/method.File.replace_contents.html) mention that it internally uses `replace()` so it seems to be just a different API where you give it the contents instead of writing to the output stream

Kneemund:
Hm, is there any way of differentiating between deletions and (external) saves on Windows then? Does the latter produce any other events, or only the delete event that you've posted above?

Issue #379, opened 2022-11-27
Could not type Chinese in Linux version with Fcitx5


Jisu-Woniu:
**Describe the bug**
Could not type Chinese in Linux version with IME. (I use Fcitx5)

**To Reproduce**
1. Go to 'Typewriter' mode
2. Create a textbox
3. Try entering Chinese with IME
4. Only letters get entered

**Expected behavior**
Chinese characters should get entered.

** Console Output **
No output.

**Screenshots**
![Screenshot](https://user-images.githubusercontent.com/31986081/204158166-e25a2843-7223-42be-b8aa-3bc7dd746dfa.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version: 22.10
 - Installation source: Flatpak, also tried build from source
 - Desktop Environment: KDE 5.26

**Additional context**
I can type in filename in "save file" dialog, so both the IME and GTK 4 configuration should have no problem.

Jisu-Woniu:
Thank you @Mr-Spade. This works well.

PR #410, opened 2022-12-21
Typewriter supports input methods of CJK and other languages

Mr-Spade:
Allow typewriter to use system input methods, which solve the issue like [this](https://github.com/flxzt/rnote/issues/379).


flxzt:
Thank you!

Issue #239, opened 2022-08-21
Remove duplicated tool icon


martin-desktops:
# Context
In version 0.5.4, currently selected tool is displayed in form of icon in mini-sidebar and as highlighted icon in headerbar (desktop view)  / under headerbar (mobile view). (see screenshots).

I understand where you were coming from - it was meant to indicate what the settings on sidebar refer to.
I think though that we should keep consistency across UI elements - every other icon in the main Rnote window serving as a button makes the duplicated tool icon looking out of place. In my opinion highlighted icon on headerbar is understandable enough, and also doesn't lead to confusions like this.

# Screenshots
![Screenshot from 2022-08-21 10-46-31](https://user-images.githubusercontent.com/64419373/185783321-5efee9cf-1a08-4232-9842-cfcc74bc79d5.png)
![Screenshot from 2022-08-21 10-46-39](https://user-images.githubusercontent.com/64419373/185783324-13a84c18-5a40-4e56-be5f-67b60595a1c6.png)

PR #420, opened 2022-12-23
Remove redundant sidebar tool icons 

martin-desktops:
Closes https://github.com/flxzt/rnote/issues/239

### DIFF
--- a/rnote-ui/data/ui/penssidebar/brushpage.ui
+++ b/rnote-ui/data/ui/penssidebar/brushpage.ui
@@ -10,18 +10,6 @@
     <property name="hexpand">false</property>
     <property name="vexpand">false</property>
     <child>
-      <object class="GtkImage">
-        <property name="icon_name">pen-brush-symbolic</property>
-        <property name="margin-top">9</property>
-        <property name="margin-bottom">9</property>
-      </object>
-    </child>
-    <child>
-      <object class="GtkSeparator">
-        <property name="orientation">vertical</property>
-      </object>
-    </child>
-    <child>
       <object class="GtkBox">
         <property name="orientation">vertical</property>
         <child>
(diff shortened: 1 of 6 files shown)


flxzt:
Hey, thanks for the PR. I am not sure this is actually what I want, since the first sidebar controls are awkwardly in the corner. Can you post a screenshot how it then looks? I am inclined to keep the icons or simply replacing them with empty space if the argument is that it is too cluttered visually.

martin-desktops:
Hello! Here you are

PR version:

Screenshot from 2022-12-23 15-00-13

Pre PR:
Screenshot from 2022-12-23 15-00-30

For me, a clutter wasn't the issue, Rnote's UI is quite simple.  Something just feels off that exactly the same icons are next to each other, it feels redundant, and two icons don't give more cues that one, and because one, highlighted one is enough, the second makes me think "Is it a clickable button?", "why there's another, exactly the same icon?". It's confusing.

Screenshot from 2022-12-23 15-04-12

flxzt:
Right, that is a good point. I don't like it when the icons are simply removed though. What we could do, is have a spacer there instead of the icon and bind its visibility to when "revealed" property of the pen toggle revealer.

martin-desktops:
I would love to see that solution in practice, but I am worried it's beyond my skills. I am not a programmer, here, I just removed a bunch of lines :)

flxzt:
Ah, no worries then. Here is how I solved it:


https://user-images.githubusercontent.com/19841886/209436346-ed6ffd09-ef0d-41e6-963e-97fdd82b9648.mp4

martin-desktops:
Thanks for consideration! 

The space separating a horizontal bar from the vertical one is definitely on place here, so the first item in the tool setting bar doesn't look like another tool. 

One thing that doesn't look right for me though is this rectangle made by these two light gray lines. I think it would look cleaner without them, or at least without the horizontal one, since the vertical one seperates these two bars without creating an empty rectangle.

Screenshot from 2022-12-24 13-41-58

flxzt:
Tried that too, it looks odd without it. I think I'll keep it for now

PR #422, opened 2022-12-23
Remove empty .gitmodules file

orhun:
I suppose there is no need to keep an empty file in the repository, submodules were removed in https://github.com/flxzt/rnote/commit/230e613bdb73798096fced5b7c9ece9e7fc68e55 🐻


flxzt:
Thanks!

PR #436, opened 2023-01-05
File list support for rnote-cli export

D0V4HKIIN:
Discussed briefly on matrix, this makes it possible to give a list of files to `rnote-cli export`.
Ex: `rnote-cli export *.rnote --output-format pdf`.

Something similar could probably be done for the imports.

### DIFF
--- a/rnote-cli/src/cli.rs
+++ b/rnote-cli/src/cli.rs
@@ -109,7 +109,13 @@ pub(crate) async fn run() -> anyhow::Result<()> {
                 None => {
                     for file in rnote_files {
                         let mut output = file.clone();
-                        output.set_extension(engine.export_prefs.doc_export_prefs.export_format.file_ext());
+                        output.set_extension(
+                            engine
+                                .export_prefs
+                                .doc_export_prefs
+                                .export_format
+                                .file_ext(),
+                        );
                         println!("exporting `{}`", file.to_string_lossy());
                         export_to_file(&mut engine, &file, &output).await?
                     }
--- a/rnote-cli/src/cli.rs
+++ b/rnote-cli/src/cli.rs
@@ -2,7 +2,7 @@ use rnote_engine::engine::export::{DocExportFormat, DocExportPrefs};
 use rnote_engine::engine::EngineSnapshot;
 use smol::fs::File;
 use smol::io::{AsyncReadExt, AsyncWriteExt};
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use clap::{Parser, Subcommand};
 
@@ -31,18 +31,22 @@ pub(crate) enum Commands {
         #[arg(long)]
         xopp_dpi: Option<f64>,
     },
-    /// Exports the Rnote file and saves it in the output file.{n}
-    /// The export format is recognized from the file extension of the output file.{n}
+    /// Exports the Rnote file(s) and saves it in the desired format.{n}
+    /// When using --output-file, only one input file can be given. The export format is recognized from the file extension of the output file.{n}
     /// When using --output-format, the same file name is used with the extension changed.{n}
-    /// Currently `.svg`, `.xopp` and `.pdf` are supported.
+    /// --output-file and --output-format are mutually exclusive but one of them is required.{n}
+    /// Currently `.svg`, `.xopp` and `.pdf` are supported.{n}
+    /// Usages: {n}
+    /// rnote-cli export --output-file [filename.(svg|xopp|pdf)] [1 file]{n}
+    /// rnote-cli export --output-format [svg|xopp|pdf] [list of files]
     Export {
         /// the rnote save file
         rnote_files: Vec<PathBuf>,
-        /// the export output file. Optional if an output format is specified.
-        #[arg(short = 'o', long)]
+        /// the export output file. Only allows for one input file. Exclusive with output-format.
+        #[arg(short = 'o', long, conflicts_with("output_format"), required(true))]
         output_file: Option<PathBuf>,
-        /// the export output format. Optional if an output file is specified.
-        #[arg(short = 'f', long)]
+        /// the export output format. Exclusive with output-file.
+        #[arg(short = 'f', long, conflicts_with("output_file"), required(true))]
         output_format: Option<String>,
         /// export with background
         #[arg(short = 'b', long)]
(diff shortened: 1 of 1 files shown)


flxzt:
Nice, thanks! We could even parallelize the for loop that iterates over the files. [rayon](https://docs.rs/rayon/1.6.1/rayon/) is already a dependency in rnote-engine, we could use it's `par_iter()`.  Creating an engine in each one is a lot faster in comparison to the export times

rnote-cli/src/cli.rs
```diff
@@ -33,13 +33,17 @@ pub(crate) enum Commands {
...
     /// Currently `.svg`, `.xopp` and `.pdf` are supported.
     Export {
         /// the rnote save file
-        rnote_file: PathBuf,
-        /// the export output file
+        rnote_files: Vec<PathBuf>,
+        /// the export output file. Optional if an output format is specified.
         #[arg(short = 'o', long)]
-        output_file: PathBuf,
+        output_file: Option<PathBuf>,
```
flxzt, line 43: More clarification that this is only valid when only one rnote file is specified would be nice here

rnote-cli/src/cli.rs
```diff
@@ -33,13 +33,17 @@ pub(crate) enum Commands {
...
+        /// the export output format. Optional if an output file is specified.
+        #[arg(short = 'f', long)]
+        output_format: Option<String>,
```
flxzt, line 46: And it would be best to mention that when this is specified that there can only be one rnote-file

rnote-cli/src/cli.rs
flxzt: And it would be best to mention that when this is specified that there can only be one rnote-file

rnote-cli/src/cli.rs
```diff
@@ -33,13 +33,17 @@ pub(crate) enum Commands {
     },
     /// Exports the Rnote file and saves it in the output file.{n}
     /// The export format is recognized from the file extension of the output file.{n}
+    /// When using --output-format, the same file name is used with the extension changed.{n}
     /// Currently `.svg`, `.xopp` and `.pdf` are supported.
     Export {
         /// the rnote save file
```
flxzt, line 39: file -> files. Maybe add a doc comment to explain the valid combinations here

rnote-cli/src/cli.rs
```diff
@@ -70,18 +74,51 @@ pub(crate) async fn run() -> anyhow::Result<()> {
...
+            };
+
+            match format {
+                Ok(format) => {
+                    // apply given arguments to export prefs
+                    engine.export_prefs.doc_export_prefs =
+                        create_doc_export_prefs_from_args(format, with_background, with_pattern)?;
+                }
+                Err(err) => return Err(err),
+            }
```
flxzt, line 99: I think determining the format should be moved into `create_doc_export_prefs_from_args`.

rnote-cli/src/cli.rs
```diff
@@ -70,18 +74,51 @@ pub(crate) async fn run() -> anyhow::Result<()> {
...
+                        export_to_file(&mut engine, file, output).await?
+                    }
+                    None => {
+                        return Err(anyhow::anyhow!("Failed to get filename from rnote_files."))
+                    }
+                },
+                None => {
+                    for file in rnote_files {
+                        let mut output = file.clone();
+                        output.set_extension("pdf");
```
flxzt, line 116: I think this shouldn't be hardcoded as pdf, right?

rnote-cli/src/cli.rs
flxzt, line 39: file -> files. Maybe add a doc comment to explain the valid combinations here

rnote-cli/src/cli.rs
```diff
@@ -159,8 +190,8 @@ pub(crate) fn create_doc_export_prefs_from_args(
 
 pub(crate) async fn export_to_file(
     engine: &mut RnoteEngine,
-    rnote_file: PathBuf,
-    output_file: PathBuf,
+    rnote_file: &PathBuf,
+    output_file: &PathBuf,
```
flxzt, line 194: a little pedantic, but `impl AsRef<Path>` would be better

rnote-cli/src/cli.rs
```diff
@@ -70,18 +74,51 @@ pub(crate) async fn run() -> anyhow::Result<()> {
...
+                (_, out_format) => get_export_format(out_format.as_deref()),
+            };
+
+            match format {
+                Ok(format) => {
+                    // apply given arguments to export prefs
+                    engine.export_prefs.doc_export_prefs =
+                        create_doc_export_prefs_from_args(format, with_background, with_pattern)?;
+                }
+                Err(err) => return Err(err),
```
flxzt, line 98: Why not use `?`

rnote-cli/src/cli.rs
D0V4HKIIN: oops yes

rnote-cli/src/cli.rs
D0V4HKIIN: im fairly new to rust and i didn't really understand the point of the `impl AsRef<Path>`

rnote-cli/src/cli.rs
D0V4HKIIN: didn't know that notation, very useful

rnote-cli/src/cli.rs
flxzt: well you could use it from the engine after it has been applied to it

rnote-cli/src/cli.rs
flxzt: with it you can either give it an allocated `PathBuf`, `&PathBuf` or `&Path`. More important in libraries that want to provide flexible API but it is useful here too, for the future

rnote-cli/src/cli.rs
D0V4HKIIN: ooh ok got it, didn't see the `file_ext()` function in `DocExportFormat`

rnote-cli/src/cli.rs
D0V4HKIIN: Yes, I have now also used clap to handle in a better way that they are required and conflicting between each other.

D0V4HKIIN:
Should fix the issues discussed

D0V4HKIIN:
As for the parallelization I will leave it for later/somebody else.

rnote-cli/src/cli.rs
flxzt: Nice, I like it!

flxzt:
> As for the parallelization I will leave it for later/somebody else.

Alright, no problem.

Looks good to me, thanks!

Issue #302, opened 2022-10-18
[Feature Request] Preset pen sizes


curbia:
It would be nice to have a few pen size presets available for people who switch between a few sizes often, say for headings and for content. Other handwritten note-taking apps offer this feature, including Xournal++ (pictured below,) and OneNote.

![image](https://user-images.githubusercontent.com/87792521/196480541-a5fed03f-d152-4e64-afbc-40044145f92d.png)

The current interface is a bit difficult to quickly switch between sizes without many clicks, or converting my tablet back to keyboard mode so I can input the exact sizes.

PR #442, opened 2023-01-08
Preset pen sizes (#302)

D0V4HKIIN:
This is a quite buggy version of the pen size presets discussed in #302. The ui doesn't update when selecting a preset and it doesn't save the custom sizes of the presets for some mysterious reason i cannot figure out.


D0V4HKIIN:
I figured out why it wasn't changing the sizes. I still don't know how to update the ui when pressing the buttons.

D0V4HKIIN:
Had some issues with borrowing but it now updates the ui when clicking the preset buttons

D0V4HKIIN:
This breaks the colors in the file format.

flxzt:
I think it would be better if the presets only are saved in the widget, not in the pen config. We don't really need to touch anything outside of rnote-ui, and toggling the buttons simply apply the width to the stroke_width fields. Similar to the colorpicker, where the palette is also only widget state

D0V4HKIIN:
I think the same actually. I didn't realize that the whole pen config was getting written to the file and keeping this in the ui would be much better.

What should we do about the default sizes?

D0V4HKIIN:
I don't know how to handle the stroke widths so that they can be changed using the spin button and also be persistent when reopening the app. @flxzt maybe you would know?

flxzt:
Actually, I believe it would be better to do something similar to the colorpicker, where you have a palette where you can modify the selected with the spinbutton. I'd subclass `ToggleButton` and add a `width` property, and have a composite widget that bundles everything with nice API like `ColorPicker`. It would also be nice if the appearance of the button reflects the width, probably by scaling the icon up and down or doing that with css.

D0V4HKIIN:
Do you mean to dynamically scale the icon up and down so that it changes with the width property? I'm not certain that it would be

flxzt:
I hope you don't mind, but I took it on myself to implement this with #456 . Part of it was, that I thought it was tricky with the custom widget that draws the width preview, and I want to get this into the next release

D0V4HKIIN:
closed because #456 implements the feature already

PR #446, opened 2023-01-12
skip saving when already in progress

### LINKED ISSUE #432: Pressing the save button rapidly after having made changes causes a crash
**Describe the bug**
After having made a stroke, tapping the save button rapidly crashes the program. This causes data loss.

**To Reproduce**
1. Make a stroke with a stylus
2. rapidly press the save button a couple times
3. Save icon says highlighted, program stops responding, then hard crashes.

**Expected behavior**
Should either rate limit saving or only allow saving once at a time, so saving many times at once does not crash the program.

** Console Output **
( remind me to do this later, I don't have time for it rn)

 - OS: Fedora 37
 - Installation source: Flatpak
 - Desktop Environment: Gnome 43.2

**Additional context**
I just press the save button a couple times without thinking, not sure how many times you need to do it exactly to cause a crash.

flxzt:
This skips saving the document when it is alerady in progress, and should fix #432 .

@Kneemund Just to double-check, does this look like a reasonable fix to you by using `expect_write` here?

### DIFF
--- a/rnote-ui/src/canvas/imexport.rs
+++ b/rnote-ui/src/canvas/imexport.rs
@@ -175,6 +175,12 @@ impl RnoteCanvas {
     }
 
     pub(crate) async fn save_document_to_file(&self, file: &gio::File) -> anyhow::Result<()> {
+        // skip saving when it is already in progress
+        if self.output_file_expect_write() {
+            log::debug!("saving file already in progress.");
+            return Ok(());
+        }
+
         let basename = file.basename().ok_or_else(|| {
             anyhow::anyhow!(
                 "save_document_to_file() failed, could not retreive basename for file: {file:?}"
--- a/rnote-ui/src/canvas/imexport.rs
+++ b/rnote-ui/src/canvas/imexport.rs
@@ -187,12 +187,20 @@ impl RnoteCanvas {
             )
         })?;
 
-        let rnote_bytes_receiver = self
+        self.set_output_file_expect_write(true);
+
+        let rnote_bytes_receiver = match self
             .engine()
             .borrow()
-            .save_as_rnote_bytes(basename.to_string_lossy().to_string())?;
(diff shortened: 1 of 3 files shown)


Kneemund:
If this works and the flag is *always* properly cleared, then this should be good. If - for some reason - rnote doesn't catch the file event caused by saving, then that would be pretty fatal. I can't think of any circumstance where this would happen (except on Windows maybe?) though, assuming that the write was successful. 

~~BTW, if something fails within the saving process, the flag will stay set forever.~~ NVM, you just fixed that :P

I also don't know how you could circumvent this possibility (apart from e.g. clearing the flag on regular intervals, but that's a shitty solution). Also, we should probably differentiate between "success", "noop" and "error" and only show the toast that the document was saved successfully on success.

flxzt:
> EDIT: create_replace_file_future() can theoretically still fail at multiple points, right?  

Yeah that was already an issue before, I am looking into better error handling so that we can clear `expect_write` appropriately. Unfortunately the file functions in gtk-rs only give back a generic glib::Error. We could either switch to rust-native async file operations to be able to [match on ErrorKind](https://docs.rs/smol/1.3.0/smol/io/enum.ErrorKind.html) (would need to check if these work with glib's executor), but maybe we should always clear the flag when anything fails.

> Also, we should probably differentiate between "success", "noop" and "error"

fixed that, it now returns Result<bool> and is matching on that.

Kneemund:
> maybe we should always clear the flag when anything fails.

Clearing it too often shouldn't really hurt in this case. The worst thing that could happen is the reload toast appearing if the file was overwritten despite the error (vs. being unable to save until you restart the app).

I can't spot anything else that could cause any trouble

flxzt:
Those are my thoughts too.

Okay, then lets merge this as is. Thanks for reviewing! I think it is good practice to do that

Kneemund:
No problem! Yeah

PR #447, opened 2023-01-12
CI: include rustc hash in cache key

Kneemund:
The CI cache is currently not rebuilt when the rustc version changes. This explains the unusually long durations of the last three workflows:

Unbenannt

This PR adds the rustc version hash provided by the [`actions-rs/toolchain` step](https://github.com/actions-rs/toolchain#Outputs) to the cache key and restore key. The latter is done because the whole app has to be rebuilt, regardless of the old cache, so we can save the time it takes to download it.

`cargo` was moved to the front of the cache (restore) key for consistency purposes (all variables are behind it now). It's a good idea to include it in this PR because all caches will be invalidated by it anyway :P


flxzt:
nice, thanks. I noticed the long duration, but didn't have the bandwidth to look into it

PR #450, opened 2023-01-14
fix eraser splitting stroke only once

### LINKED ISSUE #438: Split eraser only erasing one segment at a time
**Describe the bug**
The split eraser only erases one segment at a time which makes that when deleting a long line with a big eraser tool, you can see the line slowly getting erased.

**To Reproduce**
1. Draw a long line
2. Select the eraser tool and set it to split eraser and set it's size to something big like 400
3. Erase the line
4. Observe the line getting deleted slowly

**Expected behavior**
I expect the line to be deleted immediately instead of having to wait for every single stroke to get deleted.

https://user-images.githubusercontent.com/43830391/211166008-e00f63b9-aff9-420b-bfea-72bfa3df4980.mp4

flxzt:
this should fix #438 .

### DIFF
--- a/rnote-compose/src/penpath/mod.rs
+++ b/rnote-compose/src/penpath/mod.rs
@@ -65,7 +65,7 @@ impl ShapeBehaviour for PenPath {
     }
 
     fn hitboxes(&self) -> Vec<Aabb> {
-        self.hitboxes_priv()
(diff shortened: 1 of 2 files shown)


flxzt:
tested it a bit, should be fine

Issue #434, opened 2023-01-04
Popup warning about file change, prompting for a reload, after saving 


wiryfuture:
**Describe the bug**
After saving your notes, a popup appears at the bottom, telling you that a change was made and prompts you to reload. This makes sense if it was modified by another instance of rnote, but is unnecessary when you have just saved it in the same program.

**To Reproduce**
1. Create a blank note and save it.
2. Make any change on the note.
3. Save the note.
4. Popup prompting a reload appears.

**Expected behavior**
Program does not prompt a reload when it is what initiated the file changing on disk.

**Desktop (please complete the following information):**
 - OS: Fedora 37
 - Installation source: Flatpak
 - Desktop Environment: Gnome 43.2

**Additional context**
A simple fix would be storing the hash of the rnotes file after saving it, then having the function that prompts a reload to only call if the hash of the rnotes file has changed since the last save (or if no save hash exists because the notes haven't been saved since opening).

Kneemund:
Which version of rnote are you using?

wiryfuture:
> Which version of rnote are you using?

0.5.12

flxzt:
@wiryfuture are you getting this only when you are saving a new file or is this also the case after loading in an existing one?

wiryfuture:
> @wiryfuture are you getting this only when you are saving a new file or is this also the case after loading in an existing one , modifying and then saving it?

both - it always shows.

flxzt:
I think I know what was happening, and I am fairly confident that #452 will fix this

flxzt:
@wiryfuture could you try the newest beta that includes the fix?

wiryfuture:
> @wiryfuture could you try the newest beta that includes the fix?

Is the build available on flathub by any chance? I can't remember how to get betas off there

flxzt:
yes, it is published in the flathub beta repo. Check here: https://discourse.flathub.org/t/how-to-use-flathub-beta/2111 

The beta should then appear as an entry in software, or ask you which repo when installing it via cli

wiryfuture:
> @wiryfuture could you try the newest beta that includes the fix?

seems to

flxzt:
Could you run the app with `RUST_LOG=rnote=debug flatpak run com.github.flxzt.rnote` and post the logs when it happens with the autosave?

wiryfuture:
[...]

PR #452, opened 2023-01-15
Fix output monitor missing events for new or changed files

flxzt:
This fixes multiple issues found related to the output monitor:

---

There was an ordering issue where the output monitor was recreated (by `set_output_file()` ) **after** actually saving to disk, causing sometimes missed file events when initially saving a new or a changed file, which then causes missing clearing `expect_write` so we end up in the state where saving is not possible anymore.

#434 *might* be related, but I can't think the sequence that would trigger that right now. Would need testing and more debugging.

Edit: we might want to do that before merging, because if we don't and this is indeed the fix, we won't have the certainty to close that issue

( 1a401e98ddd1a12b4d90b28b7ec94c93956e90a8 )

---

the output monitor is now cleared in `reinstall_output_file_monitor()` when output_file is none

( a1378b5e6d1426a622593d8d8e2210117f510e5b )

---

fixed panics when clicking `reload` in the reload toast after closing the tab that emitted it

( ced60af3932bbacff9c5486e28aeb0dab9030619 )

---

Also: did some code clean up like early returns 

(9dd0635064eb690e461d42a5630d4600318fbde1 )

---

@Kneemund thoughts?

### DIFF
--- a/rnote-ui/src/canvas/imexport.rs
+++ b/rnote-ui/src/canvas/imexport.rs
@@ -185,6 +185,12 @@ impl RnoteCanvas {
             return Ok(false);
         }
 
+        let file_path = file.path().ok_or_else(|| {
+            anyhow::anyhow!(
+                "save_document_to_file() failed, could not get a path for file: {file:?}"
+            )
+        })?;
+
         let basename = file.basename().ok_or_else(|| {
             anyhow::anyhow!(
                 "save_document_to_file() failed, could not retreive basename for file: {file:?}"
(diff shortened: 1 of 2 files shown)


flxzt:
sorry about the rebasing, I am done with them now.

There is also the issue that when the same file is closed and opened multiple times, the same amount of events are emitted. For example when open -> close -> open, and then modify and save, the `rename` from `.goutputstream` file event is emitted two times. I think the monitor is somehow still kept around from the old tab, but where and why I am not sure yet.

Kneemund:
Looks good to me. The explanations sound very plausible, but I can't reproduce/debug #434, so we'll have to wait and see or ask others to try this :P

Do we really need to store and explicitly disconnect `output_file_monitor_changed_handler`? I thought that cancelling the monitor did the same thing (which is always done after disconnecting the handler as far as I can tell).

flxzt:
> Do we really need to store and explicitly disconnect output_file_monitor_changed_handler?

Probably, but I wanted to be sure, I think it can't hurt.

I'll keep it open until at least a new beta / release.

Issue #232, opened 2022-07-28
Shape sub-options


defini-tiv:
**Is your feature request related to a problem? Please describe.**
Shape builders list has multiple entries for same shapes.
It is bigger than it needs to be and will get bigger with every additional option.

**Describe the solution you'd like**
Having only one options for every shape type (Line, Rectangle, Ellipse, Curve) and move other Options to a Shape sub-Options menu.
It could be where the constraints options are currently.  I think it fits there better  because it would be in closer relation to the shape you want to draw.

**Describe alternatives you've considered**
Maybe a fold-out sub menu for every shape builder in the list?

flxzt:
Fold-out rows seem like a good option to categorize the different shape builders when the list grows in the future.

Kneemund:
What do you think about something like this? The margins are very large and you can seemingly only adjust the top and bottom margin of the grid, and I'm not entirely sure about how this could be hooked up internally. You have to click next to the grid in order to actually select the shape. This seems to be one of the few options that work; I couldn't get collapsible (and selectable) rows to work *at all*.

[Bildschirmaufzeichnung vom 2022-12-24, 15-52-19.webm](https://user-images.githubusercontent.com/49598842/209441438-1ac0679b-93d2-4e19-88c3-6152d07ebfaa.webm)

flxzt:
That could be an alternative, though I don't like that "You have to click next to the grid in order to actually select the shape". And we'd need to think about when groups have more than three shapes (e.g. different splines ), that would grow the list horizontally. Though as long as the shape group name stays short it shouldn't be too much of a problem.

> I couldn't get collapsible (and selectable) rows to work at all  

Ah, thats unfortunate. That would have been the most elegant solution in my opinion.  

Regarding styling and margins, that can be tweaked by css with negative margin values, the same way the margins for the big icons in the sidebar are reduced.

The other solution would be to do a icon grid, "IconPicker" could be reused for that

Kneemund:
There are definitely flaws to this, yeah. Maybe you could move the shape selection using code when an element in the grid is clicked?

> Ah, thats unfortunate. That would have been the most elegant solution in my opinion.

It might be possible, but I just didn't manage to get it working. Every combination of UI elements I've tried had some flaws (visually and functionally). I'm not too familiar with GTK/libadwaita though.

> The other solution would be to do a icon grid, "IconPicker" could be reused for that

[...]

PR #455, opened 2023-01-18
grouped icon picker for selecting shape builder type

Kneemund:
This PR replaces the current row selection of the shape builder picker with an icon grid.

|dark|light|
|-----|-----|
|grafik|grafik|

TODO:
- [x] Tooltips. Because it's reusing `IconPicker`, those are currently equal to the icon name. They should either be removed, or replaced with the (localized) name of the shape. The latter would require redesigning/refactoring `IconPicker` a bit.
- [x] Grouping shapes.

Things to consider/discuss:
- The icons aren't grouped (see my [last comment](https://github.com/flxzt/rnote/issues/232#issuecomment-1381171755) in #232).
- The contrast is vastly different in dark and light mode. I'm not sure if that's intended or a bug.
- The current implementation using three match expressions is not very elegant, but I'm not sure if there's a better way.
- The existence and style of the label at the bottom. IMO it's almost required for clarity purposes. The text gets wrapped (and vertically expands the menu) if it's too long.


flxzt:
Thanks! Going with the icon grid is likely the best solution, especially when more builders are added over time.

> The icons aren't grouped

That could be an issue, though the added label for the picked icon already helps with that. It would nice, especially considering that more will be added, and it would also be useful in the cursor picker. But would need some more changes in iconpicker. Probably by refactoring to a vec of string lists with each a group title.

Depends on motivation, either would be fine with me.

> The contrast is vastly different in dark and light mode

I just checked, it seems simply to be the colors that libadwaita uses. The gridview uses `@view_bg_color` which for light theme is `#ffffff`, and the popover background should be `@popover_bg_color` which is also `#ffffff`. While in the dark theme they are different ( `#1e1e1e` and `#383838` ). see: 
https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/named-colors.html#window-colors

I don't have a preference, we could leave it unchanged or change the gridview style class to `background` which has a slightly different white in the light theme. It would mean that we lose the contrast when using the iconpicker in non-popover widgets though (considering future reuses). The popover bg color should not be changed, imo.

Regarding i18n, what we could also do: integrate the label into the iconpicker (and maybe expose it or a property to only show it optionally) and add a method to register an `Fn` which gets iterated over every icon name when the list is registered, and gives back the translated display string. We could use those then as both the tooltip and the label string. 

At least that would be my first idea, without doing any coding. Feel free to do it differently, if deemed impractical

Kneemund:
> It would be nice, especially considering that more entries will be added, and it would also be useful in the cursor picker. But would need some more changes in iconpicker. Probably by refactoring to a vec of string lists with each a group title.

I don't really know how to modify the UI components/what to use for this in general. `GtkGridView` doesn't seem to come with any grouping capabilities out of the box. I guess that you could have multiple `GtkGridView`s inside of the existing `GtkScrolledWindow`, with labels in-between, but I'm not sure if that's the best solution.

> I don't have a preference, we could leave it unchanged or change the gridview style class to background which has a slightly different white in the light theme. It would mean that we lose the contrast when using the iconpicker in non-popover widgets though (considering future reuses).

Alright. I'd leave it as-is then, because the light theme doesn't look out of place (just inconsistent compared to the dark theme). Changing the class to `background` could turn into a bigger/more annoying problem in the future, as you said.

> Regarding i18n, what we could also do: integrate the label into the iconpicker (and maybe expose it or a property to only show it optionally) and add a method to register an Fn which gets iterated over every icon name when the icon list is registered, and gives back the translated display string. We could use those then as both the tooltip and the label string.

I created a variation of `set_list()` called `set_list_localized()`, which does pretty much what you said. It's not exactly DRY, especially the beginning and end of those functions, but I'll most likely change that soon:tm: :P
I also removed the tooltips from the non-localized variant. IMO having nothing is better than having untranslated icon/file names, but I can revert this if you disagree.

Feel free to tell me if I should refactor/rename/improve any parts of the code.

flxzt:
> I guess that you could have multiple GtkGridViews inside of the existing GtkScrolledWindow, with labels in-between

that is how I would approach it too. It might be worth to check out [Icon Library](https://flathub.org/apps/details/org.gnome.design.IconLibrary), from a look with the inspector they are doing something similar (but with a flowbox instead of a gridview:

Bildschirmfoto vom 2023-01-23 18-07-33


> Alternatively, the fn that returns the i18n string could be set to |_| "".to_string() by default, and we could get rid of the other function entirely. Or make we could make that fn Option-al

Yeah I think we could make it `Option`onal. (Maybe rename the function parameter to `generate_display_name´ or something similar then. If we do `if let Some(...` I don't think it would have a runtime cost, would it? Either way, I don't think we need to worry about that ( I dare you to look up how the UI is updated when the active pen or tab changes :P Currently the ugliest part of the codebase )

rnote-ui/src/iconpicker.rs
```diff
@@ -195,30 +265,33 @@ impl IconPicker {
 
         // Factory
         let icon_factory = SignalListItemFactory::new();
-        icon_factory.connect_setup(move |_factory, list_item| {
+        icon_factory.connect_bind(move |_factory, list_item| {
```
flxzt, line 268: Was there a specific issue / reason that this is changed?

rnote-ui/src/penssidebar/shaperpage.rs
```diff
@@ -313,23 +285,33 @@ impl ShaperPage {
         }));
 
         // shape builder type
-        imp.shapebuildertype_listbox.connect_row_selected(
-            clone!(@weak self as shaperpage, @weak appwindow => move |_, _| {
+        imp.shapebuildertype_picker.set_list_localized(
+            StringList::new(globals::SHAPEBUILDERTYPE_ICONS_LIST),
+            |icon_name| match icon_name {
```
flxzt, line 290: we could move this to the type, like ´from_icon_name()` and `to_icon_name()`

rnote-ui/src/iconpicker.rs
Kneemund: Yes, when connecting to the setup signal, `list_item.item()` is always `None`. This doesn't happen when you use bind (`ChangePenStyleListFactory` uses this too BTW).

Kneemund:
Alright. I'll take a look at Icon Library, I was actually looking for something like that but e.g. the emoji picker doesn't use labels.

rnote-ui/src/penssidebar/shaperpage.rs
```diff
@@ -313,23 +285,33 @@ impl ShaperPage {
...
+            },
+        );
+
+        imp.shapebuildertype_picker.connect_notify_local(
+            Some("picked"),
+            clone!(@weak self as shaperpage, @weak appwindow => move |picker, _| {
                 if let Some(buildertype) = shaperpage.shapebuildertype() {
                     appwindow.active_tab().canvas().engine().borrow_mut().pens_config.shaper_config.builder_type = buildertype;
-
-                    match buildertype {
```
flxzt, line 321: the comment was supposed to be here:

we could move this to the type, like `from_icon_name()` and `to_icon_name()`

rnote-ui/src/penssidebar/shaperpage.rs
flxzt, line 290: but here too actually, so we can match on the enum

Kneemund:
I created a proof of concept (i.e. absolutely horrific code, I'm surprised that it works): 

https://user-images.githubusercontent.com/49598842/214447875-3f4fec9c-80f0-4120-97e5-8f8f7751e91f.mp4

We should definitely put this into a new template/widget/module though, because it's very different to the current (ungrouped) `IconPicker`.

ATM, it uses `FlowBox`es instead of `GridView`s too (like Icon Library), because `GridView` doesn't shrink automatically depending on its children.

rnote-ui/src/iconpicker.rs
flxzt: Hm, I was using expressions here which are supposed to be in `setup()`, I believe that was correct. (see: https://gtk-rs.org/gtk4-rs/stable/latest/book/list_widgets.html ) Not sure why you replaced that. But It doesn't matter when this is replaced by flowbox

Kneemund:
Just pushed my changes. 
I reverted everything related to `IconPicker` for now, if you want localized tooltips for other uses (e.g. workspace icon picker), I can revert that *again* :P

I created a new Widget called `GroupedIconPicker`, which holds a `GroupedIconPickerGroup` created using `GroupedIconPickerGroupData`. The names are wack, let me know if you can think of something better. Things *should* be working.

rnote-ui/src/iconpicker.rs
flxzt: You could have used an expression as well, that is usually how it should be done when using models with the *View list widgets, because the models don't map one to one to the widgets, they get recycled. One does get away with it when the model is never modified or reordered though.

Kneemund: Alright, thanks for the info.

flxzt:
Looks nice!  

Being able to (optionally) have custom display names would be nice for the regular iconpickers too, I liked that addition. :)  

> The names are wack

Eh, I don't know any better ones either  

> we might want to have a minimum width for the widget

I'd go with `250` or `300` (like the other popovers) as `width-request`, to anticipate future builder additions and that also gives more space for the display bottom label. Regarding that I think the label should have constant height to have room for 2 lines of text

> contrast issues with the white theme became more "serious

ah true, then lets replace it with `background` in the listbox?

Some other things I noticed:

- I think it would be nice when the icons indicate on hover
- small style tweaks: we could reduce margins around the picker and around the groups, and the rounded-corners with the groups-listbox are not correct
- Maybe the groups should use plural?
- Lets create a folder for the two groupediconpicker source files
- UI files are missing in POTFILES, would need to be added there for translations

Kneemund:
Okay, I'll revert that revert then :P

> I think the label should have constant height to have room for 2 lines of text

Is there an elegant way to do that? You could hardcode the `height-request` and ellipsize the text after 2 lines, but I don't know if we should. The text size is probably not fixed. Using GtkTextView might work, but that seems a bit too powerful for this task.

> I think it would be nice when the icons indicate on hover

It seems like we have to hardcode that using CSS. GtkFlowBox supports selections but doesn't have any hover effects. I'd take the [transition](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/stylesheet/_common.scss#L1) and [colors](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/stylesheet/_colors.scss#L62-L66) from libadwaita; Icon Library does this too. The current icon picker (i.e. GtkGridView) doesn't have any transitions, so it would be somewhat more consistent when compared to other buttons, unless you're using themes (even then, it should be fine - the colors are dynamic).

> Maybe the groups should use plural?

I don't have a strong opinion on this, but I thought that using singular everywhere would make it more universal if we have groups with only one element in the future. On the other hand, we'll probably throw all of those into the Miscellaneous group anyway.

> Lets create a folder for the two groupediconpicker source files

Should we create one for the .ui files/templates too? Or only for the source code?

flxzt:
> Only the tooltip part, or the label too?

Both actually, I think the cursor picker would benefit from the label, while it would be unnecessary for the workspacerow iconpicker. 

> Is there an elegant way to do that

Hm, I am not sure either. there is the `lines` property but that only applies as a maximum. I think it can done with css with `min-height: <val>em;` , though I don't know the exact value right now. Would need to include the label font-size and line spacing + padding. Might be easier to manually set all of those in our stylesheet then, to make sure they match

> I'd take the [transition](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/stylesheet/_common.scss#L1) and [colors](https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/stylesheet/_colors.scss#L62-L66) from libadwaita; 

Sounds good!

btw. is there a reason for `focusable>False` and `can-focus>False` in the groups ? Being able to tab through the groups and icons would be better, unless it breaks something

> Do we want to reduce the margins of all of these popovers?

we can do that, 6px margin is probably enough. Also yes to fixing the title inconsistency, I think either size would be fine

> I don't have a strong opinion on this, but I thought that using singular everywhere would make it more universal if we have groups with only one element in the future. On the other hand, we'll probably throw all of those into the Miscellaneous group anyway.

I think so too for single shapes that don't fit anywhere else, so plural is better imho  

> Should we create one for the .ui files/templates too

Would make sense, yeah

Kneemund:
> btw. is there a reason for focusable>False and can-focus>False in the groups ? Being able to tab through the groups and icons would be better, unless it breaks something

The `GtkListBoxRow`s can be made focusable. 
But the focus moves the "visual selection"/highlight of `GtkFlowBox`es, and I haven't found any way around that apart from disabling the focus altogether. And it doesn't make much sense to be able to tab through the groups if you can't tab through the items.

https://user-images.githubusercontent.com/49598842/215291975-ff9b59f1-aaeb-4655-a3a0-647f33fcf583.mp4

Let me know if you have any ideas on how you could fix this. Maybe `GtkFlowBox` just wasn't a good choice to begin with, but I don't know any other viable widget that automatically shrinks to the size of its children.

flxzt:
I see. By changing `can-focus>false` to `focusable>false` on the flowbox it would be possible to only tab through the items, but this issue still remains. Not sure how to solve it either, and probably also not very important. But I think flowbox is a good choice as a widget

Kneemund:
Yeah. Should I make the `GtkListBoxRow`s focusable anyway, or leave them as they are?

Another style question:
When reducing the margins to 6px, the label at the bottom of the shape picker is considerably larger than the title section. It looks a bit odd, especially when there's only one line (which is the case 100% of the time ATM). I can't take a screenshot, because the popover captures all mouse and keyboard input for some reason. I can push the changes I've made so far if you want to take a look yourself.

flxzt:
> Should I make the GtkListBoxRows focusable anyway, or leave them as they are

lets leave it then I think

> When reducing the margins to 6px, the label at the bottom of the shape picker is considerably larger than the title section

Maybe the label should also be `subtitle`, but not sure.

> I can push the changes I've made so far if you want to take a look yourself.

Sounds good :)

Kneemund:
Alright, I think I've got everything you mentioned, apart from restoring the changes to `IconPicker`. The style probably still needs some tweaks, feel free to tell me, or, if you have something in mind, you can just push that if you want (I *think* you should be able to).

Which reminds me - I don't know how you can retrieve the icon name during `connect_setup`, which is needed for generating the display names. If I understood `setup` and expressions correctly, then that's not possible. Except if we use something different than `StringList` as the model, which also stores the display name.

flxzt:
You know what, lets keep the changes to iconpicker separate from this PR.

> The style probably still needs some tweaks, feel free to tell me, or, if you have something in mind, you can just push that if you want (I think you should be able to).

Okay, I might do that. I am not always 100% sure with styling either, so is always subject to change anyway

Kneemund:
Alright, I'll rebase this then and mark it as ready for review :P
Should I create a new PR for the `IconPicker` changes? The relevant parts from this PR are in [this commit](https://github.com/flxzt/rnote/pull/455/commits/673722a34a718742684480342433865f431b550a), but we might end up not using much of that.

flxzt:
> Should I create a new PR for the IconPicker changes

Sure :)

Okay, I think this looks all very good.
I removed the `groupdata` re-export, for consistency with the rest of the code base

Regarding styling:
- I changed the brush options popover margins to 6px as well, to make it consistent.
- Tweaked the sizes a little bit.

flxzt:
This is a great addition / improvement, thanks :)

Kneemund:
Alright, no problem :)

PR #456, opened 2023-01-18
Stroke width picker with preset palette

### LINKED ISSUE #302: [Feature Request] Preset pen sizes

flxzt:
This introduces a stroke width palette to be able to save commonly used widths in presets. The widths are previewed as a circle with radius as stroke_width * 0.5.

Bildschirmfoto vom 2023-01-18 23-07-26

To be able to display all widths from 0.1 to 500.0, a conversion function is used that is asymptotic to a specified max size. For small widths this scales approximately linear, which should be the most common case.

fixes #302 when merged. @D0V4HKIIN already has a similar draft PR, but I thought it was easier to implement it myself, because the custom widget is tricky when not already familiar with gtk4 subclassing.

### DIFF
--- a/rnote-ui/data/app.gschema.xml.in
+++ b/rnote-ui/data/app.gschema.xml.in
@@ -34,23 +34,23 @@
       <summary>the first brush stroke palette width</summary>
     </key>
     <key name="brush-width-2" type="d">
-      <default>2.0</default>
+      <default>6.0</default>
       <summary>the second brush stroke palette width</summary>
     </key>
     <key name="brush-width-3" type="d">
-      <default>6.0</default>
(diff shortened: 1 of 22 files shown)


D0V4HKIIN:
nice, i do have some changes i didn't commit because i got some weird allocation errors with the subclass. Will close my pr.

Issue #41, opened 2022-01-17
Is there any plans to make it work on MacOS?


backtozero:
Since GTK4 got some improvement in supporting Macs

ovidius72:
I second the request, thanks.

eli-schwartz:
It's possible to check in a copy of the needed its rules files to a local directory, e.g. `resources/its/`, and pass that directory as `data_dirs: 'its'` in meson.build so that the local ITS rules files are also considered by gettext.

y-mx-b:
I attempted a build after installing the dependencies @nilbot had listed, and after running `meson compile -C _mesonbuild` I ended up with the following two messages at the very end
```
[14/16] Generating rnote-ui/data/com.github.flxzt.rnote.desktop with a custom command
/opt/homebrew/bin/msgfmt: {Absolute path to rnote}/po/LINGUAS does not exist
[15/16] Generating rnote-ui/data/com.github.flxzt.rnote.metainfo.xml with a custom command
/opt/homebrew/bin/msgfmt: {Absolute path to rnote}/po/LINGUAS does not exist
```
with the absolute path to rnote being `/Users/{HOME}/.../rnote`. This seems to be different to the `cannot locate ITS rules for resources/app.metainfo.xml.in` error message, though I'm not entirely sure what's going wrong here either.

Attempting to run the binary with `./_mesonbuild/target/release/rnote` would result in the following error message:
```
thread 'main' panicked at 'Could not load gresource file: Error { domain: g-file-error-quark, code: 4, message: "Failed to open file “/usr/share/rnote/resources.gresource”: open() failed: No such file or directory" }', rnote-ui/src/app/mod.rs:72:18
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```
I'm not sure what a `LINGUAS` file is supposed to be (I have 0 knowledge regarding `msgfmt` to begin with), and I don't know what a `gresource` file is supposed to be, so I'm quite puzzled by this.

Unless someone more knowledgeable happens to comes along, I guess I'll have to Google-fu this some other time.

eli-schwartz:
https://github.com/flxzt/rnote/blob/fab9abe80eef0ba015e1160e8e1a171c8f2b3829/meson.build#L46

There is a LINGUAS file here: https://github.com/flxzt/rnote/blob/main/rnote-ui/po/LINGUAS

It's just a text file listing the available languages that have translations -- and comes in the same directory as the translations themselves.

[...]

PR #470, opened 2023-01-24
add support for macos app bundles / dirs retrieval at runtime

flxzt:
This add some configuration to make app bundles in macos work. addresses #41


dehesselle:
Hi @flxzt ,
I've picked up work on the app again. For some reason I've had mixed results with this branch which I cannot really explain/reproduce, so before getting too deep into it, can I ask you to rebase this on current master?

One thing is already clear though: I forgot that there is one more change necessary. This needs to be made dynamic as well:
https://github.com/flxzt/rnote/blob/777fdc0132b2e5bed0df4ff1f70e783936e782a1/rnote-ui/src/app/mod.rs#L147
The resources file (`resources.gresource`) needs to be loaded from a location relative to the application binary as well. Since we're already setting `XDG_DATA_DIRS`, you can use that, e.g. `XDG_DATA_DIRS/rnote/resources.gresource`.
I am aware that this suggestion is risky as `XDG_DATA_DIRS` is supposed to contain a list of directories, but since that is something we control at this point (i.e. a list with one element), it would be safe. (Alternatively you'd have to iterate through the list of directories and try everyone of them to locate the `resources.gresource` file.)

rnote-ui/src/main.rs
```diff
@@ -77,3 +82,32 @@ fn setup_windows_env() -> anyhow::Result<()> {
...
+            } else {
+                false
+            }
+        })
+    {
+        std::env::set_var("XDG_DATA_DIRS", &current_dir.join("/../Resources/share"));
+        std::env::set_var(
+            "GDK_PIXBUF_MODULE_FILE",
+            current_dir.join(PathBuf::from(
+                "/../Resources/lib/gdk-pixbuf-2.0/2.10.0/loaders/loaders.cache",
```
dehesselle, line 108: This is the wrong path. Lose the `loaders` subdirectory:
```
/../Resources/lib/gdk-pixbuf-2.0/2.10.0/loaders.cache
```

rnote-ui/src/main.rs
```diff
@@ -77,3 +82,32 @@ fn setup_windows_env() -> anyhow::Result<()> {
...
+        .zip(current_dir.components().skip(1))
+        .any(|(a, b)| {
+            if let (Component::Normal(a), Component::Normal(b)) = (a, b) {
+                a == OsStr::new("Contents") && b == OsStr::new("MacOS")
+            } else {
+                false
+            }
+        })
+    {
+        std::env::set_var("XDG_DATA_DIRS", &current_dir.join("/../Resources/share"));
```
dehesselle, line 104: Apologies if my assumption is incorrect: I think `current_dir` does not contain the directory name but still the full path to the binary at this point. I'd expect a `skip()` here same as above.

rnote-ui/src/main.rs
flxzt: ah right, probably copy-pasted the path from the win setup here

rnote-ui/src/main.rs
flxzt: no, you were right.

flxzt:
okay, *I think* I addressed the issues.

What will the result be of your efforts, a script that can be executed on macos to create the bundle?

dehesselle:
Great, I'll test it again later today.

The result will be this: https://gitlab.com/dehesselle/rnote_macos
(Still work in progress, see https://gitlab.com/dehesselle/zim_macos for a polished example.)

I'm using GitLab as I have CI infrastructure there (and GitHub's CI  is sadly rather unstable for me).

dehesselle:
Almost there. This is the output I'm getting (I have added two `println!()` statements for `data_dir` and `lib_dir`):
```
data_dir=/../Resources/share
lib_dir=/../Resources/lib

(Rnote:51502): GLib-GIO-CRITICAL **: 23:23:31.465: g_settings_schema_source_lookup: assertion 'source != NULL' failed

(Rnote:51502): GLib-GIO-CRITICAL **: 23:23:31.465: g_settings_schema_source_lookup: assertion 'source != NULL' failed
thread 'main' panicked at 'Could not load gresource file: Error { domain: g-file-error-quark, code: 4, message: "Failed to open file “/../Resources/share/rnote/resources.gresource”: open() failed: No such file or directory" }', rnote-ui/src/app/mod.rs:155:14
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

flxzt:
argh, I don't know why that slash landed there. The executable name doesn't appear in your print output, but it is apparently lost when joining with the (invalid) absolute path (see [here](https://github.com/rust-lang/rust/issues/16507) )

dehesselle:
Sorry you have to check this once more, now it wants to open
```
MacOS/../Resources/share/rnote/resources.gresource
```
This needs to be the full canonical path.

flxzt:
The mention of "dir name" [here](https://github.com/flxzt/rnote/issues/41#issuecomment-1401114485) and the comment [here](https://github.com/flxzt/rnote/pull/470#discussion_r1130252110) was confusing.

dehesselle:
Apologies for the confusion, the intention was to "think like coreutils", e.g. `dirname /whatever/path/to/some/random/binary`

flxzt:
no worries, I also read it wrong in the sense of "folder name" .

dehesselle:
Awesome, Rnote works now out of the box as macOS app! 🎉 

I'll have to do some more polishing on my end, but you can basically merge this to master and with 0.5.17 (or whatever comes next) you have a macOS app.

One more thing though, which is either for me to figure out or it might be related to what we've been doing here: can you tell me from where the logo in the about screen is  supposed to come from?
image

flxzt:
Thanks a lot for doing this!

I want to mention and link to it in the README & on the website, do you have any suggestions for that / a formulation for macos users how to install / use it?

> can you tell me from where the logo in the about screen is supposed to come from?

That should be the icon which gets installed to `<data dir>/icons/hicolor/scalable/apps/com.github.flxzt.rnote.svg`

dehesselle:
You can link to the latest release by using
```
https://gitlab.com/dehesselle/rnote_macos/-/releases/permalink/latest
```
This doesn't work yet since I haven't created a release yet. I can create a release this weekend if you are good to go on 0.5.17.

I'll put everything that people need to know in the readme, so for that you can link to the GitLab repo (see [zim_macos](https://gitlab.com/dehesselle/zim_macos) as example).

You can add it to your readme/website as you see fit. My first guess is to add it to "Installation" below Flathub? "Rnote is available for macOS here:".

dehesselle:
@flxzt final addendum: first release is live!

PR #480, opened 2023-01-29
Refactor select all event

TheSmuks:
This adds a function to reduce code repetition when handling "select all" event.


flxzt:
cleaning up the code always sounds good to me, thanks! only needs a `cargo fmt`

Issue #478, opened 2023-01-29
Fix header capitalization


flxzt:
Currently titles follow sentence capitalization most of the time. We should change them to using header capitalization as suggested in the [Gnome HIG](https://developer.gnome.org/hig/guidelines/writing-style.html)

This will modify many strings, so we should change this right after a release so we can give translators as much time as possible to update their translations.

PR #485, opened 2023-01-31
IconPicker: add i18n tooltips and selection label

Kneemund:
As discussed in #455, this PR adds localized tooltips and selection labels to `IconPicker`, specifically to
* the workspace icon picker (without selection label) and
* the cursor picker (with selection label).

Things to consider/discuss:
* `globals.rs` might not be the best place to store icon name lists and its i18n functions.
* A combination of `connect_setup` for creating the widgets and `connect_bind` for adding the tooltips is used, the latter might be problematic.
* Adds quite a bit of translation work (for not that much gain). We might want to merge this *after* a release.
* Some of the icons have names that don't really fit. For now, I've decided to use a (IMO) more fitting name than what the icon name suggests (e.g. Hammer instead of Build, Band-Aid instead of Heal, Projector instead of Present, ...).


flxzt:
> globals.rs might not be the best place to store icon name lists and its i18n functions

Did you have a specific location in mind?

> Adds quite a bit of translation work (for not that much gain). We might want to merge this after a release.

Yeah, probably should sync this with #478 .

> A combination of connect_setup for creating the widgets and connect_bind for adding the tooltips is used, the latter might be problematic.

It should be implementable by using [chain_closure()](https://gtk-rs.org/gtk4-rs/stable/0.5/docs/gtk4/struct.Expression.html#method.chain_closure), check the workspacebrowser where this is also used.

> Some of the icons have names that don't really fit. For now, I've decided to use a (IMO) more fitting name than what the icon name suggests

sure sounds good, but I think we should change the file names as well. Could be confusing looking them up with an additional indirection

We could also add some if you have any suggestions

Kneemund:
Right, just pushed the stuff that you mentioned. Using `chain_closure` is certainly a lot more elegant and simple. I should probably take another look at expressions.

> Do you have a specific location in mind? The other place I would expect it to be would be in the source files where the widgets are located

Yeah, that's what I had in mind too, but `globals.rs` is probably fine. I just thought that it might be a weird place, because each icon list is only used in a single other file. And it only contained constant variables prior to this PR.

> We could also add some if you think some useful ones are missing

I couldn't spot any useful additions while scrolling through Icon Library, and I didn't miss anything so far.
Well, except for another math icon (but that'd be a bit redundant and I also don't know what to use).

flxzt:
> Using chain_closure is certainly a lot more elegant and simple

Yeah at least as far as Gtk's stuff can fit into rust. The bindings are great though, and recently they added a lot of more ergonomic API. Soonish we'll be able to declare properties with a macro, that will be really nice :)

> Yeah, that's what I had in mind too, but globals.rs is probably fine

No we can absolutely do that, sounds good!

Alright, the rest looks good. I'll go through the titles and prepare a PR for that. After I do a release soon(ish) I'll merge both into main and then update the po template

Kneemund:
Using macros for properties sounds pretty convenient, yeah :P

> The other place I would expect it to be would be in the source files where the widgets are located

Alright, I moved the data from `globals.rs` to the bottom of the files that use `IconPicker` (hoping that I understood you correctly). I also moved the grouped icon picker data into separate functions at the bottom of `shaperpage.rs`  for consistency.

rnote-ui/src/penssidebar/shaperpage.rs
```diff
@@ -386,3 +340,55 @@ impl ShaperPage {
         );
     }
 }
+
+fn shape_builder_type_icons_get_groups() -> Vec<GroupedIconPickerGroupData> {
```
flxzt, line 344: why not use a static?

Kneemund: That doesn't seem to work because `StringList` doesn't implement Send/Sync, unless I'm missing something:

```
`*mut c_void` cannot be sent between threads safely
the trait `std::marker::Send` is not implemented for `*mut c_void`
required for `TypedObjectRef<*mut c_void, ()>` to implement `Sync`
[...]
main.rs(1, 1): required because it appears within the type `gtk4::StringList`
```

`GroupedIconPickerGroupData` could use e.g. `Vec<String>` instead of `StringList` (and `GroupedIconPicker#set_groups()` could construct the `StringList`s on the fly), but I'm not sure if that's better. 
Also, the group names should get re-evaluated every time because they use `gettext`, right?

flxzt: https://docs.rs/once_cell/latest/once_cell/unsync/struct.Lazy.html maybe? I think the vec's values get written only once on first use

Kneemund: I can't get that to work either; you have to make it `const` (`static` throws the same error that I posted above), but if you do that, the data is moved after accessing it

flxzt: okay, I'm not sure what the issue is then. In the end it doesn't really matter anyway because the setup is only called once but I personally like making it Vec<String> and still use lazy. But it's up to you

Kneemund: I'm also fine with either option, but I'd rather leave it as-is :P

If we store the icon names as `Vec<String>`, then we'd have to iterate all elements in order to construct the `StringList`: `StringList::new(&icon_names.iter().map(String::as_str).collect::<Vec<&str>>())`

flxzt: ok!

flxzt:
Okay, looks good thanks!

PR #506, opened 2023-02-10
CI: Check file compatibility

### LINKED ISSUE #378: Set up CI checks to ensure file format compatibility
Currently one needs to check manually if a change made in the code broke the file compatibility. When we move towards a more stable file format and being a more serious application, we definitely  need some kind of automated tests to ensure the file format does not break unintentionally.

I am not an expert in github actions, so if anyone wants to help here I would appreciate it.

flxzt:
This PR adds two things:

- a CLI command to check if rnote files can be loaded as a EngineSnapshot. If it fails, the file format broke.

- a CI job which executes this new command against the current screenshot pages in `misc/screenshots-pages`. They should have all stroke types in them so I think we won't miss anything. But it might be worth considering adding a dedicated "compatibility-check.rnote" file.

Because this needs a compiled cli binary I also added `compile` and `install` jobs, that makes the `cargo check`s a bit redundant. But because cargo reuses the compilation artifacts and `check` fails faster than a complete compile, it might be okay to keep them.

In the very rare occasion that we intentionally break file format compatibility, we can simply ignore this test because it fails last

@Kneemund since this is CI related, I'd appreciate your feedback here :)

fixes #378 when merged


Kneemund:
> Because this needs a compiled cli binary I also added compile and install jobs, that makes the cargo checks a bit redundant. But because cargo reuses the compilation artifacts and check fails faster than a complete compile, it might be okay to keep them.

"Compile" doesn't seem to use *any* cache ATM:

grafik

Keeping the `cargo check`s would make the job fail faster, yeah, but I'd assume that the checks will be successful in the majority of runs. So keeping them would probably cause more overhead. 

Also, I'd rename the "Compile" and "Install" steps to "Compile CLI"/"Install CLI" (assuming that they don't install and compile both).

flxzt:
> "Compile" doesn't seem to use any cache ATM

Looking at the logs, it apparently uses the cache from main and doesn't commit afterwards. It also does more work than usual in the `check` jobs.

The "Post Cache" log:
```
 Cache hit occurred on the primary key cargo-Linux-d5a82bbd2-f37876929ec6f99a6838796b9fa55a903227f9b08bc3a7f5b3812e4e2b27130a, not saving cache
```

Could the reason be that changing the jobs isn't reflected in the cache-key so it always reuses the previous one? Should the hash from the ci.yml file be added to the cache-key?

> Keeping the cargo checks would make the job fail faster, yeah, but I'd assume that the checks will be successful in the majority of runs. So keeping them would probably cause more overhead.

Thats true. I wanted to see how fast it is when reusing all cache, and depending on that keep or remove it

> Also, I'd rename the "Compile" and "Install" steps to "Compile CLI"/"Install CLI" (assuming that they don't install and compile both).

Unfortunately it does both, the app / cli are not equal targets in that sense. Would need more refactoring in the meson script to do that, and if the `cargo check`s are removed we need it in this way anyway

Kneemund:
> Could the reason be that changing the jobs isn't reflected in the cache-key so it always reuses the previous one? Should the hash from the ci.yml file be added to the cache-key? (Or should I simply delete the caches once this is merged, because then it will be resolved anyway)

Yep, that's most likely the reason. Including the workflow file in the hash sounds like the best solution to me. Something like this should work:
```yml
key: cargo-${{ runner.os }}-${{ steps.toolchain.outputs.rustc_hash }}-${{ hashFiles('**/Cargo.lock', '.github/workflows/ci.yml') }}
``` 

Or, without hardcoding the workflow file path:
```yml
key: cargo-${{ runner.os }}-${{ steps.toolchain.outputs.rustc_hash }}-${{ hashFiles('**/Cargo.lock') }}-${{ github.workflow_sha }}
```
There doesn't seem to be a variable that (only) contains the path of the workflow, or any other way of getting its hash in general.

The `restore-key` should stay as-is though IMO, because the old/invalidated cache could still contain useful stuff that speeds up the compilation.

---

Unrelated, but I just stumbled across [this](https://github.com/flxzt/rnote/actions/runs/4108305107/jobs/7088846412) while scrolling through the actions tab:

grafik

The cache download got stuck for *some reason*, and was automatically aborted after an hour. We should drastically reduce that timeout: https://github.com/actions/cache#environment-variables. I can't think of any reasons not to; even 10 minutes should be much more than enough.

flxzt:
Okay, thanks! I just deleted the caches to try it once more but then you commented. Hopefully this works now

flxzt:
There is also some weirdness with the logging output, e.g. cargo's compilation coming before the prints in the python script

flxzt:
Okay this seems fine now.

I also noticed that `actions-rs` is unmaintained and many project are switching to https://github.com/dtolnay/rust-toolchain , we should probably do that as well. But in another PR

PR #533, opened 2023-02-27
Drawing pad shortcuts

### LINKED ISSUE #182: Add an ability to remap drawing tablet button shortcuts in the app

We can only remap pen buttons and a mouse button at the moment, as shown in the screenshot below:

Screenshot from 2022-05-18 23-41-15

It would be helpful if we could remap the features to more buttons. 

Considering that GNOME can't remap the tablet button at all due to **[this issue](https://gitlab.gnome.org/GNOME/gnome-control-center/-/issues/1822)**, and even if it could, there's **[no per app setting](https://gitlab.gnome.org/GNOME/gnome-control-center/-/issues/1861)** available in the GNOME settings, therefore in-app remapping would be very helpful.

Add a shortcut remapping of all the tools as another tab in the settings, as shown in the additional context below.

-

**Additional context**

mockup_#image47

flxzt:
This implements shortcut keys and configurable actions for buttons 1-4 on external drawing tablets. Because I don't have an external drawing tablet I couldn't verify if this actually works, so testing would be very appreciated.

TODO before merging: remove the hardcoded reset the default actions on app startup

Closes #182 when merged


flxzt:
Alright, I think I'll merge this because I believe it will work

PR #538, opened 2023-03-05
Add a new drawing

snakedye:
Made with Rnote <3


flxzt:
Thanks, that looks great!

Sorry for my ignorance, but I don't recognize the character. Who is it ( if it's a reference)? :)

Issue #539, opened 2023-03-05
Zoom tool


flxzt:
Currently zooming is only possible through either:

- touch pinch gestures
- mouse scrolling
- holding `alt + shift` and then drag with a pointer
- clicking the plus / minus buttons in the canvas menu

Those options require either a keyboard or touch input, or are nested inside a menu. A nice addition would be to add a dedicated "zoom tool" in the tools sidebar page, which works like: down y-offsets while dragging zoom out, up y-offsets zoom in.

The main use case for this would be: being able to navigate around on the canvas easily with a tablet, while "touch-drawing" is enabled. ( No keyboard, and touch input is propagated to the pens). Adding this would work nicely together with the "two finger long press" shortcut activator

**Describe alternatives you've considered**

Using the plus / minus buttons in the canvas menu, but having to always go into the menu for it has too much friction

OmegaRogue:
Having a "zoom lock" button would also be nice, to prevent accidental zooming with touch without using the dedicated zoom tool

LeSnake04:
Agree. Especially when using a real life ruler to solve a math problem or drawing stuff. (I had to do it a few times. It really sucked bc you easily zoom on accident or by reflexes)

OmegaRogue:
> Agree. Especially when using a real life ruler to solve a math problem or drawing stuff. (I had to do it a few times. It really sucked bc you easily zoom on accident or by reflexes)

For that case, a virtual ruler would also be a nice thing, tho that's a different suggestions

LeSnake04:
I am trying to implement the block zoom part as an exercise to learn the code. It looks like I got it working already, just need to add the config and button.

LeSnake04:
About the zoom button parts: cant we just reuse the functione from the tittle bar control panel.

In general: in which file are the panel buttons located?

flxzt:
Oh I originally meant implementing it in a different way. Like the way the zooming is implemented when holding `Alt + Shift` and then dragging the pointer towards up / down

LeSnake04:
Are you thinking of a "zoom to selected area" 
 
![Peek 2023-03-24 18-41](https://user-images.githubusercontent.com/47243761/227600699-994d565d-86b1-48e1-939e-1a2c05c81d7b.gif)


Or of a Click to zoom to that position:

![Peek 2023-03-24 18-39](https://user-images.githubusercontent.com/47243761/227600414-fd419cf3-e45f-4a2d-b664-4fdaf3a0c325.gif)

[...]

Issue #490, opened 2023-02-06
Zoom to real world size


PhilDevProg:
**Is your feature request related to a problem? Please describe.**
When using a lineal or set square you have to first zoom the page to the point when it has the same size as in the real world. This is not very easy and you can destroy it by accidentally zooming when you're dragging the page.

**Describe the solution you'd like**
A button that automatically zoomes the page to the same size as it is in real world (eg A4) and then disables the zoom functionality.

LeSnake04:
If you open the menu with the zoom controls and export options, you can click on the percetage to return to 100%

PhilDevProg:
> If you open the menu with the zoom controls and export options, you can click on the percentage to set it to 100%.

That doesn't resize it to the real world size. For example: The A4 page isn't as big as a A4 size in the real world.

PhilDevProg:
> If you open the menu with the zoom controls and export options, you can click on the percentage to set it to 100%.

That doesn't resize it to the real world size. For example: The A4 page isn't as big as a A4 size in the real world.

LeSnake04:
This sounds like it's quite a challenge to implement. We need to detect the PPI of the display (But what happens when 2 displays are connected?).

A less user-friendly but way easier way would be to ask the user for the value.

Kneemund:
> But what happens when 2 displays are connected?

That's not really an issue. You can get the monitor that the app is on with `appwindow.display().monitor_at_surface(&appwindow.native().unwrap().surface())`. I've already worked a bit on this, but the difficult part is reliably calculating the zoom factor. 

The closest I've gotten was `f64::from(monitor.geometry().width()) / target_width_px`, where target_width_px is the physical monitor width converted to px using the format's DPI. That only works if the canvas is as big as the screen though (e.g. empty canvas) - I'm not sure what the zoom value actually represents. And it obviously doesn't take `monitor.scale_factor()` into account yet, but that might be easy to add.

KingJames0000:
One solution that Xournal++ uses is to, as suggested by @LeSnake04, ask the user for the scale. Xournal++ uses a resizable ruler in settings so that the user can scale the ruler until it matches a physical ruler. Thereafter, all documents can zoom to 100% in the real world. It's not obtrusive, and, I suppose, probably wouldn't need to be changed by the user unless they are using fractional scaling, and even then, only has to be set once.

PR #546, opened 2023-03-08
Add option to block zoom

### LINKED ISSUE #543: Button to disable/enable pinch to zoom
When I zoomed the page to a specific size (for example to have the right dimensions for geometry) I don't want to accidentally pinch to zoom.
Edit: I'm referring to the touch screen gesture.

A button which disables/enables pinch to zoom. It could be placed in the same menu where the button for pen sounds is.

LeSnake04:
This is the result of my work so far.  It blocks scrolling unless done by the top bar (for now by default). The option is shown, but It doesn't toggle the setting yet.  

The toggle button is shown in the toolspage at the bottom (The Icon is a placeholder).

#539 
closes #543


LeSnake04:
The error I got stuck at:
````
error[E0521]: borrowed data escapes outside of method
   --> rnote-ui/src/canvaswrapper.rs:245:17
    |
240 |           fn setup_input(&self) {
    |                          -----
    |                          |
    |                          `self` is a reference that is only valid in the method body
    |                          let's call the lifetime of this reference `'1`
...
245 | /                 self.canvas_zoom_scroll_controller.connect_scroll(clone!(@weak inst as canvaswrapper => @default-return Inhibit(false), move |controller, _, dy| {
246 | |                     if controller.current_event_state() == gdk::ModifierType::CONTROL_MASK {
247 | |                         if restrict_zoom() {
248 | |                             let new_zoom = canvaswrapper.canvas().engine().borrow().camera.total_zoom() * (1.0 - dy * RnCanvas::ZOOM_STEP);
...   |
259 | |                     }
260 | |                 }));
    | |                   ^
    | |                   |
    | |___________________`self` escapes the method body here
    |                     argument requires that `'1` must outlive `'static`
```

I coulnt find a way to tranfer the bool into the closure. Maybe someone knows a trick that could help here. 

It needs to grab the current state.

LeSnake04:
My current idea would be kinda hacky: 
Set and read an environment variable that stores the state of the option. 

Is there a better way to solve this?

Kneemund:
Are you trying to disable all zooming functionality, or would disabling the zoom gesture suffice? The issue seems to refer to disabling the gesture, and there's already `RnCanvasWrapper#canvas_zoom_gesture_enable()`, which is used to disable the zoom gesture when the touch drawing mode is enabled. (Although I'm not sure if it also disables panning ATM.)

LeSnake04:
> Are you trying to disable all zooming functionality, or would disabling the zoom gesture suffice? #543 seems to refer to disabling the gesture, and there's already [`RnCanvasWrapper#canvas_zoom_gesture_enable()`](https://github.com/flxzt/rnote/blob/main/rnote-ui/src/canvaswrapper.rs#L610-L621), which is used to disable the zoom gesture when the touch drawing mode is enabled. (Although I'm not sure if it also disables panning ATM, but I don't _think_ so.)

Its intended to be part of #539 The planned zoom tool will work when zoom is restricted making it much easier to zoom with this enabled.

And I think `ctrl+scroll` is something  you  could do out of reflexes, especially when you are used to moving/zooming with mouse and keyboard .

LeSnake04:
About `canvas_gesture_zoom()`. This might  work. I will def try it.

LeSnake04:
Update: `canvas_gesture_zom_enable` worked for pinch to zoom. 

For the other zoom methods, I found a way to get the property. Though for some reason it fails to set the property value. I added some dbg!(), and it looks the value given to set_property is correct regarding setting the property, but I still always get false when reading the property. It makes no sense to me...

LeSnake04:
@flxzt can you please take a look?

rnote-ui/src/canvaswrapper.rs
```diff
@@ -534,6 +557,16 @@ impl RnCanvasWrapper {
     pub(crate) fn set_show_scrollbars(&self, show_scrollbars: bool) {
         self.set_property("show-scrollbars", show_scrollbars.to_value());
     }
+    #[allow(unused)]
+    pub(crate) fn restrict_zoom(&self) -> bool {
+        dbg!(self.property::<bool>("restrict-zoom"))
```
LeSnake04, line 562: This always returns false, no matter what the state of the toggle button is.

flxzt:
Thanks for the PR! @Kneemund maybe you want to? Pretty busy this week, it might take a few days otherwise

rnote-ui/src/canvaswrapper.rs
```diff
@@ -179,20 +181,30 @@ mod imp {
...
+                    ),
+                ]
             });
             PROPERTIES.as_ref()
         }
 
         fn property(&self, _id: usize, pspec: &glib::ParamSpec) -> glib::Value {
             match pspec.name() {
                 "show-scrollbars" => self.show_scrollbars.get().to_value(),
+                "restrict-zoom" => self.show_scrollbars.get().to_value(),
```
Kneemund, line 207: The issue lies here - you forgot to change `self.show_scrollbars` to `self.restrict_zoom`. Because of that, `self.property::<bool>("restrict-zoom")` will always return the current value of the scrollbar visibility option instead.

LeSnake04: Yes, that was it.

flxzt:
just one thing before implement too much: I think it should be an action entry in the canvas menu, not a button in the settings.

Kneemund:
My two cents on the UI:

While working on #490, I've also experimented with the UI and this was what I came up with:
(I haven't spent any time adding functionality to the toggle; especially the interaction between touch drawing and this option will require a bit more effort. I'll only implement the "zoom to real size" button then.)

grafik

An action entry like flxzt suggested would be smaller and more consistent with the other options though.

LeSnake04:
@Kneemund i was thinking about putting it there as well. (Alongside the current location)

PhilDevProg:
> My two cents on the UI:
> 
> While working on #490, I've also experimented with the UI and this was what I came up with:
> (I haven't spent any time adding functionality to the toggle; especially the interaction between touch drawing and this option will require a bit more effort. I'll only implement the "zoom to real size" button then.)
> 
> grafik
> 
> An action entry would be smaller and more consistent with the other options though.

I think that's the best looking and understandable solution. Maybe the other options (Pen Sounds/Document Borders/Touch Input) could work and look the same to be more consistent with the settings in the settings menu but that's something which should be discussed somewhere else. I think either do all options in this style with the toggle or do all in the current style with the checkmark.

LeSnake04:
I am already working on it. Should be done soon

LeSnake04:
In fact, I already created a working toggle in the style of the other buttons of the menu. 

The question is, which zooms should be blocked?
Pint to zoom should definitely be blocked, but I am not sure about the other zooms. I disabled all but the controls in the canvasmenu for now. I personally think Ctrl+scroll should be blocked as well, and maybe alt+shift+drag,but I would like your opinion.

LeSnake04:
It should now compile and run. Its a checkbox toggle in the canvasmenu.

As far as I understand we now just have to decide 2 things:

- [ ] name of the button
- [ ] which ways of zooming should be blocked

Kneemund:
I think that touch zooming is the only way to accidentally zoom, so - assuming this option is only supposed to prevent accidental zooming - the gesture should be the only thing that gets disabled (and the option should be named accordingly). I can't think of a reason why you'd want to disable other zoom methods.

LeSnake04:
Done.
Looks ready to merge from my side

rnote-ui/src/canvaswrapper.rs
```diff
@@ -534,6 +551,16 @@ impl RnCanvasWrapper {
...
     }
+    #[allow(unused)]
+    pub(crate) fn block_pinch_zoom(&self) -> bool {
+        self.property::<bool>("block-pinch-zoom")
+    }
+
+    #[allow(unused)]
+    pub(crate) fn set_block_pinch_zoom(&self, block_pinch_zoom: bool) {
+        self.set_property("block-pinch-zoom", block_pinch_zoom);
+        self.canvas_zoom_gesture_enable(!block_pinch_zoom);
```
Kneemund, line 562: This causes conflicts with touch drawing, because that toggles the zoom gesture too. One example would be: touch drawing on & block zoom on, disable touch drawing -> gesture is enabled even though block zoom is on.

LeSnake04: Should be solved now

LeSnake04:
I took the liberty to replace the canvas_zoom_gesture_enable function with an update one that sets the value based on block_pinch_zoom and touch_drawing.  I replaced it bc all calls were the same after my fix, so I made a helper function. There was no other use of the enable one aside the helper function, so I just replaced it.

I you want to keep the enable function for future usage, I can add it back in.

Kneemund (commented):
In terms of functionality, this LGTM. I only noticed some minor stylistic things I'd personally do differently (e.g. changing the condition of `canvas_zoom_gesture_update` to `!block_pinch_zoom && !touch_drawing`), but I'd rather leave that for @flxzt to review.

LeSnake04:
I assume you mean || not &&

Kneemund:
No, `!(self.block_pinch_zoom() || self.canvas().touch_drawing())` is identical to `!self.block_pinch_zoom() && !self.canvas().touch_drawing()` ([De Morgan](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)).

LeSnake04:
Ty for the  rabbit hole 

> No, `!(self.block_pinch_zoom() || self.canvas().touch_drawing())` is identical to `!self.block_pinch_zoom() && !self.canvas().touch_drawing()` ([De Morgan](https://en.wikipedia.org/wiki/De_Morgan%27s_laws)).

Thanks for the rabbit hole. That will be an interesting dive for later.

rnote-ui/src/appwindow/mod.rs
```diff
@@ -610,6 +610,11 @@ impl RnAppWindow {
             "format-borders",
             &format.show_borders.to_variant(),
         );
+        adw::prelude::ActionGroupExt::change_action_state(
+            self,
+            "block-pinch-zoom",
+            &active_tab.block_pinch_zoom().to_variant(),
+        );
```
flxzt, line 617: we could do it like this and it works, because the property `block-pinch-zoom` is updated here when the active tab changes. But I would actually prefer it if this would be a property binding, similar to the `touch-drawing` property.

Meaning:  
- also add a "global" `block-pinch-zoom` property on the appwindow
- create a binding to the canvaswrapper property in `canvaswrapper.init_reconnect()`
- refactor the action so that is a property-action of the appwindow `block-pinch-zoom` property.

flxzt:
I added some comments.

When those are addressed, there is only one thing missing: a boolean KV-pair in the settings, so that it is remembered across sessions. Check out `touch-drawing` how to create a settings bind for it.

rnote-ui/src/appwindow/appwindowactions.rs
```diff
@@ -302,18 +302,6 @@ impl RnAppWindow {
...
-            clone!(@weak self as appwindow => move |action_block_pinch_zoom, state_request| {
-                let block_pinch_zoom = state_request.unwrap().get::<bool>().unwrap();
-                let active_tab = appwindow.active_tab();
-
-                active_tab.set_block_pinch_zoom(block_pinch_zoom);
-
-                action_block_pinch_zoom.set_state(block_pinch_zoom.to_variant());
-            }),
-        );
-
```
LeSnake04: I removed it since I am not sure what it does and if its nessary for PropertyActions. I found no similar code for touch-drawing

LeSnake04:
I tried to make this a propertyaction by making it more like touch-drawing

But it panics with this message:
'Source property block-pinch-zoom on type RnAppWindow not found

Kneemund:
You haven't created the property in `RnAppWindow` yet, so you can't bind anything to it. Take a look at this:

https://github.com/flxzt/rnote/blob/228ab134d26c8f466ea9a21562a4c2991a2e9f2c/rnote-ui/src/appwindow/imp.rs#L151-L153

and all the other related properties and functions in `appwindow/imp.rs` and `appwindow/mod.rs`. You haven't done that for `block-pinch-zoom` yet.

rnote-ui/src/appwindow/appwindowactions.rs
LeSnake04: It seems to work without it.

LeSnake04:
Thanks, that fixed it. 
@flxzt I added the changes you requested.

flxzt:
I forgot to mention that the property bind also needs to be unbound in canvaswrapper's `disconnect_handlers()`.

And I would move the action in the menu to below "Draw with Touch Input", it looks a bit odd where it is right now.

Other than that, this looks good to me. @Kneemund do you have anything else?

rnote-ui/src/appwindow/appwindowactions.rs
flxzt: If it's not overwritten by a custom callback, the default changes the action state (so: the property) in the same way

LeSnake04:
Do you mean this?
https://github.com/flxzt/rnote/blob/ce512ffc0f5c24d90aed1e1622d5219c702c30d7/rnote-ui/src/canvaswrapper.rs#L570-L576

flxzt:
no, [here](https://github.com/flxzt/rnote/blob/228ab134d26c8f466ea9a21562a4c2991a2e9f2c/rnote-ui/src/canvaswrapper.rs#L548)

Kneemund:
> Other than that, this looks good to me. @Kneemund do you have anything else?

Looks good to me too.

LeSnake04:
> I forgot to mention that the property bind also needs to be unbound in canvaswrapper's `disconnect_handlers()`.
> 
> And I would move the action in the menu to below "Draw with Touch Input", it looks a bit odd where it is right now.
> 
> Other than that, this looks good to me. @Kneemund do you have anything else?

Done

PR #551, opened 2023-03-13
Adding rust toolchain

TornaxO7:
I got the following error message:
```
error: rustup could not choose a version of cargo to run, because one wasn't specified explicitly, and no default is configured.
help: run 'rustup default stable' to download the latest stable release of Rust and set it as your default toolchain.
```
This PR should fix it and makes sure that all devs are using the same rust tools


TornaxO7:
Wait, I see, that you've removed it [here](https://github.com/flxzt/rnote/commit/1fd35a09d2a9b561aaa6b7041cb590057afdd794). Did it really destroy the windows build?

flxzt:
Cargo then always picked `msvc` instead of the windows `gnu` toolchain, even when it is set as default with `rustup default stable-gnu` . If that can be fixed I generally don't have anything against adding this file.

But also: why don't you just set `stable` as default with `rustup default stable`?

flxzt:
so this warning is fixed by `rustup default nightly` then too, right?

TornaxO7:
> so this ~warning~ error is fixed by `rustup default nightly` then too, right?

no, I'm getting the error which I posted, if I use `rustup default nightly`.

flxzt:
oh ok. If you know how to fix the issue mentioned above we could add it.

What you could also do: `rust override set stable` from the rnote directory which sets stable as the default only in there.

Issue #55, opened 2022-01-21
Add ability to edit brush and shape strokes


dalcde:
It would be convenient to be able to change all strokes in a selection to a new color.

defini-tiv:
This feature could be thought more general:

- color
- stroke width
- stroke style

And all other properties of this kind would be useful to be able to change retrospectively.

jaklt:
I second @dalcde's suggestion, I use it in Xournal++ all the time. I often write something and after some time I realise that it's probably not as significant so I make it smaller, move it to the side and change its colour to something not as prominent. This is also useful the other way around, after writing something I often decide to make it stand out, so I just change the colour of it.

e87131:
MS apps seem to allow for this (OneNote, Whiteboard)
great for exporting in different colors if youre drawing on black bg and exporting for print or something

LeSnake04:
Assuming the color is just a variable, changing it should be as easy as:

1. Get current color
2. Loop over all selected objects, setting the color of the object. 

I dont know how easy it is to get write access to selected objects though.2

PR #563, opened 2023-03-15
edit fill/stroke/text color of existing strokes 

Kneemund:
* Partially implements #55. Editing other aspects of existing strokes is not possible without UI/workflow changes, because the selection disappears as soon as you select another tool (and color is currently the only global property).
* Might need some more abstraction/cleaner code, it's not very consistent with e.g. the transformation stuff. I'm not sure if style modifications should be split into a trait/function for each property, or stay flexible as-is (like `Typewriter#change_text_style_in_modifying_stroke()`).
* Because you can deselect colors, you have to click on a color twice in order to apply it if it's already selected.

I *hope* that the rendering/geometry updates are properly taken care of, I'm not very familiar with this part of Rnote. I haven't noticed any issues though.

@flxzt it would be great if you could take a look :P

[Bildschirmaufzeichnung vom 2023-03-15, 23-14-36.webm](https://user-images.githubusercontent.com/49598842/225458255-6a53fbb6-4805-4a74-b02f-fe2b38452421.webm)

### DIFF
--- a/rnote-ui/src/overlays.rs
+++ b/rnote-ui/src/overlays.rs
@@ -262,7 +262,10 @@ impl RnOverlays {
                         }
                         PenStyle::Selector => {
                             let selection_keys = engine.store.selection_keys_unordered();
-                            let widget_flags = engine.store.change_stroke_colors(&selection_keys, stroke_color, engine.tasks_tx.clone(), &engine.camera);
+                            let widget_flags = engine.store.change_stroke_colors(&selection_keys, stroke_color);
+                            if let Err(e) = engine.update_rendering_current_viewport() {
+                                log::error!("failed to update rendering for current viewport while changing stroke color of selection, Err: {e:?}");
+                            }
                             appwindow.handle_widget_flags(widget_flags, &canvas);
                         }
                         PenStyle::Brush | PenStyle::Shaper | PenStyle::Eraser | PenStyle::Tools => {}
@@ -288,7 +291,10 @@ impl RnOverlays {
                 match stroke_style {
                     PenStyle::Selector => {
                         let selection_keys = engine.store.selection_keys_unordered();
-                        let widget_flags = engine.store.change_fill_colors(&selection_keys, fill_color, engine.tasks_tx.clone(), &engine.camera);
+                        let widget_flags = engine.store.change_fill_colors(&selection_keys, fill_color);
+                        if let Err(e) = engine.update_rendering_current_viewport() {
+                            log::error!("failed to update rendering for current viewport while changing fill color of selection, Err: {e:?}");
+                        }
                         appwindow.handle_widget_flags(widget_flags, &canvas);
                     }
                     PenStyle::Typewriter | PenStyle::Brush | PenStyle::Shaper | PenStyle::Eraser | PenStyle::Tools => {}
--- a/rnote-ui/src/overlays.rs
+++ b/rnote-ui/src/overlays.rs
@@ -2,10 +2,12 @@ use gtk4::{
     gio, glib, glib::clone, prelude::*, subclass::prelude::*, CompositeTemplate, Overlay,
     ProgressBar, ScrolledWindow, ToggleButton, Widget,
 };
+use rnote_compose::Style;
 use rnote_engine::engine::EngineViewMut;
 use rnote_engine::pens::{Pen, PenStyle};
 use rnote_engine::utils::GdkRGBAHelpers;
 use std::cell::RefCell;
+use std::time::Instant;
 
 use crate::canvaswrapper::RnCanvasWrapper;
 use crate::RnPensSideBar;
@@ -239,7 +241,7 @@ impl RnOverlays {
                     engine.pens_config.brush_config.solid_options.stroke_color = Some(stroke_color);
                     engine.pens_config.brush_config.textured_options.stroke_color = Some(stroke_color);
                     engine.pens_config.shaper_config.smooth_options.stroke_color = Some(stroke_color);
-                    engine.pens_config.shaper_config.rough_options.stroke_color= Some(stroke_color);
+                    engine.pens_config.shaper_config.rough_options.stroke_color = Some(stroke_color);
                     engine.pens_config.typewriter_config.text_style.color = stroke_color;
 
                     match stroke_style {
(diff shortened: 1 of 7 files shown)


flxzt:
Thanks!

> Editing other aspects of existing strokes is not possible without UI/workflow changes

Editing the color is by far the most important property to edit, I think its fine for now. Rnote doesn't want to be a vector editing progam and as you said it would require new UI with a dedicated, context-dependent "edit" panel or something like that. In the end it should be encouraged to simply erase & redraw or undo a stroke/ shape when it has the wrong width / fill pattern / .. .

> Might need some more abstraction/cleaner code, it's not very consistent with e.g. the transformation stuff.

I am not sure about that myself. I don't completely like the signature with those two closures. Maybe reducing the abstraction would be better, like: `recolor_strokes(key: &[StrokKey], color: Color)` ? Because else we end up moving the logic out of the engine into the UI.
I know this is not entirely consistent to how the text-strokes text-styles are modified right now, but I would write that part differently today.

> I hope that the rendering/geometry updates are properly taken care of

I'll add some comments in the code

> Because you can deselect colors, you have to click on a color twice in order to apply it if it's already selected

How about deselecting the picked palette color when strokes are selected? Since the selected palette color is then considered "outdated". Similar to how it deselects when the user switches between fill & stroke color. This would require a new field in `WidgetFlags`, `deselect_palette_color` or similar

I am also not very satisfied how all of the engine -> UI communication is handled currently, I might need to refactor the `WidgetFlags` into messages or something like that. We need better coupling between engine and UI, suggestions & ideas are very welcome.

One more thing: How about adding a intermediary multi-color state for the active fill / stroke color pads when after selecting multiple strokes with different colors, like a rainbow gradient or something like that?

rnote-ui/src/overlays.rs
```diff
@@ -260,7 +262,34 @@ impl RnOverlays {
                                 appwindow.handle_widget_flags(widget_flags, &canvas);
                             }
                         }
-                        PenStyle::Brush | PenStyle::Shaper | PenStyle::Eraser | PenStyle::Selector | PenStyle::Tools => {}
+                        PenStyle::Selector => {
+                            let mut widget_flags = engine.record(Instant::now());
```
flxzt, line 266: I'd move all of this into the function in the engine, and let it return `WidgetFlags` with `redraw` and `store_modified` flagged true. Then the code in the UI is reduced to two lines: the function call and `appwindow.handle_widget_flags()`

rnote-ui/src/overlays.rs
```diff
@@ -260,7 +262,34 @@ impl RnOverlays {
...
+                                match style {
+                                    Style::Rough(rough_style) => rough_style.stroke_color = Some(stroke_color),
+                                    Style::Smooth(smooth_style) => smooth_style.stroke_color = Some(stroke_color),
+                                    Style::Textured(textured_style) => textured_style.stroke_color = Some(stroke_color),
+                                }
+                            }, |text_style| {
+                                text_style.color = stroke_color;
+                            });
+
+                            engine.store.update_geometry_for_strokes(&selection_keys);
```
flxzt, line 279: I don't think this is needed for color modifications (not that it does much harm)

rnote-ui/src/overlays.rs
```diff
@@ -260,7 +262,34 @@ impl RnOverlays {
...
+                            engine.store.regenerate_rendering_in_viewport_threaded(
```
flxzt, line 280: I am a little surprised the rendering is updated as shown in the clip, because here is missing:

```rust
store.set_rendering_dirty(&selection_keys);
```

To flag that the rendering is outdated

Kneemund:
Alright. 

> One more thing: How about adding a interim multi-color state for the active fill / stroke color pads after selecting multiple strokes with different colors, like a rainbow gradient or something like that?

I'm not exactly sure what you mean. Should the active fill/stroke colors be set to the color of the selected strokes (they currently aren't)?

rnote-ui/src/overlays.rs
flxzt: I meant as a method on the store. 

> then it would need an engine view (specifically the task sender and camera) as a parameter
thats true. Is it a huge issue to have that as "public" API from the perspective of rnote-ui ? I think typewriter does that as well. Otherwise there are some functions where I simply added a doc comment with "strokes then need to update their rendering", which also has the advantage avoiding iterating over the currently viewed strokes multiple times when those modifications are chained (not in this case, but for example, selection-duplication).
I'd then follow (as you said) it by a call to `canvas.update_rendering()`

flxzt:
Thats at least what the typewriter does. It updates the active stroke / fill colors to the active text

rnote-ui/src/overlays.rs
flxzt: I eventually want to refactor this part by some sort of automatic diffing between the current and previous state of the engine after making modifications. We already have the immutable,  cheap-to-clone & clone-on-write (the `Arc`'s) state to do that

flxzt:
I guess the issue is that there is not really a `None` / neutral color that could be set here. But probably its not a big deal when the active color stays the same

rnote-ui/src/overlays.rs
Kneemund: > Is it an issue to have that as "public" API from the perspective of rnote-ui ? I think typewriter does that as well.

Probably not, and yeah, typewriter does it too. Actually, I'd only pass the task sender and camera as parameters because `RnoteEngine#store` is already being borrowed while calling the `change_fill/stroke_color` function. Or we go for the alternative you described, but that would (at least ATM) needlessly re-render the background.

flxzt: It would only reapply the prerendered background tile onto the gsk scene graph :) there is a comment that says creating the texture is "expensive" but that is only in the context of that function.

I think it would be best to keep the rerendering out of these functions. So that the flow is `strokes are modified -> rendering flagged dirty -> finish up with rerendering those that are dirty`. This is consistent with other functions, and there is also the benefit that if we want to reuse it somewhere, we don't end up having to thread `EngineView` or `Camera / tasks_tx` through them.

Kneemund: Alright. I've used `engine.update_rendering_current_viewport()` though, because engine is already borrowed and the canvas function does more unneeded stuff :P

flxzt: I just added [update_content_rendering_current_viewport(&mut self))[https://github.com/flxzt/rnote/blob/9637527bd86d28441a369e926d001b50c0ef5643/rnote-engine/src/engine/rendering.rs#L46] for this

Kneemund:
> the color is by far the most important property to edit, I think its fine for now. Rnote doesn't want to be a vector editing progam and as you said it would require new UI with a dedicated, context-dependent "edit" panel or something like that. 

Does this close #55 then? Or should it stay open?

flxzt:
I think it can be closed

flxzt:
ok, looks good to me :) only two clippy lints to fix

Kneemund:
Right, I often forget to look for those :P

flxzt:
thanks :)

PR #569, opened 2023-03-17
handle macOS in meson_post_install.py

dotlambda:


flxzt:
thanks!

@dehesselle how does this look to you wrt building the app bundles?

dehesselle:
This is an improvement as I've been manually compiling the schemas during my build, and I won't have to with this change. I'll give it a test run later today for a final feedback.

dehesselle:
Here's the result from my test run on macOS:
- Executable to generate icon cache is versioned on macOS, i.e. `gtk4-update-icon-cache`. 
- I don't have a `update-desktop-database` executable.
- I don't have a `update-mime-database` executable.

It take it I am the first to test this? Or can you tell me more about your macOS build @dotlambda?

I'm not opposed to add things to my build that are missing (which is quite possible, building Rnote is a recent project of mine, no long-term experience yet), but I'd like to have a reason what functionality is gained.

dotlambda:
> * Executable to generate icon cache is versioned on macOS, i.e. `gtk4-update-icon-cache`.

The same on Linux. Since Rnote uses GTK4 I think we should move to `gtk4-update-icon-cache`.

> * I don't have a `update-desktop-database` executable.
> * I don't have a `update-mime-database` executable.

They are provided by https://www.freedesktop.org/wiki/Software/desktop-file-utils/ and https://www.freedesktop.org/wiki/Software/shared-mime-info/ which can e.g. be installed using Homebrew.

> I take it I am the first to test this? Or can you tell me more about your macOS build @dotlambda?

No, this patch is already being used in Nixpkgs: https://github.com/NixOS/nixpkgs/commit/cce8dfae835a3b906301782c362a021830ccffe0
Note that the packages mentioned above are provided during the build, so the executables are found.

> I'm not opposed to add things to my build that are missing (which is quite possible, building Rnote is a recent project of mine, no long-term experience yet), but I'd like to have a reason what functionality is gained.

I'm honestly not sure whether the desktop database (`~/.local/share/applications`) and the MIME-type database (`~/.config/mimeapps.list`) are of any use on macOS. But they definitely don't hurt.

flxzt:
FYI as soon as meson is updated to >= v0.64 on all supported platforms (only the flatpak sdk is [missing](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/11421), I'll remove this post_install script and uncomment [this](https://github.com/flxzt/rnote/blob/244ab28ee43325b8cda72e66390328c41cbd2b19/meson.build#L311). I looked into the meson [sources](https://github.com/mesonbuild/meson/blob/46fe127503ec0a07f56904fb44b5ae9aac4374a2/mesonbuild/modules/gnome.py#LL362C8-L362C8), it uses `gtk4-update-icon-cache`

dotlambda:
> FYI as soon as meson is updated to >= v0.64 on all major platforms (only the flatpak sdk is [missing](https://gitlab.com/freedesktop-sdk/freedesktop-sdk/-/merge_requests/11421) ), I'll remove this post_install script and uncomment [this](https://github.com/flxzt/rnote/blob/244ab28ee43325b8cda72e66390328c41cbd2b19/meson.build#L311). I looked into the meson [sources](https://github.com/mesonbuild/meson/blob/46fe127503ec0a07f56904fb44b5ae9aac4374a2/mesonbuild/modules/gnome.py#LL362C8-L362C8), it uses `gtk4-update-icon-cache`

Nice, that would do exactly what this PR is trying to achieve.

flxzt:
> The same can be said for Windows, where update-desktop-database.exe and update-mime-database.exe are run by the post install script.

I think it is relevant when building / using applications installed into the msys environment on windows

Okay, so no objections to merging this , right?

dehesselle:
> > * Executable to generate icon cache is versioned on macOS, i.e. `gtk4-update-icon-cache`.
> 
> The same on Linux. Since Rnote uses GTK4 I think we should move to `gtk4-update-icon-cache`.

At least this should be changed.

I will work around the other things on my end.

flxzt:
ok, I'll quickly push into here

dehesselle:
I have updated my build system and gave it a test run so that I'm prepared for this to be in the next release.

Issue #463, opened 2023-01-23
A mechanism to remove/delete a page (with or without content)


mizhozan:
I couldn't find how to remove a (blank page). The only relevant issue I found is #24 which addresses another problem.

PhilDevProg:
@flxzt You can assign me, I'm working on it and should have a pull request ready soon.

PR #572, opened 2023-03-17
Add feature: Remove page of fixed size document

PhilDevProg:
Closes https://github.com/flxzt/rnote/issues/463
This pull request adds a button next to the add page button in the headerbar which removes a page. At the moment there are two issues:

1. This only reduces the document height by one page but doesn't remove the content.  @flxzt Is there a way to remove the content of a specific page or from a specific height to another height?
2. On some document types (e.g. A3 landscape) the page can be removed even if only one exists and the height of the document can go into the negative area which makes documents appear above the two red x.


PhilDevProg:
Commit was authored with a wrong name. New PR: https://github.com/flxzt/rnote/pull/573

PR #573, opened 2023-03-17
Add feature: Remove page from fixed size document

PhilDevProg:
Closes https://github.com/flxzt/rnote/issues/463
This pull request adds a button next to the add page button in the headerbar which removes a page when using fixed size document.

- [x] This only reduces the document height by one page but doesn't remove the content.  @flxzt Is there a way to remove the content of a specific page or from a specific height to another height? Edit: See https://github.com/flxzt/rnote/commit/c9b42bf88495be7f62d9e97a0c37227adf991630 Thank you, @Kneemund !
- [x] Images and unsplittable strokes are also removed, when they are only partially on the page that gets removed. See https://github.com/flxzt/rnote/commit/af4399f542d23b4a01bd75de1d12d2e2e05cf99f
- [x] On some document types (e.g. A3 landscape) the page can be removed even if only one exists and the height of the document can go into the negative area which makes documents appear above the two red x. Edit: See https://github.com/flxzt/rnote/commit/63578dcdaf0e8dff4c084c7f33aeb946cc35bd6e

I closed https://github.com/flxzt/rnote/pull/572 because the commit was authored with the wrong name.


PhilDevProg:
Oops, forgot to push updated UI file

PhilDevProg:
Now that the second issue is fixed, I need your help @flxzt. I searched for something like that but didn't found something in the code: Is there a way to remove the content from a specific height to another height?

Kneemund:
You could create an AABB/bounding box depending on the height, and then use `RnoteEngine#store.stroke_keys_as_rendered_intersecting_bounds` in order to get all strokes within that area, which you can then trash (similar to [trashing the current selection](https://github.com/flxzt/rnote/blob/main/rnote-ui/src/appwindow/appwindowactions.rs#L384-L397)). There might be a better solution to this though.

PhilDevProg:
@Kneemund I tried this but I can't find how to create a box with a specific height and with at a specific height...

Kneemund:
I'm assuming it'd be something like `Aabb::new(na::point![top_left_x, top_left_y], na::point![top_left_x + width, top_left_y + height])` (i.e. top left, bottom right).

PhilDevProg:
I now have this:

```
 let format_height = canvas.engine().borrow().document.format.height;
                 let format_width = canvas.engine().borrow().document.format.width;
                 let doc_height = canvas.engine().borrow().document.height;
                 let new_doc_height = doc_height - format_height;
                 let top_left_x = '0';
                 if doc_height > format_height {
                     Aabb::new(na::point![top_left_x, new_doc_height], na::point![top_left_x + format_width, doc_height]);
                    let remove_area = canvas.engine().borrow_mut().store.stroke_keys_as_rendered_intersecting_bounds(Aabb);
                     canvas.engine().borrow_mut().store.set_trashed_keys(&remove_area, true);
```

and I get these errors:

```
error[E0423]: expected value, found struct `Aabb`
   --> rnote-ui/src/appwindow/appwindowactions.rs:549:118
    |
549 | ...cting_bounds(Aabb);
    |                 ^^^^ help: use struct literal syntax instead: `Aabb { mins: val, maxs: val }`
    |
   ::: /home/phil/Dokumente/rnote/_mesonbuild/cargo-home/registry/src/github.com-1ecc6299db9ec823/parry2d-f64-0.13.3/src/bounding_volume/aabb.rs:25:1
    |
25  | pub struct Aabb {
    | --------------- `Aabb` defined here

error[E0308]: mismatched types
   --> rnote-ui/src/appwindow/appwindowactions.rs:548:42
    |
548 | ...   Aabb::new(na::point![top_left_x, new_doc_height], na::point![top_left_x + format_width, doc_height]);
    |                            ^^^^^^^^^^ expected `f64`, found `char`

error[E0369]: cannot add `f64` to `char`
   --> rnote-ui/src/appwindow/appwindowactions.rs:548:93
    |
548 | ...   Aabb::new(na::point![top_left_x, new_doc_height], na::point![top_left_x + format_width, doc_height]);
    |                                                                    ---------- ^ ------------ f64
    |                                                                    |
    |                                                                    char

Some errors have detailed explanations: E0308, E0369, E0423.
For more information about an error, try `rustc --explain E0308`.
error: could not compile `rnote` due to 3 previous errors
```

Kneemund:
Well, `top_left_x` has to be a float, e.g. `let top_left_x = 0.0;`. And you'll have to store `Aabb::new(...)` in a variable and then pass that variable to the function, or use a struct literal as the error message suggests.

PhilDevProg:
Thanks, I'll try that. I'm very new to Rust😅

PhilDevProg:
@Kneemund Yes, That works perfect! Thank you

PhilDevProg:
Hmm, the current solution also removes anything that isn't completely on the page... Any idea, @Kneemund @flxzt ?

PhilDevProg:
I made this PR Draft again until a better solution is found.

PhilDevProg:
> What do you mean? It removes things that were only partly on the deleted page?

Yes

> you can take a look at split_colliding_strokes()

I'll check that out.

PhilDevProg:
I'm really overwhelmed by this function 😅 I don't understand how to make it work with my code. And even if, there would still be the problem that images/PDFs/... would be removed which is probably more annoying than some strokes...

PhilDevProg:
Strokes are now split when they are also on other pages. The only remaining problem are images/not-splittable strokes/... @flxzt Maybe you have an idea?

flxzt:
I'd say trash those that are contained in the to be removed page? And those which only intersect can be left untouched. That's what I would expect.

Btw. you don't have to constrain yourself to implement it by composing existing functions together, you could craft your function `remove_area()` that does exactly what is wanted

PhilDevProg:
> I'd say trash those that are contained in the to be removed page? And those which only intersect can be left untouched. That's what I would expect.

Yes, that'd be the best solution. There is currently no existing way in the code of Rnote to only select the strokes, that are completely in the box, right?

> Btw. you don't have to constrain yourself to implement it by composing existing functions together, you could craft your function `remove_area()` that does exactly what is wanted

I already thought about that but I'm still very overwhelmed by Rnote's code and it's also the first time working with Rust for me so I gave up last time. But I will look again at the code and try to find a solution.

PhilDevProg:
> I am not sure why you'd need to select any strokes, but there is [strokes_hitboxes_contained_in_aabb](https://github.com/flxzt/rnote/blob/228ab134d26c8f466ea9a21562a4c2991a2e9f2c/rnote-engine/src/store/stroke_comp.rs#L585) that returns the keys for all strokes whose hitboxes are completely contained in the given axis-aligned bounding box

Thank you very much, I'll try it. I select the strokes of the page that gets removed to remove them since I would expect a button which removes a page to also remove the content of it.

PhilDevProg:
@flxzt Hmm, your solution doesn't work with imported files like images or unsplittable strokes like shapes.

flxzt:
sorry, but I only pointed you towards a function that does what you asked. Since this is your PR, aren't you supposed to come up with a solution here? ;)

PhilDevProg:
> sorry, but I only pointed you towards a function that does what you asked. Since this is your PR, aren't you supposed to come up with a solution here? ;)
> 
> 

Sure 😅, I just was a little bit confused because I thought that function works for imported files/unsplittable strokes.

PhilDevProg:
I figured out a solution and now everything works so it's ready for Review.

rnote-ui/src/appwindow/appwindowactions.rs
```diff
@@ -537,6 +540,27 @@ impl RnAppWindow {
             }),
         );
 
+        // Remove page from doc in fixed size mode
+        action_remove_page_from_doc.connect_activate(
+            clone!(@weak self as appwindow => move |_action_remove_page_from_doc, _target| {
```
flxzt, line 545: lets add a record call here:

```rust
let mut widget_flags = canvas.engine().borrow_mut().record(Instant::now());

// ...

```

so that the stroke trashing/splitting can be undone. Unfortunately changing document dimensions cannot be undone currently, but that would be a separate improvement.

rnote-ui/src/appwindow/appwindowactions.rs
```diff
@@ -537,6 +540,27 @@ impl RnAppWindow {
...
+                let canvas = appwindow.active_tab().canvas();
+
+                let format_height = canvas.engine().borrow().document.format.height;
+                let format_width = canvas.engine().borrow().document.format.width;
+                let doc_height = canvas.engine().borrow().document.height;
+                let new_doc_height = doc_height - format_height;
+                if doc_height > format_height {
+                    let aabb = Aabb::new(na::point![0.0, new_doc_height], na::point![0.0 + format_width, doc_height]);
+                    canvas.engine().borrow_mut().store.split_colliding_strokes(aabb, aabb);
```
flxzt, line 554: this returns WidgetFlags which should be handled. With the suggested changes from the above comment, it could be something like:

```rust
let (_, wf) = canvas.engine().borrow_mut().store.split_colliding_strokes(aabb, aabb);
widget_flags.merge(wf);
```

rnote-ui/src/appwindow/appwindowactions.rs
```diff
@@ -537,6 +540,27 @@ impl RnAppWindow {
...
+                    let remove_area = canvas.engine().borrow_mut().store.stroke_keys_as_rendered_in_bounds(aabb);
```
flxzt, line 555: let's name this a bit differently to avoid confusion, I'd append `_keys` so: `remove_area_keys`

flxzt:
Thanks, and sorry that it took a while.

I added some comments only about some minor things.

With this change the headerbar has a lot of buttons, I wonder if there could be a better place for these buttons.

PhilDevProg:
> With this change the headerbar becomes really crowded, I wonder if there could be a better place for these buttons.

The only other place I can think of is the tools menu. There already is the 'Insert vertical space' button which falls into the same rough category in my opinion as the page+ button and the page- button because all can remove and add space to the document but the page button might be less discoverable there...

rnote-ui/src/appwindow/appwindowactions.rs
PhilDevProg: I tried the following:
```
                let mut widget_flags = canvas.engine().borrow_mut().record(Instant::now());  

                let format_height = canvas.engine().borrow().document.format.height;
                let format_width = canvas.engine().borrow().document.format.width;
                let doc_height = canvas.engine().borrow().document.height;
                let new_doc_height = doc_height - format_height;
                if doc_height > format_height {
                    let aabb = Aabb::new(na::point![0.0, new_doc_height], na::point![0.0 + format_width, doc_height]);
                    let (_, wf) = canvas.engine().borrow_mut().store.split_colliding_strokes(aabb, aabb);
                    widget_flags.merge(wf);
```

but splitting doesn't work anymore...

rnote-ui/src/appwindow/appwindowactions.rs
PhilDevProg: Done

rnote-ui/src/appwindow/appwindowactions.rs
PhilDevProg: Done

PhilDevProg:
@flxzt I found a bug:
[Bildschirmaufzeichnung vom 2023-05-02, 15-50-56.webm](https://user-images.githubusercontent.com/91820316/235687166-badda39f-3688-491d-869e-e4977909333e.webm)
It looks like strokes get splitted at the wrong place sometimes... Do you have an idea why this happens?

flxzt:
It splits with the granularity of the segments. It looks like since you drew at a small scale, the distance between segments is large so this behaviour becomes more apparent. If you check "visual debugging" it should become clear how it works

PhilDevProg:
> It splits with the granularity of the segments. It looks like since you drew at a small scale, the distance between segments is large so this behaviour becomes more apparent after zooming in. If you check "visual debugging" it should become clear how it works

Thanks for the explanation. Should the stroke splitting part of the remove page button stay? It could confuse users because parts of their strokes are gone. Or should it remain, maybe as an option?

rnote-ui/src/appwindow/appwindowactions.rs
```diff
@@ -537,6 +540,31 @@ impl RnAppWindow {
...
+
+                let format_height = canvas.engine().borrow().document.format.height;
+                let format_width = canvas.engine().borrow().document.format.width;
+                let doc_height = canvas.engine().borrow().document.height;
+                let new_doc_height = doc_height - format_height;
+                if doc_height > format_height {
+                    let aabb = Aabb::new(na::point![0.0, new_doc_height], na::point![0.0 + format_width, doc_height]);
+                    let (_, wf) = canvas.engine().borrow_mut().store.split_colliding_strokes(aabb, aabb);
+                    widget_flags.merge(wf);
+                    appwindow.handle_widget_flags(widget_flags, &canvas);
```
flxzt, line 558: we should move this to outside of the condition, as the last step of the action

flxzt:
I am not completely sure, what users would expect here

rnote-ui/src/appwindow/appwindowactions.rs
PhilDevProg: Which parts specifically? Only the part that handles stroke splitting and widget flags or all 4 lines of code? Do you mean by 'condition' the if statement?

flxzt: only:

`appwindow.handle_widget_flags(widget_flags, &canvas);`

since that is "produced" by the record call and should always be handled.

> Do you mean by 'condition' the if statement?

yes, to the bottom right before the action closure is finished

PhilDevProg: Done

PhilDevProg:
> I am not completely sure, what users would expect here. But it shouldn't be an option, regardless how we decide

I think most users would propose be okay with their strokes not splitting and only those strokes that are completely on the page getting removed.

flxzt:
Okay sounds good, that also is more consistent to the other stroke types!

PhilDevProg:
Then I'll remove the code for stroke splitting.

flxzt:
Sounds good! The only other thing I noticed is that the icon svg is "inkscape svg". I usually export / save the icons as "plain svg" so that the inkscape-specific xml is stripped

PhilDevProg:
> Sounds good! The only other thing I noticed is that the icon svg is "inkscape svg". I usually export / save the icons as "plain svg" so that the inkscape-specific xml is stripped

Done!

flxzt:
Thanks @PhilDevProg !

flxzt:
Just FYI I followed up with moving the logic of add & remove page into the engine and I noticed (after merging) that your bounds approach didn't really work for strokes that span multiple pages. I quickly refactored it to get the keys through the already existing `keys_below_y_pos()`. The other functions (`*_in_bounds()`) might still come in handy in the future, so those are definitely nice additions!

PR #576, opened 2023-03-18
README.md Minor Changes

GabsEdits:
Made the **icon** smaller, centered the **title** & **description**, and other small changes.


flxzt:
thanks for the effort, but I actually like the bigger icon and the existing formatting better for our README

PR #579, opened 2023-03-20
inertial touch scrolling on canvas 

### LINKED ISSUE #291: Velocity scrolling from touch input
When scrolling with my finger, the scrolling immediately stops when my finger leaves the screen. I would expect the scrolling to continue for a short time after the gesture, as is already implemented for scrolling with trackpad and presumably mouse input. I'm not sure if this is missing feature or a bug, since this behavior is implemented for the other inputs. 

Keep up the great work :)

Rnote Version: 0.5.5
Installation Method: Flatpak
DE: GNOME 42

Kneemund:
This PR adds an option for "inertial touch scrolling" which is enabled by default and gets unsensitive if touch drawing is enabled. Enabling touch drawing overrides & disables inertial scrolling and enables the touch drag gesture (in general, the behavior should be identical to the one described [in this comment](https://github.com/flxzt/rnote/issues/291#issuecomment-1476619529)).

* Closes #291.
* Inertial scrolling seems to be much more commonly used than kinetic scrolling.

grafik

### DIFF
--- a/rnote-ui/src/appwindow/appsettings.rs
+++ b/rnote-ui/src/appwindow/appsettings.rs
@@ -78,6 +78,14 @@ impl RnAppWindow {
             )
             .get_no_changes()
             .build();
+        self.bind_property(
+            "touch-drawing",
+            &self.settings_panel().general_inertial_scrolling_switch(),
+            "sensitive",
+        )
+        .invert_boolean()
+        .sync_create()
+        .build();
 
         // regular cursor
         self.app_settings()
--- a/rnote-ui/src/appwindow/appsettings.rs
+++ b/rnote-ui/src/appwindow/appsettings.rs
@@ -69,6 +69,16 @@ impl RnAppWindow {
             .get_no_changes()
             .build();
 
+        // inertial scrolling
+        self.app_settings()
+            .bind(
+                "inertial-scrolling",
+                &self.settings_panel().general_inertial_scrolling_switch(),
+                "active",
+            )
+            .get_no_changes()
+            .build();
+
         // regular cursor
         self.app_settings()
             .bind(
--- a/rnote-ui/src/appwindow/appsettings.rs
+++ b/rnote-ui/src/appwindow/appsettings.rs
@@ -78,14 +78,6 @@ impl RnAppWindow {
             )
             .get_no_changes()
             .build();
-        self.bind_property(
-            "touch-drawing",
-            &self.settings_panel().general_inertial_scrolling_switch(),
-            "sensitive",
-        )
-        .invert_boolean()
-        .sync_create()
-        .build();
 
         // regular cursor
         self.app_settings()
--- a/rnote-ui/src/canvaswrapper.rs
+++ b/rnote-ui/src/canvaswrapper.rs
@@ -20,10 +20,12 @@ mod imp {
     pub(crate) struct RnCanvasWrapper {
         pub(crate) show_scrollbars: Cell<bool>,
         pub(crate) block_pinch_zoom: Cell<bool>,
+        pub(crate) inertial_scrolling: Cell<bool>,
         pub(crate) pointer_pos: Cell<Option<na::Vector2<f64>>>,
 
         pub(crate) appwindow_block_pinch_zoom_bind: RefCell<Option<glib::Binding>>,
         pub(crate) appwindow_show_scrollbars_bind: RefCell<Option<glib::Binding>>,
+        pub(crate) appwindow_inertial_scrolling_bind: RefCell<Option<glib::Binding>>,
         pub(crate) appwindow_righthanded_bind: RefCell<Option<glib::Binding>>,
         pub(crate) canvas_touch_drawing_handler: RefCell<Option<glib::SignalHandlerId>>,
 
@@ -104,10 +106,12 @@ mod imp {
             Self {
                 show_scrollbars: Cell::new(false),
                 block_pinch_zoom: Cell::new(false),
+                inertial_scrolling: Cell::new(true),
                 pointer_pos: Cell::new(None),
 
                 appwindow_block_pinch_zoom_bind: RefCell::new(None),
                 appwindow_show_scrollbars_bind: RefCell::new(None),
+                appwindow_inertial_scrolling_bind: RefCell::new(None),
                 appwindow_righthanded_bind: RefCell::new(None),
                 canvas_touch_drawing_handler: RefCell::new(None),
 
@@ -173,8 +177,10 @@ mod imp {
             let canvas_touch_drawing_handler = self.canvas.connect_notify_local(
                 Some("touch-drawing"),
                 clone!(@weak obj as canvaswrapper => move |_canvas, _pspec| {
-                    // Disable the zoom gesture when touch drawing is enabled
+                    // Enable the touch drag gesture, disable the zoom gesture and kinetic scrolling when touch drawing is enabled
                     canvaswrapper.imp().canvas_zoom_gesture_update();
+                    canvaswrapper.scroller().set_kinetic_scrolling(!canvas.touch_drawing() && canvaswrapper.inertial_scrolling());
+                    canvaswrapper.canvas_touch_drag_gesture_enable(canvas.touch_drawing());
                 }),
             );
 
@@ -207,6 +213,9 @@ mod imp {
                     glib::ParamSpecBoolean::builder("block-pinch-zoom")
                         .default_value(false)
                         .build(),
+                    glib::ParamSpecBoolean::builder("inertial-scrolling")
+                        .default_value(true)
+                        .build(),
                 ]
             });
             PROPERTIES.as_ref()
@@ -216,6 +225,7 @@ mod imp {
             match pspec.name() {
                 "show-scrollbars" => self.show_scrollbars.get().to_value(),
                 "block-pinch-zoom" => self.block_pinch_zoom.get().to_value(),
+                "inertial-scrolling" => self.inertial_scrolling.get().to_value(),
                 _ => unimplemented!(),
             }
         }
@@ -238,7 +248,16 @@ mod imp {
                     self.block_pinch_zoom.replace(block_pinch_zoom);
                     self.canvas_zoom_gesture_update();
                 }
+                "inertial-scrolling" => {
+                    let inertial_scrolling = value
+                        .get::<bool>()
+                        .expect("The value needs to be of type `bool`");
 
+                    self.inertial_scrolling.replace(inertial_scrolling);
+
+                    self.scroller
+                        .set_kinetic_scrolling(!self.canvas.touch_drawing() && inertial_scrolling);
+                }
                 _ => unimplemented!(),
             }
         }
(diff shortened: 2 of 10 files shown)


rnote-ui/src/canvaswrapper.rs
```diff
@@ -514,6 +536,16 @@ impl RnCanvasWrapper {
         self.set_property("show-scrollbars", show_scrollbars.to_value());
     }
 
+    #[allow(unused)]
+    pub(crate) fn inertial_scrolling(&self) -> bool {
+        self.property::<bool>("inertial-scrolling")
+    }
+
+    #[allow(unused)]
+    pub(crate) fn set_intertial_scrolling(&self, inertial_scrolling: bool) {
```
flxzt, line 545: here's a typo: intertial -> inertial

rnote-ui/src/canvaswrapper.rs
```diff
@@ -167,8 +171,10 @@ mod imp {
             self.canvas.connect_notify_local(
                 Some("touch-drawing"),
                 clone!(@weak obj as canvaswrapper => move |canvas, _pspec| {
-                    // Disable the zoom gesture when touch drawing is enabled
+                    // enable the touch drag gesture & disable the zoom gesture and kinetic scrolling when touch drawing is enabled
+                    canvaswrapper.scroller().set_kinetic_scrolling(!canvas.touch_drawing() && canvaswrapper.inertial_scrolling());
                     canvaswrapper.canvas_zoom_gesture_enable(!canvas.touch_drawing());
+                    canvaswrapper.canvas_touch_drag_gesture_enable(canvas.touch_drawing());
```
flxzt, line 177: Okay, you're right, leaving it on when kinetic-scrolling is enabled doesn't do any harm. Explaining this deliberate decision in a comment here would be helpful for my forgetful mind

flxzt:
well, you beat me :)

rnote-ui/src/canvaswrapper.rs
Kneemund: > leaving it on when kinetic-scrolling is enabled doesn't do any harm

But kinetic-scrolling is only enabled if touch drawing is disabled, meaning that the gesture is only enabled if kinetic-scrolling is disabled. Or is that what you meant?

flxzt:
sorry, the comment I made was a bit unclear: it was only referring to the configuration when kinetic-scrolling is desired, when it is optional this needs to be different.
The reason this works in the PR already (after some logging) is that touch-events apparently are also captured by the dedicated drag-empty-area gesture (originally only intended for mouse & stylus input). I might refactor to combine touch-drag and drag-empty-area, because they are overlapping in functionality at this point.

Summarizing the cases (without testing):

| inertial-scrolling| touch-drawing | kinetic-scrolling | touch-drag gesture |
| --- | --- | --- | --- |
| true | true | = disabled | = enabled |
| true | false | = enabled | = either |
| false| true | = disabled | = enabled |
| false | false | = disabled | = enabled |

rnote-ui/src/canvaswrapper.rs
flxzt: I'll resolve, this works but disabling it at all might be unnecessary (see other comment)

rnote-ui/src/appwindow/appsettings.rs
```diff
@@ -63,6 +63,24 @@ impl RnAppWindow {
...
+        // inertial scrolling
+        self.app_settings()
+            .bind(
+                "inertial-scrolling",
+                &self.settings_panel().general_inertial_scrolling_switch(),
+                "active",
+            )
+            .get_no_changes()
+            .build();
+        self.bind_property(
```
flxzt, line 75: It would be better to keep the settings binds as simple as possible, and instead use the `touch-drawing` property to bind the sensitivity. (Also, I'd make the entire row insensitive instead of just the button)

Also: do we need this? It's might not be clear to the user why the setting becomes insensitive and how to make it sensitive again.

Kneemund: > But: do we need this? It might not be clear to the user why the setting becomes insensitive and how to make it sensitive again.

I thought about that too, but it has the benefit of showing the user that the option has no effect due to their current configuration. Leaving it sensitive would seem like a bug to me. I guess this comes down to either leaving some users wondering why it doesn't work, or leaving some users wondering why it's not sensitive.

Kneemund:
> The reason this works in the PR already is that touch-events apparently are also handled by the dedicated drag-empty-area gesture (originally only intended for mouse & stylus input).

... right, I *somehow* disregarded that kinetic-scrolling disables touch scrolling altogether, not just the inertial aspect of it, even though you said that in the issue.

---

We could just leave the touch drag gesture enabled at all times then. Or would you prefer disabling it in the "either" case?

rnote-ui/src/appwindow/appsettings.rs
Kneemund: Alright, I'd change the subtitle to "Set whether touch scrolling on the canvas is inertial. Always disabled if touch drawing is enabled." then.

Kind of related: Do we want to add punctuation to the subtitles of the other options too? If not, then this one and the DPI setting would be the only ones that have it (because they contain multiple sentences).

flxzt: Sounds good to me :)

Check out the [Gnome HIG writing style](https://developer.gnome.org/hig/guidelines/writing-style.html) (last paragraph), if I understand it correctly it suggests no punctuation if it is a single sentence

flxzt: Well I don't know. I think the guideline got updated recently as well, maybe it used to be different. But I know at least Weblate warns translators if the punctuation differs in the translated string.

And I am also considering the human factor: I don't want to change too much for translators once again, it might be frustrating for them

Kneemund: Yeah, fair enough.

flxzt:
I think the two sentences in the subtitle need a manual line break

Testing this, there is one weird issue remaining:

When toggling `touch-drawing` and `inertial-scrolling` a few times, moving and zooming the canvas while doing it and arriving at `inertial-scrolling` = disabled, it sometimes gets "stuck" where touch-drag isn't working anymore.

This might be an odd interaction between the the different touch gestures claming event sequences but not denying them or something like that. I remember experiencing the same thing before between the zoom, touch-drag, two-finger-long-press gestures, I hope this is not something that can only be fixed in the scroll-gesture in the scrolledwindow.

PJansky:
I'm glad to see my issue being addressed. I'm generally not a fan of having to many settings and toggles for applications as it often becomes overwhelming. From what I've gathered in previous comments by @flxzt the need for the toggle stems from the upstream bug in GTK, which can be worked around be selectively enabling with this toggle. Is that correct? Is it possible to see how Xournalpp is doing this or are they not affected by this problem?

Kneemund:
Well yeah, I guess we could work around this issue by just not making it an option, and simply binding `kinetic-scrolling` to `touch-drawing`. Or I can look into what's causing this bug, but I wasn't able to reproduce it yet (I haven't tried for long though).

> the need for the toggle stems from the upstream bug in GTK, which can be worked around be selectively enabling with this toggle

No, it's not really a bug I think. Just a badly named property, which needs to be disabled in order for touch drawing to work. There's no need for the toggle though.

flxzt:
can you rebase, then I'll test / debug it again

flxzt:
I think it doesn't hurt. I can see that some want to keep it disabled because touch-zooming only works when the inertia is "finished".

Btw. the PR is now missing something like:

```rust
            obj.connect_notify_local(
                Some("inertial-scrolling"),
                clone!(@weak canvas => move |canvaswrapper, _pspec| {
                    canvaswrapper.scroller().set_kinetic_scrolling(!canvas.touch_drawing() && canvaswrapper.inertial_scrolling());
                }));
```

in canvaswrapper's `constructed()`, no?

flxzt:
I just stumbled upon [gtk#5863](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/5863), and [this code](https://gitlab.gnome.org/YaLTeR/identity/-/blob/5544fda73cb7e3264ef40975d56f5a27a4d74920/src/page.rs#L285). Possibly the fix for the lock-ups?

Kneemund:
Yep, that looks promising. I can implement the potential fix, but I'm unable to reproduce the bug to begin with - it would be nice if you could test it :)

Works like a charm for me too - I haven't tried it on X11 yet. Maybe we should enable kinetic scrolling on Wayland and disable it on X11? 

Not sure how much of a priority the experience on X11 is anyway, considering the pressure sensitivity issues with GTK4 etc.

flxzt:
> I can implement the potential fix, but I'm unable to reproduce the bug to begin with - it would be nice if you could test it :)

I did, unfortunately it did not resolve. At this point I think I should look into GtkScrolledWindow's code to try to understand exactly what is happening.

> Not sure how much of a priority the experience on X11 is anyway, considering the pressure sensitivity issues with GTK4 etc.

Yeah, at lot of things don't work as expected, are buggy, etc. especially with input. But as long as wayland on nvidia does not work fully (my dev setup: x1yoga connnected to an Nvidia eGPU) I think it should at least not have bugs that break the app

flxzt:
you know, what came to my mind: because the lock-up only happens after using dragging, and then toggling the option, we could just send a toast requesting an app restart after toggling this option. And maybe add "A restart is required" in the subtext. I think with that I'd be fine merging it, so we don't have to wait for a fix on gtk's side.

Kneemund:
Yeah, that could work. Toggling touch drawing while this option is enabled would have to request a restart/dispatch the same toast, unless that doesn't cause lockups.

flxzt:
No, it only happens when toggling the inertial-zoom option

Kneemund:
Hm, that's strange. It uses the exact same function for updating the property. But yeah, if that's the case, then a "restart required" toast & description should be fine. 

Do we still want to update the kinetic-scrolling property when toggling the option, or only apply the stored value in the `RnCanvasWrapper` constructer (so that lock-ups can't happen and you *have to* restart for the change to apply)?

flxzt:
okay, thats true, but this lock up really only comes apparent when it is not "masked"  by either: the kinetic scrolling and "touch-drawing" gesture. To be more precise: the transition of the inertial-scrolling option from on->off would makes this bug apparent, if the user already has set "touch-drawing" to off. If not, then the transition of "touch-drawing" from on->off would make it apparent, after disabling inertial-scrolling. So it should even only be necessary to prompt for a restart when disabling "inertial-scrolling", right?

> Do we still want to update the kinetic-scrolling property when toggling the option, or only apply the stored value in the RnCanvasWrapper constructer (so that lock-ups can't happen and you have to restart for the change to apply)?

We can update right away because a restart might not be necessary on every machine (Is is still only me that experiences this?) and not at all on wayland. (but we don't need to communicate that to users, to avoid unnecessary confusion)

flxzt:
oh I think this has some conflicts now, should I quickly resolve them?

Kneemund:
I'm working on that, but you can also resolve them if you're at it too :)

rnote-ui/src/canvaswrapper.rs
```diff
@@ -256,6 +272,16 @@ mod imp {
...
             }
         }
+
+        fn canvas_kinetic_scrolling_update(&self) {
+            self.scroller.set_kinetic_scrolling(
+                !self.canvas.touch_drawing() && self.inertial_scrolling.get(),
+            );
+
+            // Workaround for https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/5863.
+            self.scroller.queue_allocate();
```
Kneemund, line 282: Do we want to keep this? I'm assuming that it didn't do anything, so we should probably remove it.

flxzt: yeah, I think so too

PR #601, opened 2023-03-31
isometric background patterns

### LINKED ISSUE #600: Isometric background patterns
Add the option to select an isometric grid or" isometric dotted" as a background pattern. For implementation, every second line could simply be offset.

**Additional context**
The isometric perspective makes it easy for artists in particular to draw spatially, as shown here:

Kneemund:
- Closes #600.
- This PR adds "Isometric Grid" and "Isometric Dots" background patterns. It modifies the tile size calculation based on the selected pattern, which is necessary for seamless tiling. It also forces the use of the maximum tile size if None or Lines is selected.

|Isometric Grid|Isometric Dots|
|---|---|
|grafik|grafik|

Like all of the existing patterns, the dots/lines are not exactly centered at the origin, but are offset by half of their width in order to fully fit into the SVG (removing the need of having another element at the opposite border). This can look a bit out of place.

### DIFF
--- a/rnote-engine/src/document/background.rs
+++ b/rnote-engine/src/document/background.rs
@@ -50,6 +50,13 @@ impl TryFrom<u32> for PatternStyle {
     }
 }
 
+/// 3_f64.sqrt()
+const SQRT_THREE: f64 = 1.7320508075688772;
+/// 3_f64.sqrt() / 2_f64
+const HALF_SQRT_THREE: f64 = SQRT_THREE / 2_f64;
+/// 3_f64.sqrt() / 4_f64
+const QUARTER_SQRT_THREE: f64 = SQRT_THREE / 4_f64;
+
 pub fn gen_hline_pattern(
     bounds: Aabb,
     spacing: f64,
@@ -142,8 +149,49 @@ pub fn gen_grid_pattern(
     group.into()
 }
 
+pub fn gen_dots_pattern(
+    bounds: Aabb,
+    row_spacing: f64,
+    column_spacing: f64,
+    color: Color,
+    dots_width: f64,
+) -> svg::node::element::Element {
+    let pattern_id = rnote_compose::utils::svg_random_id_prefix() + "_bg_dots_pattern";
+
+    let pattern = element::Definitions::new().add(
+        element::Pattern::new()
+            .set("id", pattern_id.as_str())
+            .set("x", 0_f64)
+            .set("y", 0_f64)
+            .set("width", column_spacing)
+            .set("height", row_spacing)
+            .set("patternUnits", "userSpaceOnUse")
+            .set("patternContentUnits", "userSpaceOnUse")
+            .add(
+                element::Rectangle::new()
+                    .set("stroke", "none")
+                    .set("fill", color.to_css_color_attr())
+                    .set("x", 0_f64)
+                    .set("y", 0_f64)
+                    .set("width", dots_width)
+                    .set("height", dots_width)
+                    .set("rx", dots_width / 3.0)
+                    .set("ry", dots_width / 3.0),
+            ),
+    );
+
+    let mut rect = element::Rectangle::new().set("fill", format!("url(#{pattern_id})"));
+    rect.assign("x", format!("{}px", bounds.mins[0]));
+    rect.assign("y", format!("{}px", bounds.mins[1]));
+    rect.assign("width", format!("{}px", bounds.extents()[0]));
+    rect.assign("height", format!("{}px", bounds.extents()[1]));
+
+    let group = element::Group::new().add(pattern).add(rect);
+    group.into()
+}
+
 fn calc_width_iso_pattern(spacing: f64) -> f64 {
-    spacing * f64::sqrt(3_f64)
+    spacing * SQRT_THREE
 }
 
 pub fn gen_iso_grid_pattern(
@@ -222,13 +270,33 @@ pub fn gen_iso_dots_pattern(
     bounds: Aabb,
     spacing: f64,
     color: Color,
-    dots_width: f64,
+    hexagon_height: f64,
 ) -> svg::node::element::Element {
     // spacing: side length of the equilateral triangle
     // pattern_width: two times the height of the equilateral triangle
 
     let pattern_id = rnote_compose::utils::svg_random_id_prefix() + "_bg_iso_dots_pattern";
     let pattern_width = calc_width_iso_pattern(spacing);
+
+    let hexagon_path = |x_offset: f64, y_offset: f64| {
+        element::path::Data::new()
+            .move_to((
+                x_offset + QUARTER_SQRT_THREE * hexagon_height,
+                y_offset + hexagon_height,
+            ))
+            .line_to((
+                x_offset + HALF_SQRT_THREE * hexagon_height,
+                y_offset + 0.75 * hexagon_height,
+            ))
+            .line_to((
+                x_offset + HALF_SQRT_THREE * hexagon_height,
+                y_offset + 0.25 * hexagon_height,
+            ))
+            .line_to((x_offset + QUARTER_SQRT_THREE * hexagon_height, y_offset))
+            .line_to((x_offset, y_offset + 0.25 * hexagon_height))
+            .line_to((x_offset, y_offset + 0.75 * hexagon_height))
+            .close()
+    };
 
     let pattern = element::Definitions::new().add(
         element::Pattern::new()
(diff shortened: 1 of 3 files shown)


flxzt:
Nice, thanks!

> but are offset by half of their width in order to fully fit into the SVG (removing the need of having another element at the opposite border). This can look a bit out of place.

I can't think of a good alternative, can you? Splitting the dots / lines at the border doesn't work well, in certain zoom levels

Only one subjective thing to discuss: should the dots be rounded squares (like you implemented) or circles?

ElbertCode:
> should the dots be rounded squares (like you implemented) or circles?

Personally, I would prefer **circles** because they ...

1. ... have the simpler shape. Since the dots only form the background, their shape should be as simple as possible. Besides, the dots are rendered so small most of the time anyway that the rounded squares also look like circles and can therefore be just that.

2. ... are the same from all angles. With rounded squares, if you draw the line from one of the diagonally lying neighbouring points, you "hit" one of the corners (or actually even slightly off), while from the neighbouring point above or below, you "hit" a side.

But since the dots will be so small most of the time that you won't be able to tell the difference, rounded squares wouldn't bother me either.

Kneemund:
> Only one subjective thing to discuss: should the dots be rounded squares (like you implemented) or circles?

The dots are the same as the ones used in the regular Dots pattern. IMO, if we use circles here, then the Dots pattern should too.

I personally don't mind either option, I just went for consistency :P

flxzt:
> I personally don't mind either option, I just went for consistency :P

I'd argue triangles with rounded corners would be the most consistent, then the sides align with the pattern they form!

> EDIT: Looks like that would work. I can't tell if the format borders also need to be centered, or if they already are. I think they're already centered, since offsetting them by half their width makes everything look way worse.
(I'll create some constants for the line and dot width to reduce the amount of magic numbers.)

sounds good. We just need to make sure the export is exactly the same, because users might align content with the patterns.  
Oh and strictly this is a breaking change in the format, at least visually. If we want to be pedantic we need a conversion when opening older files that translates the content by pattern_size * 0.5 so that it stays aligned with the pattern.

ElbertCode:
> what about non-rounded hexagons with size 2.0:

Shouldn't the hexagons be slightly rotated so that the sides face the neighbouring dots, as is the case in the normal dots background?

Anyway, in my opinion hexagons are just too complex for small background dots, so it looks a bit strange, especially if they are not rounded. Anything more complex than squares (which don't really fit for an isometric layout) and circles should be avoided, and this is why these are the only two shapes used either on real paper or in apps like Xournal++. So I would stick to circles and would also change them in the normal dots background for consistency.

flxzt:
> The hexagons would have to be rotated, yeah

why? The sides are oriented so they align  along the dimensions of a isometric perspective. There is no dimension that goes horizontally

> I think they aren't equilateral here

Ah my mistake, I implemented the first search result I found instead of thinking about it myself :smile: . Then like this, if I calculated it correctly:

Bildschirmfoto vom 2023-04-06 15-45-48


Code:

```rust
    /// 3_f64.sqrt() / 2_f64;
    const HALF_SQRT_THREE: f64 = 0.866;
    /// 3_f64.sqrt() / 4_f64;
    const QUARTER_SQRT_THREE: f64 = 0.433;

    let hexagon_path = |x_offset: f64, y_offset: f64| {
        element::path::Data::new()
            .move_to((
                x_offset + QUARTER_SQRT_THREE * hex_height,
                y_offset + hex_height,
            ))
            .line_to((
                x_offset + HALF_SQRT_THREE * hex_height,
                y_offset + 0.75 * hex_height,
            ))
            .line_to((
                x_offset + HALF_SQRT_THREE * hex_height,
                y_offset + 0.25 * hex_height,
            ))
            .line_to((x_offset + QUARTER_SQRT_THREE * hex_height, y_offset))
            .line_to((x_offset, y_offset + 0.25 * hex_height))
            .line_to((x_offset, y_offset + 0.75 * hex_height))
            .close()
    };

```

> I looked at the existing patches yesterday, and it looks like making a self-contained patch would be quite tedious in this case

yeah it's a bit painful. So its not a big deal to leave it for now IMHO

ElbertCode:
>why? The sides are oriented so they align along the dimensions of a isometric perspective. There is no dimension that goes horizontally

Because currently a **corner** of the hexagon faces the neighbouring points. With the normal dot background, however, a **side** of the rectangle faces the neighbouring dots.

Kneemund:
> why? The sides are oriented so they align along the dimensions of a isometric perspective. There is no dimension that goes horizontally

Ah, I thought that you wanted to make the sides align with the sides of the neighboring elements. "[...] the sides align with the pattern they form" sounded like that to me :P

> with size 2.0

Do we want to increase the size of the Dots pattern too? That sounds quite large BTW.

> yeah it's a bit painful. So its not a big deal to leave it for now IMHO

The only way that this change could "break" things is if people strictly aligned the content to the Dots pattern. The line offset (i.e. every other current pattern) should be negligible. On the other hand, the Dots pattern is the default one. I'd leave it up to you if we need a patch for this or not :P

Kneemund:
> Because currently a **corner** of the hexagon faces the neighbouring points. With the normal dot background, however, a **side** of the rectangle faces the neighbouring dots.

Well, currently, the dots have the shape of the pattern they form. In other words, the dots surrounding a dot form a hexagon that looks like the dot itself. I think that this is what flxzt intended.

ElbertCode:
> Speaking of which, should we remove we rounded corners from that _if_ we decide to go with the hexagons here?

Well, *if* we decide to use hexagons, making them rounded, doesn't really makes sense, since they would then basically be circles at this small scale, as you already mentioned before. But if we make them without rounded corners, these should also be removed from the normal dots, for the sake of consistency.

flxzt:
Okay, this is drifting into a discussion where its about personal taste, which is always difficult.

Also: @ElbertCode I appreciate all your feedback and suggestions, but it is actually counter-productive to chime in too much while the feature is still being worked out

Just one point:

> Because currently a corner of the hexagon faces the neighbouring points. With the normal dot background, however, a side of the rectangle faces the neighbouring dots.

The dots pattern forms a 2 dimensional grid, where the sides are aligned with the dimensions of that grid.  That is entirely consistent. Where any corner face is 

Let's leave the dots for now, and focus on what we should implement here. Personally, I like the unrounded hexagons best, the size is debatable.

Kneemund:
I think that the hexagon shapes are nice and also more consistent, I'd go for them. There's not much of a difference between circles and hexagons anyway unless you zoom in a lot like you have in the screenshots.

On a different note, do we need/want a patch? If we don't, then I'd include the `tile_offset` stuff in this PR. And do we want to change/remove the rounding of the squares used in the Dots pattern? Or leave it as-is?

flxzt:
> I think that the hexagon shapes are nice and also more consistent, I'd go for them. 

perfect.

> On a different note, do we need/want a patch? 

Pragmatically it's probably fine without, since were only talking about max. 0.75 pt's.

I'd say, let's keep it separate either way. You mentioned you are offsetting in `gen_tile_image()`, that won't work when exporting. It already would need to be where the vector is generated

Kneemund:
Alright.

> You mentioned you are offsetting in `gen_tile_image()`, that won't work when exporting. It already would need to be where the vector is generated

Turns out we can simply apply the offset to the `x` and `y` attributes of the pattern. Unless I'm missing something, that works perfectly fine.

flxzt:
We could do that, but then there are noticable gaps in the patterns (see screenshot). The gaps are already there currently, but definitely less noticable when the pattern shapes are not split between tiles. The dots are also more blurry in certain zoom levels when offset by 3/4 of a pixel

Bildschirmfoto vom 2023-04-07 17-52-12

Kneemund:
Hm, fair enough. Could the background be drawn with shaders instead of manually tiling textures? The actual GLSL shouldn't be that hard to do, but I haven't worked with GSK before.

flxzt:
it might be possible, check the [gl area](https://github.com/gtk-rs/gtk4-rs/tree/master/examples/glium_gl_area/glium_gl_area) example in gtk-rs.

Personally, I am excited for [vello](https://github.com/linebender/vello), a GPU-accelerated 2d graphics renderer based on compute shaders and I plan to switch to it eventually for all of our drawing / rendering, both strokes and background :)

Kneemund:
Yeah, we should wait for that then :)
Just pushed the hexagon dots stuff, but I want to take a look at the calculation of those again.

Kneemund:
Okay NVM, looks correct to me :P
The `0.25` and `0.75` got me thinking, but the orientation is different than what I had in mind and we're using the diameter here.

flxzt:
cross-checking can't hurt!

Ok LGTM. I'll press the green button, unless you have anything else?

Kneemund:
Nope, this should be ready :)

PR #607, opened 2023-04-04
add size limits for grid builder

### LINKED ISSUE #591: Memory allocation error when drawing table
**Describe the bug**
When drawing the first cell of the table with table drawing tool, if you hold `ctrl` you can make the cell look like a straight line. (I assume it just has zero width/height). If you then try to do the second part of drawing the table, it will crash with: `memory allocation of 137438953472 bytes failed` and `SIGABRT`, which means that the program tried to allocate too much memory.

**Steps to reproduce**
- Take the table drawing tool
- Start drawing the first cell
- Hold `ctrl` and make the cell look like a straight line
- Do the second part of drawing the table
- Get crashed  

**Expected behavior**
It should not crash.
Additionally, I would expect that when you are drawing the first cell of the table, holding `ctrl` would never make the cell a straight line. Same with squares, ellipses, coordinate systems. When using those, user would almost never want to draw a straight line.

** Console Output **
```
 2023-03-29T14:54:12.136Z DEBUG rnote::app::imp > ... env_logger initialized
memory allocation of 137438953472 bytes failed
```
 - OS: Arch Linux
 - Version: rolling
 - Installation source: can reproduce on both flatpak and arch package
 - Desktop Environment: GNOME 43.2

Kneemund:
* Adds minimum dimensions for the first cell (2x2), which fixes #591.
* Adds maximum dimensions for the cell grid (500x500 cells).

### DIFF
--- a/rnote-compose/src/builders/gridbuilder.rs
+++ b/rnote-compose/src/builders/gridbuilder.rs
@@ -14,6 +14,9 @@ use super::shapebuilderbehaviour::{ShapeBuilderCreator, ShapeBuilderProgress};
 use super::ShapeBuilderBehaviour;
 use crate::Constraints;
 
+const GRIDBUILDER_FIRST_CELL_DIMENSIONS_MIN: f64 = 2.0;
+const GRIDBUILDER_CELL_GRID_DIMENSIONS_MAX: u32 = 500;
+
 #[derive(Debug, Clone, Copy)]
 enum GridBuilderState {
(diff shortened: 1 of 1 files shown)


flxzt:
actually, a maximum of 100x100 like you suggested is much more appropriate. I am also experiencing massive slowdowns / even crashes above that

flxzt:
thanks!

PR #612, opened 2023-04-08
improve typewriter keyboard navigation

### LINKED ISSUE #560: More keyboard navigation shortcuts in typewriter text box
The keyboard navigation in the typewriter text-box would be enhanced by some additional shortcuts:

| shortcut   | action                            |
|------------|-----------------------------------|
| ctrl+left  | cursor moves one word left        |
| ctrl+right | cursor moves one word right       |
| home       | cursor jumps to start of line     |
| end        | cursor jumps to end of line       |
| ctrl+home  | cursor jumps to start of text box |
| ctrl+end   | cursor jumps to end of text box   |

Kneemund:
- Closes #560.
- Implements all navigation keybindings mentioned in the issue and comment(s).
- Uses the same style that Chromium does, as described in my comment.
- ATM, quitting the selection using movement keys does not actually execute that movement. That behavior is improved:
  - move cursor to leftmost (rightmost) selected character on `NavLeft` (`NavRight`),
  - move cursor like when there's no selection on other key presses.
  
### Things for Discussion
- Word boundary/navigation style.
- If `move_cursor_text_start` and `move_cursor_line_start` (and the end variants) should be combined into one function like `move_cursor_forward` (and back), or if the latter should be split into two functions each. Or if it should stay as-is I guess.
- Is there any point in recreating the cursor instead of using `set_cursor()` when the text length does not change? `move_cursor_line_down` (and up) currently recreate it, but everything I've implemented in this PR does not. We should probably settle on a single way to do it, unless I'm missing something here.

EDIT: I could also implement the typical mouse navigation - i.e. double click to select whole word(s), triple click to select entire line(s) - but I don't know how to detect those yet. I'm assuming I could use `GestureClick`s, but everything related to it would have to be outside of the main `handle_pen_event_down` function then.

### DIFF
--- a/rnote-engine/src/strokes/textstroke.rs
+++ b/rnote-engine/src/strokes/textstroke.rs
@@ -9,7 +9,7 @@ use rnote_compose::transform::TransformBehaviour;
 use rnote_compose::{color, Color, Transform};
 use serde::{Deserialize, Serialize};
 use std::ops::Range;
-use unicode_segmentation::GraphemeCursor;
+use unicode_segmentation::{GraphemeCursor, UnicodeSegmentation};
 
 use super::strokebehaviour::GeneratedStrokeImages;
 use super::StrokeBehaviour;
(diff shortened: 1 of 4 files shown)


Kneemund:
Do you mind sending me the text you've used to test this? I know why this is happening, but I thought that the algorithms used here would always end up at char boundaries.

LeSnake04:
Also, there is an issue with texts using special characters like äöü: 

Peek 2023-04-11 22-45

LeSnake04:
> Do you mind sending me the text you've used to test this? I know why this is happening, but I thought that the algorithms used here would always end up at char boundaries.

I was just spamming some text. I am Currently trying to find a pattern

LeSnake04:
I think this is related to äöü: This is one of the strings it happend with
image

LeSnake04:
And it seems like it happens when going to the front of end of the text and pressing ctrl + arrow key there

Kneemund:
Yeah, I know what's happening and am working on it. Thanks for testing this though, it somehow never occurred to me that trying some special characters would be a good idea :)

LeSnake04:
The issue described above is fixed, thanks. 

I was still able to crash it another way, though: 

https://user-images.githubusercontent.com/47243761/231460332-510021c3-4860-4208-a571-4da81d7f43e7.mp4

I was pressing ctrl+shift+pos1, then ctrl+shift+end

```
thread 'main' panicked at 'called `Option::unwrap()` on a `None` value', rnote-engine/src/strokes/textstroke.rs:913:64
```

flxzt:
I think this looks all very good.

> Discuss [..] Word boundary/navigation style.

I didn't look at libreoffice's behaviour in detail, but this feels very intuitive

> If move_cursor_text_start and move_cursor_line_start (and the end variants) should be combined into one function like move_cursor_forward (and back), or if the latter should be split into two functions each. Or if it should stay as-is I guess.

The separation between text and line functions is good.

I think it would be better to simply write separate functions for all word operations and remove all `whole_word` arguments in the existing ones to keep everything cleaner and their naming consistent / more correct. `move_cursor_back()` and `forward()` have essentially a different body when `whole_word` evaluates true, and the others also don't have much shared code either. So:
  - `remove_grapheme_before_cursor()` -> `remove_word_before_cursor()`
  - `remove_grapheme_after_cursor()` -> `remove_word_after_cursor()`
  - `move_cursor_back()` -> `move_cursor_word_back()`
  - `move_cursor_forward()` -> `move_cursor_word_forward()`

> Is there any point in recreating the cursor instead of using set_cursor()? [..] We should probably settle on a single way to do it, unless I'm missing something here.

I don't think so, we could refactor to use `set_cursor()` everywhere where the text didn't change

> I could also implement the typical mouse navigation - i.e. double click to select whole word(s)

This can be separate. I used to have a WIP branch for this, which essentially implemented double-click gestures inside the engine. But it was messy, because they were dependent on the current state of the typewriter, and needed to be polled for the detection. My guess is that is will require some work to get that right

---

> Update: I tested a bunch of other non latin-based languages as well and there was no crash and RTL languages swap the direction (Great attention to detail!)

Thanks for testing!

Kneemund:
Alright, I've split those functions and refactored the parts that needlessly recreated the cursor.

rnote-engine/src/strokes/textstroke.rs
```diff
@@ -689,6 +689,48 @@ impl TextStroke {
...
 
+    pub fn remove_word_before_cursor(&mut self, cursor: &mut GraphemeCursor) {
+        let cur_pos = cursor.cur_cursor();
+        let prev_pos = self.get_prev_word_start_index(cur_pos);
+
+        if cur_pos == prev_pos {
+            return;
+        }
+
+        cursor.set_cursor(prev_pos);
```
flxzt, line 700: this is redundant, no? if we use prev_pos below

Kneemund: That's true. I'll also refactor the early return into an if block wrapping the rest of the body, like in the functions for removing a single grapheme.

flxzt:
Looks good to me! Thanks!

Issue #207, opened 2022-06-15
Mirror selection


Ambytes:
**Problem description**
I'm always frustrated when drawing diagrams especially mind maps and I notice that it would be much better to place a selection of the tree to the opposite side, then finding myself redrawing the branch. Also I think when importing graphics it might be useful in a lot of cases to be able to mirror an image so that for example a hand or a sign/symbol would point to the exact opposite direction.

**Considered solution**
It would be neat to have a button on the left panel under the selection tab which allows mirroring both horizontally and vertically. It might be considerable to wrap both options in a "dropdown"-menu so you don't bloat the panel too much. An icon could be a rectangle which is being split in either direction with a dotted-line or smth. similiar. Also it might be considerable to add buttons that appear below a selection after you've selected content that allow one to do things like cutting, pasting, mirroring, etc.

**Considered alternatives**
Well the only alternative method I've found is other than redrawing, having to export my selection then mirror it and then import it as an image again, which is a lot of hassle for such a "simple" user interaction.

TornaxO7:
I'd claim this issue.

NoAccount1:
Any news on this ?
I saw the PR was close due to lack of time and struggles.
I would really be interested in having such a feature

Yavuz-Kagan-Yadigar:
What is the status of that pull request? My 0.14 flatpak still does not seems to have it.

NXTler:
Anyone still working on this?

PR #614, opened 2023-04-09
Adding mirorring-feature for selected shapes.

TornaxO7:
This should fix https://github.com/flxzt/rnote/issues/207.


TornaxO7:
Where should be the "mirror-action"?
Should there be two action-buttons in my marked box?

image

Another idea would be, to have something like a single button on the bar which has the following features:
- The button displays either the horizontal or the vertical mirroring
- If it shows the horizontal mirroring for example, then the selected shapes will be mirrored horizontally
- If you hold it for a short time, two other buttons will pop up, where you can select horizontal or vertical mirroring

What do you think?

PhilDevProg:
> Where should be the "mirror-action"?
> Should there be two action-buttons in my marked box?

That would be the easiest and in my opinion best way to do this.

> Another idea would be, to have something like a single button on the bar which has the following features:
> - The button displays either the horizontal or the vertical mirroring
> - If it shows the horizontal mirroring for example, then the selected shapes will be mirrored horizontally
> - If you hold it for a short time, two other buttons will pop up, where you can select horizontal or vertical mirroring

I think it would be confusing to hold the button to change it's action and it takes longer. And what about users with a mouse or a pen? Holding a button with a mouse or a pen isn't something you normally have to do.

yuval-herman:
Perhaps a button placed directly inside the selection box could be more helpful and intuitive.
Example:
image

TornaxO7:
Ok, I've to sadly announce that I may not have the time to approach this PR for the time being since I still can't get a [dev-environment to working](https://github.com/flxzt/rnote/discussions/813). If anyone has the time to approach this: Feel free to do! I'm sorry for this announcement :(

PR #617, opened 2023-04-13
Improve exiting/closing logic

adih-20:
Hi,

This pull request addresses the following things I noticed:
- Rnote doesn't close the tab if the user tries to close a tab with an unsaved document and selects `Save` (and saves successfully)
- Exiting Rnote with files not saved to disk (and selecting `Save`) prompts Rnote to save the documents to the user's `Documents` directory

Specifically, the flow for exiting Rnote with files not saved to disk could be jarring to users as it differs from the flow for exiting a tab with files not saved to disk (when I first encountered this, I thought Rnote had crashed).

To remedy, this pull request makes the following changes:
- Moves some of the exiting logic into a new `filechooser_save_doc_as_and_exit` function
<details><summary>Details</summary>
<p>

I'm new to Rust/GTK/libadwaita, but I believe the reason Rnote fails to close the tab is due to the fact that `filechooser_save_doc_as` returns soon after `filechooser.show()` is called (to avoid blocking the UI thread). Conducting a check for unsaved changes immediately afterwards will almost always fail as the user might take some time to find the directory to save their document in.

As a result, I moved some of the exiting logic into the `connect-response` for the filechooser, so that the tab can exit once the file has successfully saved.

</p>
</details> 

- Changed Rnote window closing logic to accomodate Save-As

<details><summary>Details</summary>
<p>

Currently, Rnote prevents the user from saving their (unsaved to disk) work in the exit flow if a user-specific `Documents` directory is not found (as per my understanding) and defaults to a save path in the `Documents` directory.
If I'm understanding that portion of code correctly, this could be problematic in the edge case that an user does not have a `Documents` directory.

To remedy this, I changed the association of `None` for `save_folder_path` in `dialog_close_window()` to indicate that the current document has not been saved to disk and handled saving and exiting accordingly.

To handle the case where the user decides to cancel a save for unsaved document(s), a tab is closed as each save occurs until no more tabs marked to be saved are left, after which it closes the window. This allows the user to come back to their unsaved document(s) if they cancel.

</p>
</details> 

Thanks!
(Would appreciate any critiques/discussion/testing to make sure everything is good)


flxzt:
thanks for addressing this! I agree that the behaviour is not ideal.

But instead of displaying dialogs for every unsaved document when closing the window (How would the user know which file corresponds to which dialog? And it feels odd that there are many dialogs stacked on top of each other), I'd simply change the close-tab prompt: after a successful save the tab should then be closed. I think this is more intuitive and consistent to how other apps handle it (see for example gnome-text-editor, which where I got inspiration for the tab implementation)

adih-20:
Thanks for the feedback. I agree, opening multiple dialogs isn't intuitive for a user since it would be pretty unclear which document they were saving.

just to clarify: do you think that the default save to Documents directory behavior should be preserved, or should we continue with the Save-As? I was a bit unsure as to what you meant by changing the close-tab prompt.

Also, I am still a bit unsure if we should preserve the save to documents behavior if that is the preferred option. This is mainly because this behavior differs from the close-tab behavior (which opens a dialog for the user to save the file) while the exit flow just saves to the Documents directory and exits. Another concern I had was that the file(s) are saved as `New Document [ - X].rnote`, which overwrites any previously saved files named `New Document [ - X].rnote` in the Documents directory. I believe gnome-text-editor tries to get around the latter issue by making the filename (and document title) the first few letters of the document, but since Rnote is primarily drawing-based, I don't think something like that would work.

adih-20:
After further thought, if we are to preserve the save to documents behavior, should we consider changing the default save directory from Documents to the user's workspace directory and change the close-tab behavior to match? This would mitigate the overwriting problem and solve the discrepancy between the close-tab behavior and the close-window behavior.   

It would also make the default save directory during an exit user configurable instead of the current behavior.

flxzt:
> After further thought, if we are to preserve the save to documents behavior, should we consider changing the default save directory from Documents to the user's workspace directory and change the close-tab behavior to match? This would mitigate the overwriting problem and solve the discrepancy between the close-tab behavior and the close-window behavior.

Okay, sounds good.
We can change the output directory of the window-close dialog to the current active workspace, that might be more useful and intuitive for users.

And match the behaviour of window-close and tab-close . So close the active tab after (but only after) a successful save.

adih-20:
Alright, I will go ahead and do that. I think I'll go ahead and make a new PR for that and close this one (I think it would be cleaner for commit history)

flxzt:
Don't worry about that, I'll squash anyway

adih-20:
Hi,

I integrated the changes that were discussed.   

Now, the close-tab dialog will save to the current active workspace directory (I changed the UI of the prompt to better reflect the destination).

image

Additionally, the close-window prompt will save to the current active workspace directory.

image

I also noticed that the default suffixing behavior to prevent collisions sometimes did not work (as seen below with the latest release of Rnote `v0.5.18`), so I replaced it with logic that prevents Rnote from overwriting existing files (by creating a unique filename with a running index) to mitigate the file collision issue.

image

flxzt:
Looks good, thanks!

I'll only quickly tweak two things regarding the UI: making the check button `sensitive = false` instead of `can_target = false` and adjust the margins.

I also only made some cosmetic code changes: improved readability by adding some space and removed the code nesting while iterating over the files by continuing if the file is already saved.

Issue #155, opened 2022-05-05
Allow selection on partial overlap


FriederHannenheim:
When selecting things using the freehand or square tool the stroke must be fully selected. The invisible stroke boundary is bigger than the actual stroke so you need to have some padding when selecting things. This seems unintuitive. The Apple notes app on the IPad selects things that are only partially in the selection area. This way seems better. 
**The solution**
Maybe all strokes where more than 50% of their area is in the selection zone should be selected. 
**Problems with this solution**
Big strokes could be accidentally selected if more than 50% of their area is selected. Some clever math could be done with the total are of the selection area and the stroke are to see if the selection makes sense.

flxzt:
Can you take a recording of the behaviour in the apple notes app? I don't have an apple device.

But I think this is a good idea as an additional selector mode and should not be too hard to implement.

Also: Recently I improved the bounding /hitboxes for the strokes and shapes to make them tighter, especially for lines and ellipses and the new curve shapes. So that should improve the existing selector modes a little bit.

FriederHannenheim:
I don't have an Apple device either I just saw one of my classmates taking notes

PhilDevProg:
> But I think this is a good idea as an additional selector mode and should not be too hard to implement.
@flxzt I don't think that this should be a new selector, it should be an option for the rectangle and freehand selectors in my opinion that could be placed under the 4 selector modes and would be grayed out when using the 2 other selectors.

ElbertCode:
> it should be an option for the rectangle and freehand selectors in my opinion that could be placed under the 4 selector modes and would be grayed out when using the 2 other selectors.

Maybe this could even be done with a configuration menu, where you can set the percentage overlap needed for a selection and, if implemented, a toggle whether a more complex determination algorithm should be used.

dennisorlando:
on it

PR #630, opened 2023-04-19
Allow partial overlap (Selector)

dennisorlando:
Hi, I addressed the comments in #155 by adding a new toggle on the sidebar when using the selector tool.
- it works with rectangle selection and "freehand" selection 
- I implemented it by checking if the selection intersects with hitboxes instead of containing them
- missing: it is not greyed when "select intersecting path" or "select one by one" are ennabled as suggested by one of the issue's comments
- missing: there's no "percentage of overlapness" customization
- missing: somebody has to set an icon to it because I have no idea how xD (it's currently using the "lock aspect ratio" one)

### DIFF
--- a/rnote-engine/src/strokes/brushstroke.rs
+++ b/rnote-engine/src/strokes/brushstroke.rs
@@ -8,7 +8,7 @@ use crate::{
 };
 use p2d::bounding_volume::{Aabb, BoundingVolume};
 use piet::RenderContext;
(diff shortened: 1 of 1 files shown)


flxzt:
Thanks! I think this looks good, I only would name the widget id differently, appending `_togglebutton`. Can you rebase onto main?

The missing icon is not an issue, I can create one.

PR #632, opened 2023-04-20
add a new drawing

puppax:


PhilDevProg:
You should revert the first commit (Add files via upload https://github.com/puppax/rnote/commit/e9e6f09fe9762ca889d24bd5526b8fd046f1a83b ). It adds the drawing to the main folder of the branch which isn't where it belongs. It also is already present in misc/drawings so it's unnecessary.

PR #646, opened 2023-04-25
Adding Farsi/Persian translation under fa.po file

iamamirhe:


iamamirhe:
Currently, the translation work is not yet finished and only about 20% of it has been completed. If it is added to the main repository and the website GUI is accessible for further work, it will significantly expedite the process toward completion.

flxzt:
Sounds good, thanks!

PR #650, opened 2023-04-26
add dist archive CD

flxzt:
This adds a workflow that generates the dist archive (tarball) automatically on release.

`meson dist` is run with `--no-tests`, since we already have the CI for that. I am opening this as a draft because I want to add the CI is a "dependency" that needs to be completed successfully before this.

We should probably do the same for the `release-windows` workflow.

### DIFF
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -19,12 +19,11 @@ jobs:
       - name: Install toolchain
         id: toolchain
         uses: dtolnay/rust-toolchain@stable
-      - name: Configure
+      - name: Setup
         run: |
           meson setup --prefix=/usr _mesonbuild
-          meson configure _mesonbuild
       - name: Run meson dist
-        run: meson dist -C _mesonbuild
+        run: meson dist --no-tests -C _mesonbuild
       - name: Upload dist archive (Workflow Artifact)
         uses: actions/upload-artifact@v3
         with:
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -1,6 +1,7 @@
 on:
   release:
     types: [published]
+  workflow_dispatch:
 
 name: Generate Dist Archive
 
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -18,7 +18,7 @@ jobs:
       - name: Install dependencies
         run: |
           sudo dnf upgrade --refresh -y
-          sudo dnf install -y gcc gcc-c++ clang python3 make cmake meson git kernel-devel gtk4-devel libadwaita-devel poppler-glib-devel poppler-data alsa-lib-devel libappstream-glib desktop-file-utils
+          sudo dnf install -y gcc gcc-c++ clang python3 make cmake meson git gh kernel-devel gtk4-devel libadwaita-devel poppler-glib-devel poppler-data alsa-lib-devel libappstream-glib desktop-file-utils
       - name: Install toolchain
         id: toolchain
         uses: dtolnay/rust-toolchain@stable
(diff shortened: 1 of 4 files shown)


Kneemund:
We could achieve that like described [here](https://stackoverflow.com/a/71489231/11075203), although I'm not sure if the current setup would run the CI workflow twice. If it does, we could use a single workflow for building release assets, containing multiple jobs (one for the windows installer, one for the dist archives).

flxzt:
> although I'm not sure if the current setup would run the CI workflow twice

That seems to be the case: https://github.com/flxzt/rnote-testing/actions

I am not sure if it is desirable to pack that into one action. Especially when I want to extend it with flatpak CD as well. I think I would rather have them separate, even it that means we run the CI multiple times

However the action is currently failing in the `meson dist` job. It needs the full git repo, so I need to fix that too

Kneemund:
> I am not sure if it is desirable to pack everything for release into a single action. Especially when I want to extend it with flatpak CD as well. I think it would be better to have them separate, even it that means the CI is being run multiple times

We could use some sort of "uber-workflow" for CD that has multiple jobs which just call a [reusable workflow](https://docs.github.com/en/actions/using-workflows/reusing-workflows) for each target (dists, windows, flatpak). That way, we can still separate them into individual files, but only need to run CI once.

flxzt:
That would be a very good solution. I think I'll just remove the CI dependency again in this PR, and follow up with a refactor that implements the "uber-release-workflow"

.github/workflows/dist.yml
```diff
@@ -0,0 +1,50 @@
...
+jobs:
+  dist:
+    runs-on: ubuntu-22.04
+    container: fedora:38
+    permissions:
+        contents: write # needed for uploading release artifact
+    steps:
+      - name: Register archive name
+        id: register_archive_name
+        shell: bash
```
Kneemund, line 16: I don't think that this is needed here. It's needed in the windows installer workflow because the default shell is MSYS2 bash there, which can't write to `GITHUB_OUTPUT` properly.

Kneemund (commented):
Looks good to me :)

I'm assuming that fetching the entire history is done for distributing it alongside the code, and that adding the workspace to the safe directories is necessary for the meson target (?).

.github/workflows/dist.yml
flxzt: I see, thanks. Btw it is kind of unpleasant that we can only "match" to the tarball file naming, but I don't think there is a way to set the archive name by us in meson. So this workflow actually only works if the tag matches the version in meson

Kneemund: Hm, maybe we can at least retrieve it from the file system instead of hardcoding it? We should be able to write the path of files that match a glob to the step output, and then use that in the following steps.

Kneemund: After running `meson dist`, we could use the output of `ls _mesonbuild/meson-dist/*.tar.xz` and `ls _mesonbuild/meson-dist/*.tar.xz.sha256sum` to `$GITHUB_OUTPUT` (or stricter if there are multiple files with those extensions).

flxzt: Thats true, there shouldn't be any files in there, unless we would start using caching. I refactored it like you suggested

flxzt:
> I'm assuming that fetching the entire history is done for distributing it alongside the code, and that adding the workspace to the safe directories is necessary for the meson target (?).

Yes, `meson dist` needs a full git repo including history, and if we don't add the workspace as safe directory (in the container) the "checkout" action will not fetch the `.git` folder

Seems to be doing what we want: https://github.com/flxzt/rnote-testing/actions/runs/4870681988

Thanks for reviewing @Kneemund !

PR #651, opened 2023-04-26
File Recovery

### LINKED ISSUE #545: Auto Save shall also save drafts or add an option for it please
Auto Save shall also save drafts or add an option for it please

auto save on, battery empty, file not saved, draft gone

Option like checkbox for save drafts

Remind user to save

**Additional context**

gone work

LeSnake04:
closes #545

This adds an automatic save in /tmp/rnote-recovery/  

The status of the recovery save is indicated by a dot behind the file name  

DISCUSSIONS:
- [ ] seperate timers for recovery and autosave?
- [ ] do we need an autosave indicator

TODO:
- [ ] fix crash at launch 
- [ ] preserve state of file recovery state/delay settings between restarts
- [ ] avoid making text cursive on recovery save
- [ ] recovery dialog


rnote-ui/src/canvas/imexport.rs
```diff
@@ -302,15 +302,25 @@ impl RnCanvas {
 
         if let Err(e) = res {
             self.set_save_in_progress(false);
+            if self.recovery_in_progress() {
+                self.set_recovery_in_progress(false)
+            }
 
             // If the file operations failed in any way, we make sure to clear the expect_write flag
             // because we can't know for sure if the output_file monitor will be able to.
```
LeSnake04, line 310: @flxzt How important if this, when the user isn't supposed to access the folder where the file gets saved (as in the case of recovery files). Do I need to make another monitor for the recovery file?

flxzt: Sorry I didn't look at the code too closely so far since this you marked it as a draft, I don't really know

flxzt:
I just saw that the recovery mechanism is essentially a second autosave feature, even with a second indicator visible to the user. I am sceptical if it should be done this way, I would rather have an implementation that saves a file when unrecoverable errors are detected or a crash/panic can somehow be caught (using `catch unwind` maybe?). If this is not possible, a periodic save should at least be hidden from the user, everything else is confusing. There is also the concern about the save time because it takes quite a while currently for very large documents. And the most important part here is to have a dialog that detects if a crash happened on startup (maybe check if any files exists in a dedicated directory, we should probably use [directories::data_dir()](https://docs.rs/directories/latest/directories/struct.ProjectDirs.html#method.data_dir) to adhere to xdg specification here) and prompts if those files should be recovered.

LeSnake04:
I am currently moving the indicator to a less obvious place, so the users dont get confused, but still can confirm it if interested (to the recovery settings, to be specific)

I am not sure how possible is it to catch an error and this wont cover full system crashes. So I dont think there is an alternative to a second autosave unfortionatly. We can consider don't doing recovery saving when the file is saved autosave is enabled. 
I will defiantly turn off recovery if both the recovery and autosave timers are the same, since this caused issues in testing already.

The dialog is planned. I will do something similar to the edit workspace dialog.  I will work on that after I moved the indicator.

LeSnake04:
One issue I couldn't fix myself is that the recovery state and delay is not being saved between sessions. 

Also for some reason meson doesnt detect changed in rnote-ui/dialogs/recovery.rs

LeSnake04:
I tried to make the confirm overwrite document dialog as generic as possible so it can be reused for something like the filebrowser.

flxzt:
Alright, I like the recovery dialog, it provides useful options for the to-be-recovered data. There are some improvements that can be made with regards to the user experience, but it already roughly does what I think users would expect.

I am still sceptical about exposing options to the user. When would a user want to disable the recovery mechanism? If we have this feature, it should be enabled all the time and work in the background.

And I think it still is worth exploring if it is possible to wrap the entire application in a [catch_unwind](https://doc.rust-lang.org/std/panic/fn.catch_unwind.html). Here a small snippet to illustrate how I think it could work:

```rust
use std::sync::{Arc, RwLock};

fn main() -> std::thread::Result<()> {
    let state = Arc::new(RwLock::new(String::from("no need to panic.")));
    let res = std::panic::catch_unwind(|| {
        will_panic(Arc::clone(&state));
    });
    println!("this state that can be saved into a file for recovery: {}", state.read().unwrap());
    res
}

fn will_panic(state: Arc<RwLock<String>>) {
    *state.write().unwrap() = String::from(
        "
    this can be periodically written to with recovery state.
    As long as the panic does not happen while we write to the RwLock,
    the state will be accessible after the panic is caught in catch_unwind",
    );
    panic!();
}

```

AndreasVerden:
What is blocking this? Just had rnote crash on me and lost a fair bit of work

PR #670, opened 2023-05-13
feature: invert document colors

### LINKED ISSUE #635: Feature request: inverted pen color for export on documents with black background
### Discussed in https://github.com/flxzt/rnote/discussions/498

<sup>Originally posted by **liakhovitch** February  7, 2023</sup>
When exporting a file that has a dark background, it would be nice to have an "invert pen color" option along with the "remove background" option. This helps avoid the following scenario:
- I work on a homework assignment with a black background and white pen in order to save my eyes.
- When I export as a PDF so I can submit, I check "remove background".
- The exported file is now an all-white page

I don't know how OneNote does this, but there are a few variations on how this could work:
- Just swap any color exactly matching "black" and "white" on the default color palette (probably good enough for most cases)
- Invert the value of any color below a certain saturation

This would cover pretty much all of _my_ use cases, but it would also be nice to have the option to do this when switching backgrounds. Having an option to automatically invert both the background color and pen colors when switching between light/dark mode would also be helpful.

</div>

Kneemund:
- Closes #635.
- Should work fine functionally, but needs more discussion (and most likely more work) in terms of UX/UI.

---

Currently, this PR introduces an action which brings up a dialog that lets you choose between inverting the background colors and inverting all stroke colors in the current document.

Those two things are separated because:
- Changing the background color can't be undone using the history (you have to invert it again). That's because changes to the document settings generally aren't recorded ATM. 
- Changing the background color affects the settings and thus the defaults for creating new documents, making it more destructive than changing stroke colors.
- This feature is likely primarily used for printing. In that case, inverting a dark gray background would result in a light gray background - which is horrible for printing. The user should either export the document without the background or manually adjust the background colors.

This PR doesn't invert the colors automatically based on the selected app theme (nor does it expose an option for that). I don't think that it's feasible given Rnote's flexibility.

---

There are 3 general directions to take here.
- Invert both the background and strokes, and record changes to the document settings.
- Separate the actions (as they are right now).
- Always (and only) invert the strokes, and let the user handle the background manually.

---

A new dependency, [palette](https://github.com/Ogeon/palette), is added for converting between sRGB and (Ok)HSL. We could also implement the color conversion algorithms ourselves or use a different library.


flxzt:
Thanks! Functionality wise this is good, we can bring in palette, and maybe we should implement a [conversion](https://docs.rs/palette/0.7.1/palette/convert/index.html) from/into palette's [srgba](https://docs.rs/palette/0.7.1/palette/rgb/type.Srgba.html) for our custom color type. This is also what [roughr::Srgba](https://docs.rs/roughr/latest/roughr/type.Srgba.html) uses/re-exports (currently in an outdated version, I opened an [issue](https://github.com/orhanbalci/rough-rs/issues/12)).

Regarding the UX. I think there are a couple of things that are related to it, and some ideas I had (for discussion):
- We definitely should refactor to include the document and background into the history. I *think* the best way to do that is to move `StrokeStore` into `Document`, and move the history out of the store into the document. This way pens can still record/undo/redo via `EngineView`. This would also enable recording adding/removing pages for the fixed-size layout.
 It would need a slight modification to the file format / `EngineSnapshot`, but a conversion shouldn't be too difficult.
- A thing to consider: if we keep the document / stroke color inversion separate, we could think about having the stroke inversion as an action in the selector. This way it would be more explicit: "you have selected these strokes, so you are aware that some strokes might become very low-contrast to the background / even invisible (white on white, black on black).
- Alternatively/additionally: If this feature is intended to be mainly used for export/printing, what about an export option that automatically optimizes for it? Like: always set the background to white, and invert only those colors that become low-contrast/ brightness too similar to the background? I think there should be such a functionality in `palette`.
Of course this would work best if we had a preview in the export, or we keep having this as an action on the "live" document.

> This PR doesn't invert the colors automatically based on the selected app theme (nor does it expose an option for that). I don't think that it's feasible given Rnote's flexibility.

I think so too, I currently don't see a good way to do that automatically while not breaking edit flow other than to lock the colorpicker palette colors and background color entirely.

Kneemund:
> if we keep the document / stroke color inversion separate, we could think about having the stroke inversion as an action in the selector.

I also thought about that, but inverting only a part of all strokes seemed a bit pointless to me. It actually isn't though; it's useful for e.g. moving notes from a dark background onto a bright slide/image.

We could add an "invert background" button to the document settings to fully get rid of the action menu entry.

> If this feature is intended to be mainly used for export/printing, what about an export option that automatically optimizes for it? Like: always set the background to white, and invert only those colors that become low-contrast/ brightness too similar to the background?

Yeah, that's a good idea. Maybe we don't even need the background inversion at all.

Strokes on top of images would need special care. I imagine that we could just exclude all strokes whose bounding boxes intersect with images.

---

I agree with everything else too :)

On another note; I've found what seems like a much better color space for inverting the colors, HWB. Simply swapping the second and third component seems to give much more pleasing results (e.g. inverting yellow doesn't result in brown anymore). I haven't tested it much yet, but I'll upload a comparison in a few hours.

Kneemund:
> On another note; I've found what seems like a much better color space for inverting the colors, HWB.

I'm not too sure about that anymore. Here's a comparison.

|Original|Invert using Okhsl|Invert using Okhwb|
|---|---|---|
|Bildschirmfoto vom 2023-05-17 14-00-56|Bildschirmfoto vom 2023-05-17 14-11-30|Bildschirmfoto vom 2023-05-17 14-01-24|

In my opinion, HSL makes the colors too dark and produces unexpected results like yellow -> very dark brown. HWB has unexpected base colors, e.g. I wouldn't expect the default green color to be considered dark and get much brighter when inverting it.

The latter is also problematic for printing/exporting; but because we don't have the restriction that inverting twice should result in the original colors, we can just choose between the original and inverted color and pick the one with the highest contrast ratio compared to the background.

flxzt:
> We could add an "invert background" button to the document settings to fully get rid of the action menu entry.

Sounds good, this should be more intuitive.

> In my opinion, HSL makes the colors too dark and produces unexpected results like yellow -> very dark brown. HWB has unexpected base colors, e.g. I wouldn't expect the default green color to be considered dark and get much brighter when inverting it.

Interesting! After skimming over Wikipedia's [lightness](https://en.wikipedia.org/wiki/Lightness) article I learned that these are known [psychological effects](https://en.wikipedia.org/wiki/Lightness#Other_psychological_effects).

Color perception is a complex topic, to account for these effects we would probably also need to shift the color hue & saturation in certain cases & colorspaces. I am pretty sure we could spend a lot of time figuring out a "correct" approach :smile:  
Realistically, implementing it in a way that is "close-enough" is fine here in my opinion. At first glance Oklab would be a worth a try as well, but simply inverting it's `lightness` field did not bring good results after testing it a bit.

The question about the reversibility also affects this: in my opinion, it is not an issue when we optimize the colors in a print/export option, but it might be undesirable when it is an action on the live document because it would "degrade" it (or we would need to name the action differently because "inverting" then does not fit anymore)

I am not sure what the best answers would be here

Kneemund:
Alright, I think I've taken care of most of the things we talked about.
This is what's left to consider/discuss:

- Instead of adding the `optimize_for_printing` method to snapshots, we could put the logic directly into the `take_snapshot` method and control it using a boolean argument.

- Transparent images might cause issues when optimizing for printing, as described above. I'm not sure if we want to try to handle those cases, and if keeping this as a switch in the export dialogs is the best option.

- The icon for inverting the stroke colors of the selection is quite large, we might want to edit the SVG in order to make it smaller (or use a different one). I got the current one from icon library.

- We have to decide on using either HWB or HSL (or something else entirely). Both are implemented in `Color` ATM. I don't have a strong opinion on that, but HWB feels a bit more predictable to me (especially when "optimizing for printing").

flxzt:
> Instead of adding the optimize_for_printing method to snapshots, we could put the logic directly into the take_snapshot method and control it using a boolean argument.

I don't have strong opinion about that, we can keep it like this since  `take_snapshot()` is used not only for export and the argument would be a bit out of place in those places.

> Transparent images might cause issues when optimizing for printing, as described above. I'm not sure if we want to try to handle those cases, and if keeping this as a switch in the export dialogs is the best option.

It would be really difficult to handle this corner-case, wouldn't it?

> The icon for inverting the stroke colors of the selection is quite large, we might want to edit the SVG in order to make it smaller (or use a different one). I got the current one from icon library.

Yeah, most standard icons taken from icon library are scaled down 2x in the toolbars because the toolbar button styling scales the icons to 32x32 instead of the usual 16x16 pt's.

> We have to decide on using either HWB or HSL (or something else entirely). Both are implemented in Color ATM. I don't have a strong opinion on that, but HWB feels a bit more predictable to me (especially when "optimizing for printing").

Sounds good to me/I'll leave it to you. Let's go with the version that worked best in your tests

> Should this be enabled when printing directly from within the app? There are currently no settings to change that, but looking at PrintOperation, it seems fairly simple to add custom widgets to the print dialog, I could add that in another PR.

Yeah that would be nice. Either here or in a follow-up, but since the option is explicitly mentioning printing, users probably expect it to be present there as well when they check the new feature out. Also: we would get a preview "for free".

> Export previews and including document options in the history would make this feel a lot more polished. Now that inverting the document background is a button in the settings, the lack of the latter isn't as noticeable anymore. Do you consider those things to be blockers?

I would say: including the document in the history is not critical, but the preview would be nice. Unless you're saying you want this feature out as soon as possible/you're not planning to do much work on it in the near future. I could also try to implement it, because I'd like to have this in other places as well - using [TabOverview](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.TabOverview.html) with tab previews could be really nice and previews are generally useful when exporting document pages in my opinion - then an option to only export a range of pages could also be added.

I do have some ideas how this could be implemented in a general way, but I am limited on available time at the moment unfortunately.

--

Back to the colors:

The only other thing I noticed in the optimize-printing option is: do we really want to always choose the darkest color of the pattern? Personally I would have checked if the background color is "close" to white in brightness and only if not, invert the pattern color? Would this make more sense?

Kneemund:
> It would be really difficult to handle this corner-case, wouldn't it?

Yes, it would be. I'd be fine with leaving it as-is.

> Yeah that would be nice. Either here or in a follow-up, but since the option is explicitly mentioning printing, users probably expect it to be present there as well when they check the new feature out. Also: we would get a preview "for free".

I've just tried adding custom widgets, but that's sadly not supported by the GTK desktop portal:
`GtkPrintOperation::create-custom-widget not supported with portal`

Alternatively, we could add a dialog that comes before the system print dialog for configuration. That really isn't ideal though.

> I would say: including the document in the history is not critical, but the preview would be nice.

I agree, but I'm not sure if I'd be able to come up with a good general implementation.

> do we really want to always choose the darkest color of the pattern? Personally I would have checked if the background color is "close" to white in brightness and only if not, invert the pattern color? Would this make more sense?

That's how I implemented it at first, but doing it this way prevents the "unpredictable" behavior of HWB, while keeping the advantages over HSL. Personally, I always want the colors to be as dark as possible when printing - otherwise, they could become unreadable.
I'm not sure if this introduces unintended side-effects, but I haven't stumbled across any during my testing. To be fair, this might not be very intuitive.

flxzt:
> I've just tried adding custom widgets, but that's sadly not supported by the GTK desktop portal:
GtkPrintOperation::create-custom-widget not supported with portal

Okay, then let's not waste too much time on it, we can look at it again when Gtk's new print API will be available.

> Personally, I always want the colors to be as dark as possible when printing - otherwise, they could become unreadable.

I think that most choose the pattern color in a similar brightness and hue to the background color so that it doesn't become too dominant/blends with the background

---

I implemented a preview feature in #670 . They way I envisioned it to work with the optimize-printing feature is that this option will be applied non-destructively in [StrokeContent::gen_svg()](https://github.com/flxzt/rnote/blob/779849480bec17b3dde9684e3ede344d46f15b41/rnote-engine/src/engine/strokecontent.rs#L68) and [StrokeContent::draw_to_cairo()](https://github.com/flxzt/rnote/blob/779849480bec17b3dde9684e3ede344d46f15b41/rnote-engine/src/engine/strokecontent.rs#L104). The neat thing about this is that these are the only functions used for export and drawing the preview for all exports, so they should always match. The option can then be stored inside the paintable, similar to `draw_background`, `draw_pattern`, `margin`.

Performance could be an issue when inverting the colors on the fly in the preview, so we have to implement some caching. Would need some testing.

Kneemund:
> I think that most choose the pattern color in a similar brightness and hue to the background color so that it doesn't become too dominant/blends with the background

That's true, but let's say the document has a dark background and uses the red from the default color palette for some strokes. When using the option, all colors would be inverted, making the red quite a bit brighter and less readable on a white background compared the original document IMO.

It doesn't become completely *unreadable* though, and I'd be fine with your suggestion too.

---

Nice. It seems quite straightforward to integrate this :)

> Performance could be an issue when inverting the colors on the fly in the preview, so we may have to implement repaint caching. Would need testing.

Yeah, we'll see.

Kneemund:
This should work in terms of functionality now. The code needs some work though.

flxzt:
> invert all stroke colors if the background fits a certain brightness/luma threshold, or always use the darkest color.

Sorry, I probably didn't express myself clearly: when I brought up the colors, I was only referring to the pattern color. So the question for me was if it were better if the pattern color should behave like this: If the existing background color is bright color (before setting it to white), don't modify the pattern color, but the background color is not a bright color, invert the pattern color, so that the contrast between pattern and background stays roughly the same.

Always choosing the darkest color for the content/strokes sounds good to me. (Like it is implemented right now)

I don't know what to do about the printing dialog. I don't know if the [new print api](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/5484) will land in the next Gtk release cycle and even when it lands it is not clear to me if it initially will be possible to add custom options with it. So we probably shouldn't wait for it. Considering that and because custom options with the existing API does not work either, I would reconsider and would be fine if the options stay unexposed for now. We could point users to the pdf export when it comes up.

Kneemund:
> If the existing background color is a bright color (before setting it to white), don't modify the pattern color, but if the background color is not a bright color, invert the pattern color, so that the contrast between pattern and background stays roughly the same.

Yeah, that's better than the current approach.

> Considering that and because custom options with the existing API does not work either, I would reconsider and would be fine if the options stay unexposed for now. We could point users to the pdf export when it comes up.

Alright, should "optimize for printing" be disabled or enabled for now? This is specifically designed for printing, but it can also lead to very unexpected results.

flxzt:
If it can't be toggled, it would probably be better to disable it in the print action. Or expose a toggle **somewhere**, but I don't know where that could be. The only other option I can imagine is a separate dialog before the actual print dialog, but that has it's own issues like not being able to go back to it from the print dialog.

No of the options are ideal, but only exposing it in the export dialogs seems to be the best solution for now in my opinion

Kneemund:
Okay, done :)

Should I create separate methods in the `Stroke` struct for doing the following two operations?

https://github.com/flxzt/rnote/blob/416351f52c6d1fb69fa209baa0aeb352d311dd7f/crates/rnote-engine/src/store/stroke_comp.rs#L308-L346
https://github.com/flxzt/rnote/blob/416351f52c6d1fb69fa209baa0aeb352d311dd7f/crates/rnote-engine/src/engine/strokecontent.rs#L160-L204

They are currently implemented directly in the functions that use them, but splitting this might make things more readable. But it would make it less clean to only mark the strokes as dirty which are actually modified when inverting colors. We'd either have to mark the entire selection as dirty, or make the `Stroke` method return a boolean value storing if the stroke was modified or not.

flxzt:
> make the Stroke method return a boolean value storing if the stroke was modified or not.

this sounds good to me  :+1:

Kneemund:
I think this is ready for review now. `palette` is technically outdated (0.7.2 => 0.7.3), but as far as I can tell the other dependencies we use still depend on the "old" version too, and the [change log](https://github.com/Ogeon/palette/blob/master/CHANGELOG.md#version-073---2023-08-10) doesn't seem to affect the stuff used in this PR.

Kneemund:
Hm, it compiled fine on my machine. I'll look into that later.

Kneemund:
Turns out I just forgot about the CLI.
Because both 'o' (open) and 'p' (no pattern) are already in use, I've decided not to set a short name for the `--optimize-printing` option. 

Also, the split order was mapped to incorrect strings, which caused exports to fail. I've fixed that while I was at it.

Although I can get past the CLI now, the actual export fails for some reason:
```
Gsk:ERROR:../gtk/gsk/gskrendernode.c:277:gsk_render_node_alloc: assertion failed: (gsk_render_node_types[node_type] != G_TYPE_INVALID)
Bail out! Gsk:ERROR:../gtk/gsk/gskrendernode.c:277:gsk_render_node_alloc: assertion failed: (gsk_render_node_types[node_type] != G_TYPE_INVALID)
```

I'm not sure if this is due to the file I'm trying to export, my environment or even related to this PR at all. I'll try to investigate that tomorrow.

On another note, the order seems to be mixed up here:

grafik

From what I can tell, the usage is `rnote-cli export <COMMAND> [OPTIONS] [RNOTE_FILES]...` and the message should reflect that. I haven't found a way to change it within the code, I'll probably have to take a look at the documentation of the crate.

flxzt:
the code looks very good to me. If you don't have anything else, it can be merged (squash).

> I'm not sure if this is due to the file I'm trying to export, my environment or even related to this PR at all. I'll try to investigate that tomorrow.

Oh, I am getting something similar: `GLib-GObject-ERROR **: 18:28:06.626: cannot create new instance of invalid (non-instantiatable) type '<invalid>'

I think I know what's going on, because of some refactor the rendering part of the engine (which uses `TextureNode`s/gtk4) is now invoked as well, which causes the error. Making `gtk4` a dependency for the cli and  calling [gtk4::init()](https://gtk-rs.org/gtk4-rs/stable/0.7/docs/gtk4/fn.init.html) before everything else in `main` fixes it.

Kneemund:
Okay, this should be good to go :)

> I think I know what's going on, because of some refactor the rendering part of the engine (which uses TextureNodes/gtk4) is now invoked as well, which causes the error.

Ah, that'd make sense.

> Regarding the order of the cli arguments. I think it's because some of the arguments are on the same level as the subcommands and are propagated down through claps [global](https://docs.rs/clap/latest/clap/struct.Arg.html#method.global). Not sure how/if it can be changed

I've found `display_order` in the docs but that seems to affect the order in which the individual subcommands/options are listed, not the order of the "usage" part.

flxzt:
The arguments probably need to be moved into the subcommands (and duplicated) to have a better output in the usage text. It's code duplication, but I can't see another way

PR #672, opened 2023-05-13
Pen events blocking improvements

### LINKED ISSUE #534: Writing zip-zap when wrting with tablet
**Describe the bug**

As you can see in the following picture.The stroke experience very large zip-zap in normal writing

image

image

**To Reproduce**

I didn't know how to back to the normal state, so I don't know how to reproduce. Therefore I seek for help in order to diagnose the problem

**Expected behavior**

** Console Output **
Run the app in the terminal with `flatpak --env=RUST_LOG=rnote=debug --env=RUST_BACKTRACE=1 run com.github.flxzt.rnote` and post the output

This is the debug log right after I spot an occurrence of the zip-zap:

[rnote_debug.log](https://github.com/flxzt/rnote/files/10847656/rnote_debug.log)

**Screenshots**

 - OS: Archlinux
 - Installation source: AUR
 - Desktop Environment: KDE 5.27.1 on wayland

**Additional context**

During the trouble shooting, I have seen the followings:

1. I won't experience this problem when drawing using mouse
2. Changing different pressure curves doesn't help
3. No problem with excalidraw and xournalpp drawing with the same tablet, which indicates that is a rnote only problem
4. I tried to disable both touchpad and mouse. The tablet is the only cursor manipulator, and the problem still exists.

Following are additional screenshots.

image

image

### LINKED ISSUE #624: Zoom and positioning using touch
In the last version, I could move the canvas using 2 fingers in touch. In this update, there are some weird lines while using 2 fingers (attaching a screenshot). Can there be an option for zooming and moving with 2 finger touch, just like many other drawing apps?

image

### LINKED ISSUE #665: Pointer events from other devices cause pen to jump
**Describe the bug**  
Pointer events from other devices cause the pen to jump to the other device's pointer position, usually making an annoying mark on the page.

**To Reproduce**
Steps to reproduce the behaviour:
1. Open Rnote in Wayland with a pen device (e.g. Wacom drawing tablet) and a mouse.
2. Start drawing using the drawing tablet (or other pen device) and then move the mouse around.
3. Observe the pen jumping to the mouse position and making unwanted marks.

**Expected behavior**  
Ideally the cursor might still visually jump, but not draw to a different pointer device's position.

**Console Output**  
(with the (log::debug command in canvas/src/input.rs)[https://github.com/flxzt/rnote/blob/ff4d1b73309efe34f83c6f38f25be087332f81cc/rnote-ui/src/canvas/input.rs#L167], also with pointer position (x, y) and length and index of elements vector):

```
 2023-05-10T15:16:57.614Z DEBUG rnote::canvas::input > (518.0, 172.3) handle event, state: Proximity, event_time_d: 0ns, modifier_keys: [], pen_mode: Some(Pen), length: 2 [1]
 2023-05-10T15:16:57.628Z DEBUG rnote::canvas::input > (518.1, 171.3) handle event, state: Proximity, event_time_d: 15ms, modifier_keys: [], pen_mode: Some(Pen), length: 2 [0]
 2023-05-10T15:16:57.628Z DEBUG rnote::canvas::input > (518.3, 170.4) handle event, state: Proximity, event_time_d: 0ns, modifier_keys: [], pen_mode: Some(Pen), length: 2 [1]
 2023-05-10T15:16:57.644Z DEBUG rnote::canvas::input > (518.8, 168.8) handle event, state: Proximity, event_time_d: 0ns, modifier_keys: [], pen_mode: Some(Pen), length: 1 [0]
 2023-05-10T15:16:57.644Z DEBUG rnote::canvas::input > (518.8, 168.8) handle event, state: Down, event_time_d: 0ns, modifier_keys: [], pen_mode: Some(Pen), length: 1 [0]
(rnote:12): Gdk-DEBUG: 16:16:57.644: proxy 0x5555572bc320 already has listener

(rnote:12): Gdk-DEBUG: 16:16:57.644: proxy 0x5555572bc320 already has listener

 2023-05-10T15:16:57.670Z DEBUG rnote::canvas::input > (367.8, 100.1) handle event, state: Down, event_time_d: 0ns, modifier_keys: [], pen_mode: None, length: 1 [0]
 2023-05-10T15:16:57.674Z DEBUG rnote::canvas::input > (518.8, 169.0) handle event, state: Down, event_time_d: 31ms, modifier_keys: [], pen_mode: Some(Pen), length: 2 [0]
 2023-05-10T15:16:57.674Z DEBUG rnote::canvas::input > (518.9, 169.4) handle event, state: Down, event_time_d: 0ns, modifier_keys: [], pen_mode: Some(Pen), length: 2 [1]
 2023-05-10T15:16:57.689Z DEBUG rnote::canvas::input > (519.0, 170.3) handle event, state: Down, event_time_d: 14ms, modifier_keys: [], pen_mode: Some(Pen), length: 2 [0]
```
(note the line with `pen_mode: None` and a very different pointer position).

**Screenshots**  
image

 - OS: Fedora Kinoite 38
 - App Version: v0.6.0-devel
 - Installation Source: Built with GNOME Builder flatpak
 - Desktop Environment: KDE Plasma 5.27.4

**Additional context**  
Why is this a problem?
- Most devices and desktop environments only send out input events when the device moves, and so this issue is not really a problem.
- However, in my case the combination of my laptop's touchpad and KDE Plasma results in events being sent for the touchpad pointer even when it is unused. I imagine this is possible (but rare) with other devices. This causes a lot of frustration when using Rnote with a pen input device.

cyclane:
**Fixes #665 , possibly #534 :**
~`PenHolder` stores pen device on first event and only handles events from that device until the pen is reinstalled.~

Pen device source type is stored on first event and events from other device source types are blocked and inhibited from propagating until the pen is up again.

**#624 :** ([not fixed](https://github.com/flxzt/rnote/pull/672#issuecomment-1565593980))
~The zooming gesture overrides touch drawing and undoes the previous stroke if pen was not idle when the gesture began (and also from the same device source type, otherwise the stroke is ended).~

**Why both implementations in the same PR?**
Both implementations depend on comparing [gdk::Device](https://gtk-rs.org/gtk4-rs/stable/0.6/docs/gdk4/struct.Device.html)s and since I think that is the main concern for both I just did everything in one PR.

**To consider before merging :**
- [x] Currently devices are compared with the [gdk::Device::eq](https://gtk-rs.org/gtk4-rs/stable/0.6/docs/gdk4/struct.Device.html#method.eq) (ie. `==`) which compares pointers [under the hood](https://docs.rs/glib/latest/src/glib/object.rs.html#656-658). It is not clear to me whether this is robust enough or not, but I have not encountered any issues with this in my testing so far.
  - If after-all this is not an option, I believe the only other unique [gdk::Device](https://gtk-rs.org/gtk4-rs/stable/0.6/docs/gdk4/struct.Device.html) property is [gdk::DeviceExt.name()](https://gtk-rs.org/gtk4-rs/stable/0.6/docs/gdk4/prelude/trait.DeviceExt.html#tymethod.name) which could be compared instead.
  - **Solution:** Compare device source types instead of devices.
- [x] Currently as soon as the zooming gesture ends there might still be one finger on the screen which can result in an unintended stroke if not blocked. Unfortunately a new `TouchBegin` event is sent for this finger (even though it is already on the screen), so blocking all touch events until the next `TouchBegin` event is not an option.
  - In my opinion this would probably be quite irritating, but I can't think of a better way to handle this other than blocking touch events for a short period after the gesture.
  - However, in my testing (and I think this should be the general case since <10Hz touchscreens are very unlikely to be used for writing), a short delay of 100ms fixes this issue quite well (commit [3ff0663](https://github.com/flxzt/rnote/pull/672/commits/3ff0663f16fc89d7f373df7f1e6280629dc532db)).
  - **Solution:** Implementation in the point above is sufficient.
- I agree with [#624 comment](https://github.com/flxzt/rnote/issues/624#issuecomment-1529072695) in that this shouldn't be an option and should just be implemented since the alternative behaviour is extremely unlikely to be the intended behaviour by the user, and additionally the inputs are coming from the same device so it is difficult to tell the difference between two fingers.
  - Also note the "Block Pinch to zoom" option pretty much does this anyway (as the comment points out).
- [ ] [Just undoing the current stroke on zoom isn't great](https://github.com/flxzt/rnote/pull/672#issuecomment-1556151371). Ideally the removed stroke should not be in the stroke history at all. Same applies for erasing a stroke.

**Note:** I have been (and am currently) using these for notes to find any bugs with my implementations. So far I have applied fixes to any problems I have encountered, and more importantly I have not encountered any issues related to how [gdk::Device](https://gtk-rs.org/gtk4-rs/stable/0.6/docs/gdk4/struct.Device.html)s are compared which I think is the primary concern for this PR.


flxzt:
Thanks for doing this work! 

I think there are two things that need discussion:

1. identifying unique input devices: I am a bit sceptical about pointer-comparing here. While it might work in most cases, I believe it might just be enough to check the same-ness of the [input-source](https://gtk-rs.org/gtk4-rs/stable/0.6/docs/gdk4/prelude/trait.DeviceExt.html#tymethod.source) ? I also don't see another way to uniquely identify devices, Gtk's devs also told me that wayland simply does not have this capability. Checking VID/HID wouldn't work well either, as some input devices are connected through the same usb device (Logitech unify receivers) or not through usb at all (PS/2, touchpads through smbus).
2. This seems "frontend" specific, and I am not sure if we should pass gdk::Device's into the engine at all. I am planning to remove the gtk dependency to the engine entirely. Maybe we can make this work only in `rnote-ui` ?

cyclane:
Thank you for your comments.
> 1. identifying unique input devices: I am a bit sceptical about pointer-comparing here. While it might work in most cases, I believe it might just be enough to check the same-ness of the [input-source](https://gtk-rs.org/gtk4-rs/stable/0.6/docs/gdk4/prelude/trait.DeviceExt.html#tymethod.source) ? I also don't see another way to uniquely identify devices, Gtk's devs also told me that wayland simply does not have this capability. Checking VID/HID wouldn't work well either, as some input devices are connected through the same usb device (Logitech unify receivers) or not through usb at all (touchpads through smbus).

Alright. Though I have had no issues, there definitely could be issues under other conditions (or in future versions of GTK e.t.c.) and for most use-cases I think it would be enough to compare [input-source](https://gtk-rs.org/gtk4-rs/stable/0.6/docs/gdk4/prelude/trait.DeviceExt.html#tymethod.source)s instead since most setups will only have at most one of each type of input _if it worked as expected_. Unfortunately, on Wayland pens (I have tested both touchscreen stylus and Wacom drawing tablet) are `InputSource::Mouse`, so issue #665 and #534 would remain unresolved.
  - However, [gdk::Device::name](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gdk4/prelude/trait.DeviceExt.html#tymethod.name) is also available and does actually work as expected, so I would suggest this instead of [input-source](https://gtk-rs.org/gtk4-rs/stable/0.6/docs/gdk4/prelude/trait.DeviceExt.html#tymethod.source). I think it is possible for the name to change, but since these comparisons are very short-term I do not believe that would cause problems.

> 2. This seems "frontend" specific, and I am not sure if we should pass gdk::Device's into the engine at all. I am planning to remove the gtk dependency to the engine entirely. Maybe we can make this work only in `rnote-ui` ?

That makes sense. I wasn't aware of this plan so I just implemented it in the easiest way (to at least show that it works). Once we figure out point 1, I'll push an implementation that that works only with `rnote-ui`.

> 3. Regarding blocking input after a pinch-zoom has finished. I don't see a better way to do this either, so I think it is totally fine

Okay, I'll keep this implementation then (with 100ms - I think that's a sensible delay to use, not too long but should cover the vast majority of devices).

flxzt:
> Alright. Though I have had no issues, there definitely could be issues under other conditions (or in future versions of GTK e.t.c.) and for most use-cases I think it would be enough to compare [input-source](https://gtk-rs.org/gtk4-rs/stable/0.6/docs/gdk4/prelude/trait.DeviceExt.html#tymethod.source)s instead since most setups will only have at most one of each type of input if it worked as expected. Unfortunately, on Wayland pens (I have tested both touchscreen stylus and Wacom drawing tablet) are InputSource::Mouse, so issue https://github.com/flxzt/rnote/issues/665 and https://github.com/flxzt/rnote/issues/534 would remain unresolved.

That's the reason I don't want to rely on it - we would rely on the internal structure of how gtk handles device instances, which could break at any moment.
Regarding using input source, I now remember that I even reported this myself in [gtk's bug tracker](https://gitlab.gnome.org/GNOME/gtk/-/issues/4374). Not a good approach then.
But using the name is also not great - on X11 this is always "Virtual Core Pointer" (for stylus, touchpad, touchscreen and mouse). On Wayland this works better, at least the stylus has a different name.
In any case, we should be careful and lean towards a solution that may end up not detecting different devices in some cases.

> That makes sense. I wasn't aware of this plan so I just implemented it in the easiest way (to at least show that it works). I have pushed an implementation that works almost only in rnote-ui (it needs to be able to know whether engine.pen_progress is PenProgress::Idle

I think we can simply use `current_pen_progress()` directly, it already is `pub`.

cyclane:
> But using the name is also not great - on X11 this is always "Virtual Core Pointer" (for stylus, touchpad, touchscreen and mouse). On Wayland this works better, at least the stylus has a different name.

As far as I am aware on X11 everything uses the same pointer, so #534 and #665 are impossible anyway since any updates from devices will move the used pointer by a minimal/zero amount.
This can be demonstrated by trying to draw with a pen and mouse intentionally (with current Rnote, not this PR):
- On Wayland you get the expected pointer jumping from #534 and #665.
- On X11 you just make the line squiggly by moving the used pointer relative to its current position.

And also since X11 seems to send the same device (Virtual Core Pointer) with events from all devices anyway, I think there's genuinely no way to change this by blocking events.

> In any case, we should be careful and lean towards a solution that may end up not detecting different devices in some cases.

I agree - ideally this would be using `input-source` which we have ruled out. I think using names gives almost the same effect as comparing devices and is fine to do since it is extremely unlikely (if not impossible) for the name to unintentionally change during a stroke. Additionally, it is less likely to break in the future in the way the pointer comparison could.

Another way would be to compare the [gdk::Event.device_tool()](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gdk4/struct.Event.html#method.device_tool), but this would only fix #665 and #534 for pens (and not touch drawing) and can also result in unintended stroke deletion with the touch-drawing zooming implementation.

> I think we can simply use `current_pen_progress()` directly, it already is `pub`.

Oh right, I didn't notice `RnoteEngine.penholder` was `pub`.

flxzt:
> On X11 you just make the line squiggly by moving the used pointer relative to its current position.

Right, the can't be a second cursor at a different position, but that wouldn't necessarily mean that the occasional jump by a faulty touchscreen/touchpad couldn't happen, would it? To me it sounds like still would be an improvement if other input is blocked for the duration of an in-progress stroke. I am not sure I understand your argument.

> That would work as well, but I think there is no point if we can find a good way to compare devices. I think comparing names would be fine, but custom input source detection is safer (in the sense we know what is going on) than comparing names.

Yeah but I wouldn't consider comparing the names a good way to detect the different devices. At least on X11 the case definitely exists where the same name  would be reported, but our custom detection would recognize a difference in input source.
 I think going with our own detection would be more robust and mimic gtk's detection in it's gestures, which has already been proven to work relatively well.

cyclane:
> Right, there can't be a second cursor at a different position, but that wouldn't necessarily mean that the occasional jump by a faulty touchscreen/touchpad couldn't happen, would it? To me it sounds like it still would be an improvement if other input is blocked for the duration of an in-progress stroke. I am not sure I understand your argument.

Since touchpads are relative it solves a faulty touchpad I think, but you're right - it doesn't solve a faulty touchscreen.

> Yeah but I wouldn't consider comparing the names a good way to detect the different devices. At least on X11 the case definitely exists where the same name would be reported, but our custom detection would recognize a difference in input source.
I think going with our own detection would be more robust and mimic gtk's detection in it's gestures, which has already been proven to work relatively well.

Okay, I see your point now - I didn't realise on X11 touch events still have their touch event types, even though the device is the same (Virtual Core Pointer). I'd like to note though the stylus detection does not work on X11 (due to everything being the same device I'm guessing).

I have pushed an implementation with manual input source detection. I still used the `gdk::InputSource` enum so that if GTK does fix their input source Rnote can easily switch to that.

flxzt:
> I still used the [gdk::InputSource](https://gtk-rs.org/gtk4-rs/stable/latest/docs/gdk4/enum.InputSource.html) enum so that if GTK does fix their input source Rnote can easily switch to that.

Sounds good. Maybe a more detailed comment explaining this above `input_source_from_event()` would be nice for our future selves.

Okay, the input blocking part looks good, I see that you used `PenState`, which is probably better when thinking about it - using `PenProgress` could result in a "locked" state where it might be unclear to the user how to finish an in-progress stroke.

Regarding the zooming while having the touch-drawing option enabled - unfortunately this does not work well on X11. Touch events still are drawn as strokes on the canvas when a zoom-pinch starts, and also sometimes when it ends.  

Also - undoing is not ideal here, I don't think the user would ever want to have this stroke in the history. We should probably remove those strokes entirely out of the `StrokeStore`. That would mean we need to account for that in a different part of the `StrokeStore` [here](https://github.com/flxzt/rnote/blob/main/rnote-engine/src/store/mod.rs#L191). I think we would need to explicitely move the rendering components from the history into the current engine.

cyclane:
> Regarding the zooming while having the touch-drawing option enabled - unfortunately this does not work well on X11. Touch events still are drawn as strokes on the canvas when a zoom-pinch starts, and also sometimes when it ends.

I forgot to test X11 after changing the implementation to use input sources. It is fixed now.

> Also - undoing is not ideal here, I don't think the user would ever want to have this stroke in the history. We should probably remove those strokes entirely out of the `StrokeStore`. That would mean we need to account for that in a different part of the `StrokeStore` [here](https://github.com/flxzt/rnote/blob/main/rnote-engine/src/store/mod.rs#L191). I think we would need to explicitely move the rendering components from the history into the current engine.

Yes, I'm not sure why I overlooked that. I don't quite understand why anything needs to be accounted for in `StrokeStore.import_history_entry()`, but also I don't think it should be handled only for strokes since otherwise other (not Brush) pens could still result in unintended selections / erases e.t.c. . I would suggest maybe adding a `cancel_action` (or similar) flag to `PenEvent::Cancel` and handling cancelling the current action for each pen?

flxzt:
Regarding the zooming:

I made some changes that are improving some of the mentioned issues:  
- 510d8a0136c0d261cb68f5730dad5acd02116673 should allow removing strokes entirely out of the engine without breaking the rendering
- 689a49ea2d722dcae0158fa6e20b275742569fe2 refactored the way the history is handled - the current state is now recorded into the history *after* a change has happened. This could enable removing an already drawn stroke by passing something like `cancel_action` with a `PenEvent::Cancel` similar to how you described it, but without having to undo & remove future (which in my opinion is a hack and shouldn't be done if we can avoid it)

*But* I am starting to doubt that this is the best way to implement this. There will always be some unsolvable jankiness to it, like moving the typewriter textbox and deselecting. And I also don't like that the to be deleted strokes are visible for a short time, it looks like a glitch.

Going forward maybe we can approach this differently, and implement some alternative controls when touch-drawing is enabled - like easier to access zoom controls (for example as a separate overlay, like in previous versions) and a better way to drag the view around (maybe like a software joystick or something like that?)

What do you think? I definitely want the input source part of this PR to land, so maybe would be best to split and do another PR/issue/discussion for the zooming, which in my opinion still requires more thought and work

cyclane:
> I made some changes that are improving some of the mentioned issues:
> 
> * [510d8a0](https://github.com/flxzt/rnote/commit/510d8a0136c0d261cb68f5730dad5acd02116673) should allow removing strokes entirely out of the engine without breaking the rendering
> 
> * [689a49e](https://github.com/flxzt/rnote/commit/689a49ea2d722dcae0158fa6e20b275742569fe2) refactored the way the history is handled - the current state is now recorded into the history _after_ a change has happened. This could enable removing an already drawn stroke by passing something like `cancel_action` with a `PenEvent::Cancel` similar to how you described it, but without having to undo & remove future (which in my opinion is a hack and shouldn't be done if we can avoid it)

Okay thank you, so we can make an implementation now that will not affect history at all on zooming (if we do decide to stick with this...).

> _But_ I am starting to doubt that this is the best way to implement this. There will always be some unsolvable jankiness to it, like moving the typewriter textbox and deselecting. And I also don't like that the to be deleted strokes are visible for a short time, it looks like a glitch.
> 
> Going forward maybe we can approach this differently, and implement some alternative controls when touch-drawing is enabled - like easier to access zoom controls (for example as a separate overlay, like in previous versions) and a better way to drag the view around (maybe like a software joystick or something like that?)

I implemented it this was because of [comment in #624](https://github.com/flxzt/rnote/issues/624#issuecomment-1518883200) which I thought would be the easiest to use implementation. However, I don't understand why we can't solve the typewriter textbox and deselecting if we handle each brush manually through `PenEvent::Cancel` (can't we just reset the position / selection state when zooming starts?)

I see your point about the line before the zooming, which we obviously can't remove. It looks a little weird, but doesn't actually affect anything after-all (and doesn't even appear if you manage to get both fingers on the screen at once). I think alternative controls could also work, but I don't see how we could make those as fast & easy to use as the normal zooming gesture (which can also be used for panning) which I think is a bigger issue than a temporary stroke.

> What do you think? I definitely want the input source part of this PR to land, so maybe would be best to split and do another PR/issue/discussion for the zooming, which in my opinion still requires more thought and work

Yeah I agree, I've removed the zooming implementations from this PR so it can be merged. I'm thinking let's continue the discussion back in #624 until this gets merged, at which point I can create new PR for zooming & panning with touch drawing?

rnote-ui/src/canvas/input.rs
```diff
@@ -268,19 +285,21 @@ fn debug_gdk_event(event: &gdk::Event) {
...
 
 /// Returns true if input should be rejected
-fn reject_pointer_input(event: &gdk::Event, touch_drawing: bool) -> bool {
+fn reject_pointer_input(
+    event: &gdk::Event,
+    state: PenState,
+    input_source_matches: bool,
+    touch_drawing: bool,
+) -> bool {
+    if matches!(state, PenState::Down) && !input_source_matches {
```
flxzt, line 294: a comment explaining what this is solving would be helpful here

rnote-ui/src/canvas/input.rs
```diff
@@ -210,6 +221,12 @@ pub(crate) fn handle_pointer_controller_event(
         }
     }
 
+    if matches!(state, PenState::Down) {
```
flxzt, line 224: do we need check for the pen state here? Unless I am missing something the input would be rejected correctly regardless and this behaviour might be confusing if `pen_input_source()` is reused elsewhere in the future.

flxzt:
> Okay thank you. I think this means the zooming gesture with touch drawing can be implemented without any stroke-history issues now.

Yes, but it would still need an additional flag in the `Cancel` events to tell the pens that the current state should not be recorded and any strokes/state that was committed during the progress of the pen should be reverted.

> why we can't solve the typewriter textbox and deselecting if we handle each brush separately through PenEvent::Cancel (can't we just reset the position / selection state when zooming starts?)

Yeah we could do that, but it would require more work. For example: as soon as we are in the selecting state of the selector, we don't have the last selected strokes anymore that would be needed to revert the unselect. Another example: We would need to save the trashed strokes in the eraser, or when it is in the "split-stroke" mode save the state of the modified strokes and revert that.

And implementing it in this way still would have the issue of visibly doing & reverting something on the canvas..

But yeah, we can continue this discussion in #624 . I'd love to get zooming/better navigation working in combination with touch-drawing, so let's discuss more possible solutions there

rnote-ui/src/canvas/input.rs
cyclane: Thanks for pointing that out, you're right - this is unnecessary.

flxzt:
Let's leave #534 open until it is confirmed fixed.

Testing this is currently difficult because I gnome-shell is constantly crashing when using the stylus (see [this bug](https://gitlab.gnome.org/GNOME/mutter/-/issues/2796) ) but here is what I observed:

|first input method|second input method| behaviour|
|---|---|---|
|touch|stylus|the first stroke gets finished, and a second is created for the stylus input. This seems to be because gtk emits a touch-end or touch-cancel event. After the stylus is finished, another stroke is generated when continuing to draw with touch|
|touch|mouse| if the mouse is only moved the events seem to be correctly rejected. But if the mouse clicked down and drags, The view gets moved|
|mouse|stylus|the strokes jumps to and back from the stylus, without being finished. Shouldn't the stylus input be rejected here?|
|mouse|touch| This seems a bit problematic: The touch input seems to be rejected sometimes, but this then causes the event to be propagated to the drag-canvas gesture, which then moves the view around. And not every event seems to be rejected, so the stroke still jumps to the touch input|
|stylus|touch| after starting a stroke with the stylus, there don't seem to be any touch events coming in. I think this is because [palm rejection](https://wayland.freedesktop.org/libinput/doc/1.12.0/palm-detection.html) on my device works well. I don't know if they are correctly rejected for devices where palm rejection at the driver level doesn't work well.|
|stylus|mouse|The stroke zips around a bit, but continues. Some events seem to be rejected but not all.|

I believe we need to differentiate between "reject but let the event continue to propagate", and "reject and drop the event" to resolve some of these issues.

I also noticed that the pen state sometimes goes into an "invalid" state where a "pen up" seems to be missing so the stroke continues to be drawn even when the mouse button is released. I couldn't tell what consistently reproduces this.

So there are still some unresolved issues/even new bugs that get introduced by this. Since this is really critical to the core functionality of the app we need to be careful here

cyclane:
Thank you for the testing!

> touch | stylus | the first stroke gets finished, and a second is created for the  stylus input. This seems to be because gtk emits a touch-end or  touch-cancel event. After the stylus is finished, another stroke is  generated when continuing to draw with touch

Unfortunately I cannot reproduce this and so can't really test it...

> touch | mouse | if the mouse is only moved the events seem to be correctly rejected.  But if the mouse clicked down and drags, The view gets moved

I think I have fixed this with your suggestion of inhibiting events from other devices instead of just blocking them.

> mouse | stylus | the strokes jumps to and back from the stylus, without being finished. Shouldn't the stylus input be rejected here?

In a [previous message](https://github.com/flxzt/rnote/pull/672#issuecomment-1553750703) I mentioned that stylus detection does not work on X11 - everything is the same device (Virtual Core Pointer) and so no device tool exists, even if it is a stylus.

I think this is the cause of this issue?

This works properly (with the stylus input being rejected) on Wayland.

> mouse | touch | This seems a bit problematic: The touch input seems to be rejected  sometimes, but this then causes the event to be propagated to the  drag-canvas gesture, which then moves the view around. And not every  event seems to be rejected, so the stroke still jumps to the touch input

Testing on my device it seems when the mouse is down touchscreen events come in as mouse events, seems to be an X11 issue (I guess probably another extension of the fact everything uses the same pointer?).

> stylus | touch | after starting a stroke with the stylus, there don't seem to be any touch events coming in. I think this is because palm rejection  on my device works well. I don't know if they are correctly rejected  for devices where palm rejection at the driver level doesn't work well.

I believe we can simulate this by just using a finger (in my case I receive touchscreen events and they are rejected as intended).

> stylus | mouse | The stroke zips around a bit, but continues. Some events seem to be rejected but not all.

Similar to `mouse | stylus`, I believe this is due to the stylus appearing as a mouse on X11 (due to the lack of a `.device_tool()`). Again this works fine in Wayland for me.

-------

> I believe we need to differentiate between "reject but let the event continue to propagate", and "reject and drop the event" to resolve some of these issues.

Yes it solved the `touch | mouse` issue, and also the other combinations on Wayland (which would drag instead of having the issues you described).

> I also noticed that the pen state sometimes goes into an "invalid" state where a "pen up" seems to be missing so the stroke continues to be drawn even when the mouse button is released. I couldn't figure out how to consistently reproduce this.

I haven't experienced this at all... I'll try to do some X11 testing, maybe it's X11-specific again.

cyclane:
I don't have much time at the moment, so I'll take a deeper look into all of this at the end of next week, but for here are my thoughts:

- Testing on KDE Plasma 5.27 Wayland I cannot reproduce any of the Wayland issues you mention, BUT I can reproduce them on GNOME 44. So it seems to me the behaviour can also vary between compositors.
- I have managed to reproduce most/all of the X11 issues in GNOME (KDE X11 is crashing for some reason, so I couldn't test) and most of them seem to be from the stylus not having a device tool.

Once I get around to looking into this deeper next week I'll make a list of the cause of each issue and implement solutions if I come up with any. Seeing how wildly this varies between different environments, I will also test this on Windows to see what sort of issues come up there.

> I do think we are able to differentiate - the device tool should exist for the stylus, because thats also how a pressed stylus eraser button is recognized - and that works on X11.

I don't get a device tool on X11... If you do maybe it's yet another device-specific thing.

> Yeah detecting the correct source on X11 seems to be pretty difficult/impossible in the app. Adding rejection based on velocity could be another possible solution, but I can only speculate how robust and error-prone it would be.

Yeah, The velocity thing wouldn't even work if the two devices are pointing close to each other, so I don't think it's worth considering at all. I think X11 will probably have to remain with some issues in the end.

cyclane:
Apologies for the much longer than anticipated delay.

I've looked at X11/Wayland on GNOME/KDE, here's what I have gathered. Note all logs are collected at the beginning of `handle_pointer_controller_event` like so:

```rust
pub(crate) fn handle_pointer_controller_event(
    canvas: &RnCanvas,
    event: &gdk::Event,
    mut state: PenState,
) -> (Inhibit, PenState) {
    let now = Instant::now();
    let mut widget_flags = WidgetFlags::default();
    let touch_drawing = canvas.touch_drawing();
    let gdk_event_type = event.event_type();
    let gdk_modifiers = event.modifier_state();
    let _gdk_device = event.device().unwrap();
    let backlog_policy = canvas.engine_ref().penholder.backlog_policy;
    let input_source = input_source_from_event(event);
    let is_stylus = input_source == Some(gdk::InputSource::Pen);

    //std::thread::sleep(std::time::Duration::from_millis(100));
    //super::input::debug_gdk_event(event);

    let device_tool = event.device_tool();
    let pen_input_source = canvas.pen_input_source();

    let inhibit_o = reject_pointer_input(
        event,
        state,
        input_source == canvas.pen_input_source(),
        touch_drawing,
    );
    log::debug!("event={gdk_event_type:?} input_source={input_source:?} inhibit={inhibit_o:?} device={_gdk_device:?} device_tool={device_tool:?} | state={state:?} pen_input_source={pen_input_source:?}");

    if let Some(inhibit) = inhibit_o {
        return (inhibit, state);
    }
    ...
```

## X11
### KDE Plasma (kwin) 5.27.6
- mouse then touch (touch drawing enabled)
  - Touch input is inhibited correctly, but pointer still moves (because everything uses the same pointer in X11).
  - This causes the stroke to go the the touch position once the next mouse input comes in.
- mouse then touch (touch drawing disabled)
  - For some reason even if `pointer_controller` has `Capture` as its propagation phase, and `canvas_zoom_gesture` has `Bubble` or even `Target`, the zoom even is still received first and so is not inhibitted as expected.
- touch then mouse (touch drawing enabled)
  - Similar issue to _mouse then touch (touch drawing enabled)_, except the pointer moves relatively so much less noticeable.
- touch then mouse (touch drawing disabled)
  - Drawing starts at finger as expected
- stylus then touch (touch drawing enabled)
  - input ignored correctly (not the same as mouse then touch because the stylus has an absolute position, unlike the mouse)
- stylus then touch (touch drawing disabled)
  - Similar issue to _mouse then touch (touch drawing disabled)_.
- touch then stylus (touch drawing enabled)
  - input ignored correctly
- touch then stylus (touch drawing disabled)
  - Similar issue to _mouse then touch (touch drawing disabled)_.
- mouse then stylus
  - Similar issue to _mouse then touch (touch drawing enabled)_.
- stylus then mouse
  - Similar issue to _touch then mouse (touch drawing enabled)_.

### GNOME (mutter) 44
Exactly the same as X11 KDE (kwin) 5.27.6 .

## Wayland
### KDE Plasma (kwin) 5.27.6
- mouse then touch
  - input ignored correctly
- touch then mouse
  - input ignored correctly
- stylus then touch
  - input ignored correctly
- touch then stylus
  - input ignored correctly
- mouse then stylus
  - input ignored correctly
- stylus then mouse
  - input ignored correctly

### GNOME (mutter)
- mouse then touch
  - On touch all mouse events stop being sent, pausing the input while there is touch input.
  - If the mouse is let go, the ButtonRelease event is never received, meaning once touch is also let go Rnote thinks the pen is still down even though it is not.
  - This seems to be a GTK issue so I'm not sure we can fix this (considering the events are not being sent at all).
- touch then mouse
  - input ignored correctly
- stylus then touch
  - input ignored correctly
- touch then stylus
  - input ignored correctly
- mouse then stylus
  - input ignored correctly
- stylus then mouse
  - input ignored correctly

**Note:** I was unable to replicate the behaviour you found with GNOME Wayland...

## Summary
Wayland on KDE works perfectly, and on GNOME mostly perfectly apart from the _mouse then touch_, which I think is a GTK issue that can't be fixed in Rnote.

However, you reported many more issues on GNOME Wayland, so it seems to be very device-specific and I can't really look into that further myself.

As for X11, most of the issues arise from the nature of X11 having the same pointer for everything, and so cannot be fixed. _mouse then touch (touch drawing disabled)_ could be improved, but then it'll just have the same issue as _mouse then touch (touch drawing enabled)_.

I think I've done all I can with my device, so for the issues you had with GNOME Wayland it'd be great if you could investigate them on your device.

flxzt:
At one point we decided to drop support for X11 (see item of list: https://github.com/flxzt/rnote#pitfalls--known-issues) because it is increasingly difficult to test for it and X11 stylus input always had many issues too complex to solve for all cases.

So I think this MR is outdated and a new one should be created if it is still relevant for Wayland.

PR #686, opened 2023-05-25
refactor history to be able to record after a change

flxzt:
This PR is refactoring the history and `record()` calls to be made after a change on the document happens. This is more elegant and was a design flaw in the original implementation. Previously, there was the issue that possible changes had to be anticipated before they actually happened, which caused some redundant record calls in certain conditions.

This also improves the history when using the typewriter, state after typing characters that are non-whitespace is now appended to the latest history entry, and only text that contains whitespace results in new record calls.

Remaining issues that need to be fixed:

- ~~the selected state of strokes when undoing is inconsistent. It might be better to simply unselect after every undo/redo, or even remove the selection components from the store entirely so that this state only lives in the selector pen and is not part of the recorded history state.~~ fixed by removing selected state entirely from the history.

the work of this PR is needed for #672 so that in-progress strokes can be removed without them being recorded into the history

### DIFF
--- a/rnote-engine/src/store/mod.rs
+++ b/rnote-engine/src/store/mod.rs
@@ -191,8 +191,6 @@ impl StrokeStore {
     pub(crate) fn record(&mut self, _now: Instant) -> WidgetFlags {
         let mut widget_flags = WidgetFlags::default();
 
-        log::debug!("recording state to history");
-
         if self
             .history
             .back()
@@ -225,8 +223,6 @@ impl StrokeStore {
     pub(crate) fn update_latest_history_entry(&mut self, _now: Instant) -> WidgetFlags {
         let mut widget_flags = WidgetFlags::default();
 
-        log::debug!("update latest history entry with current state");
-
         if self
             .history
             .back()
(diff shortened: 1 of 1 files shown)


flxzt:
Oh thanks, I missed handling this. The cleanest way would probably be adding a `initial_history_state` argument in `store.clear_history()` and `store.clear()` and passing it's initial state in `store.import_from_snapshot()`.

But feel free to push, especially bug-fixes!

Kneemund:
Alright. Is there a situation where we don't want the initial state to be the current state though? Conceptionally, I think it would make the most sense to always set it to `store.create_history_entry()` instead of the default entry. That also doesn't change the behavior of `store.clear()`.

flxzt:
Yeah that's true. That would work as well, it's just a bit less explicit and one has to be a little bit more careful when using `clear_history()` - it is not immediately obvious for a caller that the order in which the history and the current state is cleared matters.

Kneemund:
Fair enough, I've implemented it like you suggested :)

flxzt:
okay :) thanks for noticing and fixing this, I don't think I would have noticed until next release

PR #702, opened 2023-06-09
feature: export page order

Kneemund:
Adds an option to export dialogs for changing the order in which the pages of infinite layouts are exported. It is disabled if the document layout is not (semi-)infinite or if the export format is not finite. The default is the current standard, i.e. "Horizontal First".

image

### Things to Discuss

- Name and description of this option.
- It might be better to create separate AABB helper functions for every `SplitOrder` instead of having a single one (`split_extended_origin_aligned`). That would be more readable but less DRY.
- `SplitOrder` is currently used directly within the export preferences structs and the values have different names than the options in the UI (`RowMajor` <-> "Horizontal First").

### DIFF
--- a/rnote-compose/src/helpers.rs
+++ b/rnote-compose/src/helpers.rs
@@ -14,8 +14,8 @@ use serde::{Deserialize, Serialize};
     num_derive::FromPrimitive,
     num_derive::ToPrimitive,
 )]
-#[serde(rename = "split_direction")]
-pub enum SplitDirection {
+#[serde(rename = "split_order")]
+pub enum SplitOrder {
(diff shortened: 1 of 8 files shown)


flxzt:
Nice.

- yeah the subtext could be a bit longer/more descriptive.
 - `SplitOrder` as an argument for `split_extended_origin_aligned` is fine with me
 - naming it differently in the UI is okay, naming it RowMajor would be too technical imho.

flxzt:
What about: "The page order when documents with layouts \n that expand in X and Y directions are cut into pages"

flxzt:
Looks good, thanks :)

PR #728, opened 2023-06-27
Shortcut mode for holding a button (like Xournal++)

TecCheck:
I recently discovered rnote and I really like it. However I need to have a pen mode like the one in Xournal++ where you hold the shortcut button to activate the tool and once you release it, it goes back to the previous tool.

I don't have much experience with rust so any feedback on code quality is very welcome. Also I'm not quite sure if the name "Hold" is good or not so input is welcome there too.

Otherwise I hope you like it :)

### DIFF
--- a/rnote-engine/src/engine/mod.rs
+++ b/rnote-engine/src/engine/mod.rs
@@ -698,6 +698,26 @@ impl RnoteEngine {
         )
     }
 
+    /// Handle a released shortcut key.
+    pub fn handle_released_shortcut_key(
+        &mut self,
+        shortcut_key: ShortcutKey,
+        now: Instant,
+    ) -> WidgetFlags {
+        self.penholder.handle_released_shortcut_key(
+            shortcut_key,
+            now,
+            &mut EngineViewMut {
+                tasks_tx: self.tasks_tx(),
+                pens_config: &mut self.pens_config,
+                doc: &mut self.document,
+                store: &mut self.store,
+                camera: &mut self.camera,
+                audioplayer: &mut self.audioplayer,
+            },
+        )
+    }
+
     /// Change the pen style.
     pub fn change_pen_style(&mut self, new_style: PenStyle) -> WidgetFlags {
         self.penholder.change_style(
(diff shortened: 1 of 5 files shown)


flxzt:
The code itself looks very good. However the implemented feature doesn't work as one would expect for all tools. I see two possible improvements:

- when the stylus is lifted and moved away from the drawing tablet, it seems that an "button released" event is emitted and the pen always switches back to the original one. So the user would need to keep the pen in proximity. We should detect this event and ignore it (if it doesn't break anything else) or fix this in some other way.
- In the current implementation one has to hold the button constantly, regardless of the state of the action that was performed while using the switched tool. For example, when using the selector I would expect that it switches permanently after the user has selected something. To be able to detect this we need to add a new state to `PenProgress` and refactor the selector to emit it while being in the selected state. We should probably do the same for the shaper/shape builders that have a multi-step build process.

Regarding the naming: "Hold" sounds fine to me :)

TecCheck:
Regarding the first point I personally feel like this is OK as the button is pressed again when the pen comes into proximity again.

For the second point I'd say the Temporary mode already covers this behavior so having it in the hold mode would be kind of redundant. However I don't think it's that hard to implement and if you want to have it I'll look into implementing it

flxzt:
I think the two suggested improvement would need to be implemented, most users would expect it to work like that.

TecCheck:
Ok then. I guess it will take some time for me to implement this as I haven't done all that much with rust. But I'll try :smile:

TecCheck:
I'm closing this PR as I don't plan on working on this code again. It's probably better to start fresh, but I cannot promise anything. I'm sorry, it didn't go anywhere

Doublonmousse:
Oh no it's okay, I think we can build from your PR to implement the feature.

It's not lost work if that's what you're worried about.

TecCheck:
I wouldn't mind if it's lost. But I would like to have this feature as it keeps me from using rnote :sweat_smile: I'll have more time soon. Maybe then I have the motivation to revive this

PR #734, opened 2023-07-02
Stroke content previews

flxzt:
This PR adds previews to the export dialogs.

This is done by extending `StrokeContent` which is now the used struct when exporting the document, pages and selections. All exports were refactored to use it so that the export always matches the preview (as long as the same arguments to the draw/export functions are passed)

For the preview there is now a `StrokeContentPaintable` which holds `StrokeContent` and is able to paint it. It also holds some of the export options and they are applied non-destructively when it gets painted.

`StrokeContentPreview` is a widget that wraps the paintable, holds `Vec<StrokeContent>` and an index, intended to be used for both single and multi-page content. When used for multiple pages it also then displays some controls to navigate through the pages.

Also: the pages export is now parallelized per-page, which decreases the export time significantly.

Things to discuss:

- caching: currently the paintable/preview does not cache the painted content. When using it with  expensive-to-render content it takes a while to repaint. It is most noticable when resizing the window.
- the export document dialog currently also displays the content page-by-page, which can be confusing. It would be more accurate to paint the entire document as a single content, but it would be too expensive to do without caching and not very useful without zoom functionality in the preview
- related to the previous point: zoom in the preview would of course be nice, it would need a similar implementation as in the canvas. But not needed for a MVP in my opinion.
- the dialogs could be more responsive, for example collapse into two pages when the width is reduced. There will be new API for such things in libadwaita 1.4 so I want to wait for it. Can also be a follow-up in my opinion.
- another follow-up that will be possible with this: the ability to include/exclude pages in the export. Similar to most print dialogs, where the user is able to type in ranges in a text field (e.g. `1-3,6,7` to export pages 1,2,3,6,7). 

@Kneemund this is also preliminary work for #670, will that work/what should be changed? I'll add a comment there as well explaining how i think it can be extended for the invert-colors feature.


rnote-ui/data/ui/strokecontentpreview.ui
Kneemund, line 1: The new line at the end seems to be missing.

rnote-ui/data/ui/strokecontentpreview.ui
```diff
@@ -0,0 +1,101 @@
...
+                <property name="input-purpose">digits</property>
+                <property name="max-width-chars">4</property>
+                <property name="xalign">0.5</property>
+                <property name="placeholder-text" translatable="yes">Page</property>
+                <style>
+                  <class name="osd" />
+                </style>
+              </object>
+            </child>
+          </object>
```
Kneemund, line 70: What about using a `GtkSpinButton` instead of implementing this manually? I guess the increment/decrement symbols would be different, and all page controls would be centered instead of being split up and aligned to the upper corners. 

I'm not sure how it'd affect the UX. Using a spin button (and having a single toolbar at the bottom) *might* look better when adding new elements for e.g. zoom in the future, but the way it is now also looks good and feels intuitive to me.

Kneemund (commented):
Very cool :)
The code looks good to me, and I didn't notice any bugs while testing it. 

> caching: currently the paintable/preview does not cache the painted content.

Yeah, that's definitely something that should be addressed.
Alternatively or in addition to caching, we could try to draw the `StrokeContentPaintable` *before* invalidating the `GtkPicture` so that it doesn't block the main thread - though I don't know how feasible that is. It's probably not possible to make this also work for internal/non-explicit invalidations like resizing the dialog window (except by displaying a placeholder until the preview is done I guess).

> the export document dialog currently also displays the content page-by-page, which can be confusing.

This only really applies to the exporting the document as SVG, right? But yeah, it is quite confusing. And the preview still respects the insensitive page order option.

> zoom in the preview would of course be nice, it would need a similar implementation as in the canvas. But not needed for a MVP in my opinion.

I agree, it would be good but it's not essential.

> the dialogs could be more responsive, for example collapse into two pages when the width is reduced. There will be new API for such things in libadwaita 1.4 so I want to wait for it. Can also be a follow-up in my opinion.

Yeah, breakpoints will be the best solution for this.

> another follow-up that will be possible with this: the ability to include/exclude pages in the export.

That would/will be nice, yeah :)

rnote-ui/data/ui/strokecontentpreview.ui
flxzt: I am unsure if the `+` and `-` buttons fit to navigating between pages. What about leaving it for now, but grouping the navigation buttons and page entry together when/if zooming gets added to keep it balanced visually?

Related: what do you think, should the controls be at the top or at the bottom?

flxzt:
> Alternatively or in addition to caching, we could try to draw the StrokeContentPaintable before invalidating the GtkPicture so that it doesn't block the main thread

I am not completely sure what you mean, painting with `gdk::Snapshot` in gtk is entirely single threaded, we need to use our own, `Vec<u8>` backed image type to be able to render in a different thread.

I refactored it that it now paints into an paint cache once when the content and/or the `draw_background`, `draw_pattern`, etc. properties have changed. The image dimensions are determined by the content size, capped to configurable `paint-max-width` and ´paint-max-height` values. Repainting the paint cache is not as automatic as I'd like it to be, and the paint cache currently is not updated when the picture size changes. When we want to render dynamically we'd need to do it asynchronously after a timeout that gets reset whenever the allocation changes. But the current implementation might already be good enough for now

Kneemund:
> I am not completely sure what you mean, painting with gdk::Snapshot in gtk is entirely single threaded, we need to use our own, Vec<u8> backed image type to be able to render in a different thread.

~~Ah, I thought it would be simpler to move the rendering part to a different thread and re-render changes in advance (before redrawing/invalidating the `GtkPicture`).~~
I meant exactly what you've done in this commit: 3edb3a61da138fbc2ed1d9429e6d795beac36df9. I guess you already had that in mind when talking about caching :)

rnote-ui/data/ui/strokecontentpreview.ui
Kneemund: Yeah, sounds good to me.

I think as long as they are floating buttons, it's fine if they're at the top. But if we add more buttons and functionality, we should (IMO) consider using a non-floating/non-overlapping toolbar instead, which probably should be at the bottom.

rnote-ui/data/ui/dialogs/export.ui
Kneemund, line 1: The margins are a bit messed up here I think, e.g. the horizontal margin of the entire dialog content is larger than the vertical one, the preferences group has its own inner margin and also an 18px right margin (which is larger than the 12px left margin of the entire content), ...

flxzt: ah true, fixed it. But I kept the right inner margin because of the scrollbar of the scrolledwindow that contains the preferences

flxzt:
> I meant exactly what you've done in this commit: https://github.com/flxzt/rnote/commit/3edb3a61da138fbc2ed1d9429e6d795beac36df9. I guess you already had that in mind when talking about caching :)

I see :+1: 

Okay, the remaining question is about the per-page preview in the doc export. I am leaning towards merging it, but with the change to always keep the page-order option sensitive. The subtext already hints that this option is only relevant for split-page export formats.

Merging it would also unblock your work and I can always follow up later with another PR that implements zoom, and together with it a different preview for the doc export.

rnote-ui/data/ui/strokecontentpreview.ui
flxzt: yeah I agree. I'll keep it in mind when implementing the follow-up

Kneemund:
> I am leaning towards leaving it as it is right now, but with the change to always keep the page-order option sensitive. The subtext already hints that this option is only relevant for split-page export formats.

Yeah, I think that's fine. We could also add it back once the different preview & zoom are implemented.

BTW, the right margin of export dialogs is still 6px bigger than the left one. Is that intended? Other than that, sounds and looks good to me :)

flxzt:
yes, that's intentional to give the the scrollbar enough space

PR #749, opened 2023-07-16
Update symbolic icons

oscfdezdz:
Noticed that some icons were outdated and updated them with the latest changes from https://gitlab.gnome.org/GNOME/adwaita-icon-theme.

`remove-page-symbolic` is a modification based on `add-page-symbolic` and `text-indent-more-symbolic` of `text-indent-less-symbolic`.


flxzt:
nice, thanks!

going through the icons, I noticed some differences in the icon scaling for a few icons. Some were intentionally scaled down by 0.5x because they are used in the 32px size in the app (mostly in the tool bars), and a full size icons are too large in those places. So the new ones would need that adjustment as well.

- pen-tools-offsetcameratool-symbolic
- pen-typewriter-fontchooser-symbolic

oscfdezdz:
By reducing them 0.5x they look much smaller than before, I guess it changed the size of the new ones.

flxzt:
yeah, but I am misremembering it, it probably wasn't exactly 0.5x. I'll quickly make adjustments and push to the branch.

oscfdezdz:
I went with 0.25x and it's still not the same size, but it's pretty similar.

flxzt:
ah you adjusted it already. Looks good to me now

PR #750, opened 2023-07-18
added --crop-to-content option to cli

### LINKED ISSUE #682: Rnote_cli: export only non-blank area
If I export a document using this command: `rnote_cli export --output-file test.svg test.rnote` I get the following output.

As you can see, it includes the whole page which can be a waste of visiblespace when one tries to attach this somewhere. 
 
The application already provides a "export selection" feature. It would be useful to add a flag in the rnote_cli tool that allows to export the smallest area that includes all the points in the document, also excluding any kind of background (like the dots in the picture above).

LeSnake04:
closes #682


LeSnake04:
I will add these changes as part of a bigger pr

PR #751, opened 2023-07-18
Cli enhancements

### LINKED ISSUE #682: Rnote_cli: export only non-blank area

LeSnake04:
This pr adds numerous improvements to the export CLI

- [x] toggles no longer need an Argument
- [x] sync export defaults with GUI export defaults (export with background/grid by default)
- [x] enable args_required_else_help for main CLI
- [x] use ValueEnum for --file-format
- [x] ask before overwriting files
  - [x] allow overwrites via option
- [x] added missing export options
- [x] refactor export into 3 subcommands
  - [x] selection
    - [x] all (closes #682)
    - [x] rect
  - [x] doc   
  - [x] doc-pages 

```
$ rnote-cli export --help
Exports the Rnote file(s) and saves it in the desired format.
 See subcommands for usage

Usage: rnote-cli export [OPTIONS] [RNOTE_FILES]... <COMMAND>

Commands:
  doc        Export entire document
                  When using --output-file, only one input file can be given.
                  The export format is recognized from the file extension of the output file.
                  When using --output-format, the file name and path of the rnote file is used with the extension changed.
                  --output-file and --output-format are mutually exclusive but one of them is required.
                  Currently `.svg`, `.xopp` and `.pdf` are supported.
                  Usages:
                  rnote-cli export doc --output-file [filename.(svg|xopp|pdf)] [1 file]
                  rnote-cli export doc --output-format [svg|xopp|pdf] [list of files]
  doc-pages  Export each page of the documents individually.
                  Both --output-dir and --output-format need to be set
                  Usage:
                  rnote-cli export doc-pages --output-dir [folder] --output-format [svg|png|jpeg] [list of files]
  selection  Export selection of a document
                  When using --output-file, only one input file can be given.
                  The export format is recognized from the file extension of the output file.
                  When using --output-format, the same file name is used with the extension changed.
                  --output-file and --output-format are mutually exclusive but one of them is required.
                  Usages:
                  rnote-cli export doc --output-file [filename.(svg|xopp|pdf)] [1 file] selection
                  rnote-cli export doc --output-format [svg|xopp|pdf] [list of files] selection
                  Available selection args: only use one of them
                  --all: select all strokes
                  --rect X Y deltaX deltaY : Select all strokes in given area

  help       Print this message or the help of the given subcommand(s)

Arguments:
  [RNOTE_FILES]...  the rnote save file

Options:
  -e, --on-conflict <ON_CONFLICT>  What to do if exported files already exits [default: ask] [possible values: ask, overwrite, skip, suffix]
  -b, --no-background              export without background
  -p, --no-pattern                 export without background pattern
  -h, --help                       Print help
```

```
rnote-cli export doc --help
Export entire document
 When using --output-file, only one input file can be given.
 The export format is recognized from the file extension of the output file.
 When using --output-format, the file name and path of the rnote file is used with the extension changed.
 --output-file and --output-format are mutually exclusive but one of them is required.
 Currently `.svg`, `.xopp` and `.pdf` are supported.
 Usages:
 rnote-cli export doc --output-file [filename.(svg|xopp|pdf)] [1 file]
 rnote-cli export doc --output-format [svg|xopp|pdf] [list of files]

Usage: rnote-cli export doc [OPTIONS] <--output-file <OUTPUT_FILE>|--output-format <OUTPUT_FORMAT>> [RNOTE_FILES]...

Arguments:
  [RNOTE_FILES]...  the rnote save file

Options:
  -e, --on-conflict <ON_CONFLICT>      What to do if exported files already exits [default: ask] [possible values: ask, overwrite, skip, suffix]
  -o, --output-file <OUTPUT_FILE>      the export output file. Only allows for one input file. Exclusive with --output-format
  -b, --no-background                  export without background
  -f, --output-format <OUTPUT_FORMAT>  the export output format. Exclusive with --output-file [possible values: pdf, xopp, svg]
  -p, --no-pattern                     export without background pattern
  -P, --page-order <PAGE_ORDER>        The page order when documents with layouts that expand in horizontal and vertical directions are cut into pages [default: horizontal] [possible values: horizontal, vertical]
  -h, --help                           Print help
```

```
$ rnote-cli rnote-cli export doc-pages --help
Export each page of the documents individually.
 Both --output-dir and --output-format need to be set
 Usage:
 rnote-cli export doc-pages --output-dir [folder] --output-format [svg|png|jpeg] [list of files]

Usage: rnote-cli export doc-pages [OPTIONS] --output-dir <OUTPUT_DIR> --output-format <OUTPUT_FORMAT> [RNOTE_FILES]...

Arguments:
  [RNOTE_FILES]...  the rnote save file

Options:
  -e, --on-conflict <ON_CONFLICT>
          What to do if exported files already exits [default: ask] [possible values: ask, overwrite, skip, suffix]
  -o, --output-dir <OUTPUT_DIR>
          the folder the pages get exported to
  -b, --no-background
          export without background
  -s, --output-file-stem <OUTPUT_FILE_STEM>
          the folder the pages get exported to
  -f, --output-format <OUTPUT_FORMAT>
          the export output format [possible values: svg, png, jpeg]
  -p, --no-pattern
          export without background pattern
  -P, --page-order <PAGE_ORDER>
          The page order when documents with layouts that expand in horizontal and vertical directions are cut into pages [default: horizontal] [possible values: horizontal, vertical]
      --bitmap-scalefactor <BITMAP_SCALEFACTOR>
          bitmap scale factor in relation to the actual size on the document [default: 1.8]
      --jpeg-quality <JPEG_QUALITY>
          quality of the jpeg image [default: 85]
  -h, --help
          Print help
```

```
$ rnote-cli export selection --help
Export selection of a document
 When using --output-file, only one input file can be given.
 The export format is recognized from the file extension of the output file.
 When using --output-format, the same file name is used with the extension changed.
 --output-file and --output-format are mutually exclusive but one of them is required.
 Usages:
 rnote-cli export doc --output-file [filename.(svg|xopp|pdf)] [1 file] selection
 rnote-cli export doc --output-format [svg|xopp|pdf] [list of files] selection
 Available selection args: only use one of them
 --all: select all strokes
 --rect X Y deltaX deltaY : Select all strokes in given area


Usage: rnote-cli export selection [OPTIONS] <--output-file <OUTPUT_FILE>|--output-format <OUTPUT_FORMAT>> <--all|--rect <X> <X> <X> <X>> [RNOTE_FILES]...

Arguments:
  [RNOTE_FILES]...  the rnote save file

Options:
  -e, --on-conflict <ON_CONFLICT>
          What to do if exported files already exits [default: ask] [possible values: ask, overwrite, skip, suffix]
  -o, --output-file <OUTPUT_FILE>
          the export output file. Only allows for one input file. Exclusive with --output-format
  -b, --no-background
          export without background
  -f, --output-format <OUTPUT_FORMAT>
          the export output format. Exclusive with --output-file [possible values: svg, png, jpeg]
  -a, --all
          export all strokes. Exclusive with --rect
  -p, --no-pattern
          export without background pattern
  -r, --rect <X> <X> <X> <X>
          Export an rectengular area of the canvas, exclusive with --all
           usage: X Y deltaX deltaY
           Goes to given coordiates and selects all strokes in a given rectangle based of the given delta values
  -i, --bounds <BOUNDS>
          if the lines inside or intersecting with the given bounds are exported. Exclusive with --all [default: inside] [possible values: inside, intersecting]
      --bitmap-scalefactor <BITMAP_SCALEFACTOR>
          bitmap scale factor in relation to the actual size on the document [default: 1.8]
      --jpeg-quality <JPEG_QUALITY>
          quality of the jpeg image [default: 85]
      --margin <MARGIN>
          margin around the document [default: 12]
  -h, --help
          Print help
```


flxzt:
Thanks!

Some notes:

- the negative phrased `without-pattern` / `without-background` options are unintuitive. What was wrong about `with-...`, with the default set to true? Also the default for those options already was matching with the UI, the `Default::default()` values are used when they were not specified as cli arguments.
- do we want to prompt when a file conflict is detected? Most standard unix tools would overwrite the file afaik. If we do, how does it work when used in non-interactive contexts, like scripts? I am not very familiar with `dialoguer`, did you test that?
- `crop-to-content` is also a bit unintuitive, especially because it is not clear that the available output file formats are different and that it is essentially entirely different export mode. I would suspect that most users would expect that `crop-to-content´ will not require any changes to the other export options. In my opinion it should be renamed. Also: the documentation for `crop-to-content` is wrong, and documentation that some of the other options require it is missing entirely.

I have more (minor) things, but we should dicsuss the higher-level questions about the feature first before digging into the code line-by-line

flxzt:
A random thought, and no expectionations about implementing it this way, but a nice way to implement more export options would be to refactor the export to `export-doc`, `export-doc-pages`, `export-selection` subcommands. Then we could add arguments analogous to the UI, like `--select-all`, `--select-region`, etc. . It would then be more clear which options are supported in which export subcommand

LeSnake04:
I Will try to find a way to test it non-interactive. I assume it will show an error.

We should default to overwriting when the shell  is non-interactive.

While standard unix tools do overwrite by default, I think it isnt a good standard. Users might have an assigment.pdf and an assigment.rnote, so not prompting would lead to the original assignment being overwritten. 
This cli is used differently from something like cp, especially since we generate the output file path. 

When the user explicitly sets an --output-file, overwriting by default is fine IMO

LeSnake04:
I am not sure what would be a better name for crop-to-content

LeSnake04:
That sounds like a good idea. I can do that and turn export --crop-to-content into export selection --all

about the without: I would consider bool args bad practice in cli. But I agree without isnt great, so how about --no-background -no-pattern?

LeSnake04:
I implemented it the way you suggested. See pr description for the --help.

I don't understand how parry2d works. Can you please tell me how to get an Aabb from the top left and button right point of a rectangle?

LeSnake04:
Ready for next review.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,715 @@
...
+                    "Failed to export page {i} of document {}",
+                    rnote_file.display()
+                ))?
+            }
+        }
+    };
+    Ok(())
+}
+
+async fn export_doc_page(
```
flxzt, line 695: This function name is wrong, the export happened already earlier. Also the user needs to be asked prior to actually exporting.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,715 @@
...
+    Selection {
+        #[command(flatten)]
+        file: FileArgs<SelectionOutputFormat>,
+        #[command(flatten)]
+        selection: SelectionArgs,
+        #[arg(short = 'i', long, default_value_t = Bounds::default(), conflicts_with = "all")]
+        /// if the lines inside or intersecting with the given bounds are exported. Exclusive with --all
+        bounds: Bounds,
+        /// bitmap scale factor in relation to the actual size on the document
+        #[arg(long, default_value_t = 1.8)]
```
flxzt, line 91: here as well: is it possible to use the default impl from SelectionExportOptions?

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,715 @@
...
+    /// Available selection args: only use one of them {n}
+    /// --all: select all strokes {n}
+    /// --rect X Y deltaX deltaY : Select all strokes in given area {n}
+    Selection {
+        #[command(flatten)]
+        file: FileArgs<SelectionOutputFormat>,
+        #[command(flatten)]
+        selection: SelectionArgs,
+        #[arg(short = 'i', long, default_value_t = Bounds::default(), conflicts_with = "all")]
+        /// if the lines inside or intersecting with the given bounds are exported. Exclusive with --all
```
flxzt, line 88: rename `lines` to `strokes`

crates/rnote-cli/src/cli.rs
```diff
@@ -32,37 +34,53 @@ pub(crate) enum Commands {
...
-        #[arg(short = 'b', long)]
-        with_background: Option<bool>,
-        /// export with background pattern
-        #[arg(short = 'p', long)]
-        with_pattern: Option<bool>,
+        /// What to do if exported files already exits
+        #[arg(short = 'e', long, default_value = "ask", global = true)]
+        on_conflict: OnConflict,
+        /// export without background
+        #[arg(short = 'b', long, action = ArgAction::SetTrue, global = true)]
```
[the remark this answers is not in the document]
LeSnake04: That's impossible with clap. The short versions need to be 1 letter. 
Also, it's common that the short version is not be negated for negated long args.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,715 @@
...
+        bounds: Bounds,
```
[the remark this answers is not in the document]
LeSnake04: I wasn't happy with that name either, but had no better idea. 

An alternative idea I had was making a --intersecting flag instead and use inside when it's not provided

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,715 @@
...
+#[derive(Args, Debug)]
+#[group(required = true, multiple = false)]
+pub(crate) struct SelectionArgs {
+    /// export all strokes. Exclusive with --rect
+    #[arg(short = 'a', long, action = ArgAction::SetTrue)]
+    all: bool,
+    /// Export an rectengular area of the canvas, exclusive with --all  {n}
+    /// usage: X Y deltaX deltaY {n}
+    /// Goes to given coordiates and selects all strokes in a given rectangle based of the given delta values
+    #[arg(short = 'r', long, value_name = "X", num_args = 4)]
```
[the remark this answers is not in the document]
LeSnake04: I don't think so. The value name is always the same for each input

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,715 @@
...
+                        let dy = rect[3];
+                        let v1 = Vector2::new(x, y);
+                        let v2 = v1 + Vector2::new(dx, dy);
+                        let points = vec![v1.into(), v2.into()];
+                        let aabb = Aabb::from_points(&points);
+                        (
+                            match bounds {
+                                Bounds::Inside => engine.store.keys_sorted_chrono_in_bounds(aabb),
+                                Bounds::Intersecting => {
+                                    engine.store.keys_unordered_intersecting_bounds(aabb)
```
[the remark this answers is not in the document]
LeSnake04: I changed it to sorted_chono. Do you mean that?

crates/rnote-cli/src/export.rs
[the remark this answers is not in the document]
LeSnake04: We could do that with another subcommand. I will try that

crates/rnote-cli/src/cli.rs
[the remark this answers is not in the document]
LeSnake04: example from cp:
  -T, --no-target-directory    treat DEST as a normal file

crates/rnote-cli/src/export.rs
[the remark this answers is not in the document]
flxzt: yes

crates/rnote-cli/src/cli.rs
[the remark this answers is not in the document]
flxzt: that's unfortunate. Okay then keep it.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,724 @@
...
+        /// The directory the pages get exported to.{n}
+        /// Uses file stem of input file if not set.{n}
+        /// Cannot be used when exporting multiple files
+        #[arg(short = 's', long)]
+        output_file_stem: Option<String>,
+        /// the export output format
+        #[arg(short = 'f', long)]
+        output_format: DocPagesOutputFormat,
+        /// The page order when documents with layouts that expand in horizontal and vertical directions are cut into pages
+        #[arg(short = 'P', long, default_value_t = PageOrder::default())]
```
flxzt, line 60: Considering that this is not an option that is that frequently used, the short version can be removed. It avoid cluttering the namespace with `-p` *and* `-P`.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,724 @@
...
+    /// When using --output-file, only one input file can be given.{n}
+    /// The export format is recognized from the file extension of the output file.{n}
+    /// When using --output-format, the file name and path of the rnote file is used with the extension changed.{n}
+    /// --output-file and --output-format are mutually exclusive but one of them is required.{n}
+    /// Currently `.svg`, `.xopp` and `.pdf` are supported.{n}
+    Doc {
+        #[command(flatten)]
+        file: FileArgs<DocOutputFormat>,
+        /// The page order when documents with layouts that expand in horizontal and vertical directions are cut into pages.
+        #[arg(short = 'P', long, default_value_t = PageOrder::default())]
```
flxzt, line 41: see the other comment. The short version can be removed

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,715 @@
...
+    Intersecting,
+}
+
+impl Display for Bounds {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", format!("{self:?}").to_lowercase())
+    }
+}
+
+#[derive(ValueEnum, Clone, Debug)]
```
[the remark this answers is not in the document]
LeSnake04: I tried that but seems like features don't work well with workspace dependency and I also couldn't get optional dependencies working directly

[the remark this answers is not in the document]
flxzt: Hm, I don't think it is that complicated. I quickly implemented it for `DocExportFormat`. Check out the `clap-derive-showcase` branch to see how it can work.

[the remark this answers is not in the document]
LeSnake04: see https://github.com/flxzt/rnote/commit/33315db8eaaad7bd737c296065a5e37673d537c6#commitcomment-123206327

crates/rnote-cli/src/export.rs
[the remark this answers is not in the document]
LeSnake04: Do you have a better name for this argument, do we keep it as is, or should I turn it into a boolean flag instead?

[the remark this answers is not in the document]
flxzt: what about `SelectionCollision` ? And I'd move this type into the engine, for future usage in the UI

LeSnake04:
I resolved all the suggestions. Is anything else needed for merging?

crates/rnote-cli/src/cli.rs
```diff
@@ -32,37 +34,66 @@ pub(crate) enum Commands {
...
-        with_background: Option<bool>,
-        /// export with background pattern
-        #[arg(short = 'p', long)]
-        with_pattern: Option<bool>,
+        /// The action that will be performed if the to be exported file(s) already exist(s)
+        #[arg(long, default_value = "ask", global = true)]
+        on_conflict: OnConflict,
+        /// export without background
+        #[arg(short = 'b', long = "no-background", action = ArgAction::SetFalse, global = true)]
+        background: bool,
```
flxzt, line 51: I am confused. Is this now supposed to be negated or not? Because it seems that you have changed it to non-negated after the review and discussion in Matrix, but missed updating the long-argument string.

crates/rnote-cli/src/cli.rs
```diff
@@ -32,37 +34,66 @@ pub(crate) enum Commands {
...
+
+impl std::fmt::Display for OnConflict {
+    fn fmt(&self, f: &mut std::fmt::Formatter) -> std::fmt::Result {
+        write!(
+            f,
+            "{}",
+            match self {
+                Self::Ask => "Open existing file for inspection and ask again",
+                Self::Overwrite => "Overwrite existing file",
+                Self::AlwaysOverwrite => "Always Overwrite exitings file",
```
flxzt, line 87: Here are some typos: `existings` -> `existing` and `Overwrite` should be lowercase.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+        margin: f64,
+    },
+}
+
+#[derive(Subcommand, Debug)]
+pub(crate) enum SelectionCommands {
+    /// Export all strokes
+    All,
+    /// Export a rectangular area of the document{n}
+    /// usage: X Y deltaX deltaY{n}
```
flxzt, line 105: remove this line, it is outdated.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+    /// Goes to given coordiates and selects all strokes in a given rectangle based of the given delta values
```
flxzt, line 106: This is also outdated, I think the line above should be enough to explain it.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+use crate::validators;
+
+#[derive(Subcommand, Debug)]
+pub(crate) enum ExportCommands {
+    /// Export the entire document.{n}
+    /// When using --output-file, only one input file can be given.{n}
+    /// The export format is recognized from the file extension of the output file.{n}
+    /// When using --output-format, the file name and path of the rnote file is used with the extension changed.{n}
+    /// --output-file and --output-format are mutually exclusive but one of them is required.{n}
+    /// Currently `.svg`, `.xopp` and `.pdf` are supported.{n}
```
flxzt, line 36: Remove this line, the possible options are listed and this feels out of context here.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+        /// quality of the jpeg image
+        #[arg(long, default_value_t = DocPagesExportPrefs::default().jpeg_quality)]
+        jpeg_quality: u8,
+    },
+    /// Export a selection in a document.{n}
+    /// When using --output-file, only one input file can be given.{n}
+    /// The export format is recognized from the file extension of the output file.{n}
+    /// When using --output-format, the same file name is used with the extension changed.{n}
+    /// --output-file and --output-format are mutually exclusive but one of them is required.{n}
+    /// Currently `.svg`, `.png` and `.jpeg` are supported.{n}
```
flxzt, line 74: Same thing here, remove this line.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+    /// Available selection arguments - one of:{n}
+    /// --all: select all strokes{n}
+    /// --rect X Y width height: Select all strokes in given area{n}
```
flxzt, line 77: this is also outdated because they are now subcommands, not arguments anymore

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+        if i + 1 < 10 { "0" } else { "" },
+        i + 1,
+    ));
+    if let Some(new_out) = check_file_conflict(out.as_ref(), on_conflict, on_conflict_overwrite)? {
+        out = new_out;
+    }
+    Ok(out)
+}
+
+async fn store_export(output_file: impl AsRef<Path>, bytes: &[u8]) -> anyhow::Result<()> {
```
flxzt, line 668: What about naming this `write_bytes_to_file` ?

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+    i: usize,
+    output_dir: &Path,
+    out_ext: &str,
+    output_file_stem: &str,
+    on_conflict: OnConflict,
+    on_conflict_overwrite: &mut Option<OnConflict>,
+) -> anyhow::Result<PathBuf> {
+    let mut out = output_dir.join(format!(
+        "{output_file_stem} - page {}{}.{out_ext}",
+        if i + 1 < 10 { "0" } else { "" },
```
flxzt, line 659: I believe there is a formatting option for this. It would break if files are in the hundreds or more. If you want to be clean, count the number of files and specify the leading zeros based no that.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+                OnConflict::AlwaysOverwrite,
+                OnConflict::Skip,
+                OnConflict::AlwaysSkip,
+                OnConflict::Suffix,
+                OnConflict::AlwaysSuffix,
+            ];
+            while matches!(on_conflict, OnConflict::Ask) {
+                match dialoguer::Select::new()
+                    .with_prompt(format!("File {} already exits:", output_file.display()))
+                    .items(options)
```
flxzt, line 473: Is it possible to make the first option the default selected? Because right now it is not entirely clear that this is a prompt when it appears, and a initial `>` might make it more clear that the user is able to interact with it.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+                OnConflict::Overwrite,
+                OnConflict::AlwaysOverwrite,
+                OnConflict::Skip,
+                OnConflict::AlwaysSkip,
+                OnConflict::Suffix,
+                OnConflict::AlwaysSuffix,
+            ];
+            while matches!(on_conflict, OnConflict::Ask) {
+                match dialoguer::Select::new()
+                    .with_prompt(format!("File {} already exits:", output_file.display()))
```
flxzt, line 472: Here's another typo.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+                    .items(options)
+                    .interact()
+                {
+                    Ok(0) => {
+                        if let Err(e) = open::that(output_file) {
+                            println!(
+                                "Failed to open {} with default program, {e}",
```
flxzt, line 479: print error messages with `{e:?}`. This applies to the next error message right below this one as well.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+                validators::file_has_ext(rnote_file, "rnote")?;
+                let new_path = if doc_pages {
+                    // conflicts are allowed here, generated output path will be ignored
+                    None
+                } else {
+                    match check_file_conflict(output_file, on_conflict, &mut on_conflict_overwrite)
+                    {
+                        Ok(r) => r,
+                        Err(e) => {
+                            println!("{e}");
```
flxzt, line 284: This needs a better error message (and `{e:?}`)

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+    /// --output-file and --output-format are mutually exclusive but one of them is required.{n}
+    /// Currently `.svg`, `.png` and `.jpeg` are supported.{n}
+    /// Available selection arguments - one of:{n}
+    /// --all: select all strokes{n}
+    /// --rect X Y width height: Select all strokes in given area{n}
+    /// When not using --all, use --bounds to switch between intersecting and inside bounds
+    #[command(alias = "sel")]
+    Selection {
+        #[command(flatten)]
+        file: FileArgs<SelectionExportFormat>,
```
flxzt, line 82: name this file_args, file as variable name is confusing.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+pub(crate) enum ExportCommands {
+    /// Export the entire document.{n}
+    /// When using --output-file, only one input file can be given.{n}
+    /// The export format is recognized from the file extension of the output file.{n}
+    /// When using --output-format, the file name and path of the rnote file is used with the extension changed.{n}
+    /// --output-file and --output-format are mutually exclusive but one of them is required.{n}
+    /// Currently `.svg`, `.xopp` and `.pdf` are supported.{n}
+    Doc {
+        #[command(flatten)]
+        file: FileArgs<DocExportFormat>,
```
flxzt, line 39: name this `file_args`, `file` as variable name is confusing.

crates/rnote-cli/src/export.rs
```diff
@@ -0,0 +1,673 @@
...
+                            println!(
+                                "Failed to open {} with default program, {e}",
+                                output_file.display()
+                            );
+                        }
+                    }
+                    Ok(c) => on_conflict = options[c],
+                    Err(e) => {
+                        return Err(anyhow::anyhow!(
+                            "Failed to show select promt, retry or select an behavior with --on-conflict, {e}"
```
flxzt, line 487: Here's another typo.

flxzt:
To be honest the code is really hard to read, and I feel like it will need to be refactored as soon as more features get added to the cli. Especially the logic that determines the output files/the dialog is very convoluted. It needs more clean up before it can be merged.

crates/rnote-cli/src/cli.rs
LeSnake04: The Arg isn't changed, see the value given to no-background:  the variable is set to true by default and set to false when its provided. I think it makes the code a bit cleaner, but if you prefer the value being negated in the code as well, I can reverse that change.

crates/rnote-cli/src/export.rs
LeSnake04: I made it default to overwrite since it's the option that will likely be chosen the most

crates/rnote-cli/src/export.rs
LeSnake04: I moved the explanation of the inputs to the individual Arg documentations

crates/rnote-cli/src/export.rs
LeSnake04: I changed it to say that --output-file currently supports ...

Or do you think the possible file extensions are clear from the possible values for output-format?

LeSnake04:
I extracted some parts of run_export into extra functions, including the part about the output file to make the code more readable. Most of the code left in there is for error handling which is hard to extract further.

crates/rnote-cli/src/cli.rs
flxzt: I see, but revert it please.

crates/rnote-engine/src/selectioncollision.rs
```diff
@@ -0,0 +1,23 @@
...
+#[cfg_attr(feature = "clap-derive", derive(clap::ValueEnum))]
+pub enum SelectionCollision {
+    #[default]
+    /// All strokes completely inside the area
+    Contains,
+    /// All Strokes intersecting with the area
+    Intersects,
+}
+
+#[cfg(feature = "clap-derive")]
```
flxzt, line 11: remove this condition, implementing `Display` does not need to be conditional to the clap-derive feature

crates/rnote-engine/src/lib.rs
```diff
@@ -18,6 +18,7 @@ pub mod ext;
 pub mod fileformats;
 pub mod pens;
 pub mod render;
+pub mod selectioncollision;
```
flxzt, line 21: Re-Export the enum below to avoid having to import `selectioncollision::SelectionCollision`

crates/rnote-compose/src/splitorder.rs
```diff
@@ -12,20 +12,33 @@ use serde::{Deserialize, Serialize};
...
     num_derive::ToPrimitive,
+    Default,
 )]
+#[cfg_attr(feature = "clap-derive", derive(clap::ValueEnum))]
 #[serde(rename = "split_order")]
 pub enum SplitOrder {
+    #[default]
     /// Split in row-major order.
     #[serde(rename = "row_major")]
+    #[cfg_attr(feature = "clap-derive", value(name = "horizontal"))]
```
flxzt, line 23: Rename this to `horizontal-first`, `vertical-first` to match the export option in the GUI

crates/rnote-cli/src/export.rs
flxzt: yes, it is enough when they are listed at the help text of the `--output-format` argument

flxzt:
alright, there are still some minor things that I'd name/do differently but all in all it looks merge-able.

Can you resolve the conflicts?

flxzt:
thanks!

PR #792, opened 2023-09-11
Move shape builders button & constraints button to top and shaper style buttons to bottom

### LINKED ISSUE #230: Order of shaper tool option
The tool “Shape builders” button should be on the top.
Because the most far-reaching setting is what i want to draw.
All the other options are just attributes of the shape.

The Constraints are pretty shape independent and should not be clustered with the Shape builders (and should have their own icon).

PhilDevProg:
This Pull request moves the shape builders button and the constraints button to the top and the shaper style buttons to the bottom of the shaper sidebar.

Fixes #230

Bildschirmfoto vom 2023-09-14 13-08-14

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -2,3 +2,4 @@ Felix Zwettler
 Óscar Fernández Díaz
 Seio Inoue
 Moritz Mechelk
+PhilProg
--- a/crates/rnote-ui/data/ui/penssidebar/shaperpage.ui
+++ b/crates/rnote-ui/data/ui/penssidebar/shaperpage.ui
@@ -47,6 +47,15 @@
       </object>
     </child>
     <child>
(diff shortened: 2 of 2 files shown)


flxzt:
thanks. The only "issue" with this is that the stroke width picker is not at the same position between brush, shaper and eraser any more. I am unsure if this is that important though.

PhilDevProg:
> thanks. The only "issue" with this is that the stroke width picker is not at the same position between brush, shaper and eraser any more. I am unsure if this is that important though.

I could move the shape style buttons to the bottom. I don't think it's important to have them above the stroke width picker.

flxzt:
yeah that sounds good :) btw. add yourself to `AUTHORS` if you want

flxzt:
looks good!

Issue #818, opened 2023-10-02
Auto-switch to brush tool if I pick a color with the eraser


Mitia99:
**Is your feature request related to a problem? Please describe.**  
It's not convenient that after using the eraser tool, I always have to go to brush tool then choose a color to draw.
I mean, there are brush with different colors shortcuts in other whiteboard apps toolbar.

**Describe the solution you'd like**  
If the tool I choose is an eraser, the app should auto-switch to brush if I select a color.
It's kinda non-sense to be able to pick a color if the tool selected is the eraser.

PR #824, opened 2023-10-08
#818 - Auto-switch to brush tool when picking a color with the eraser

xweskingx:
*  When a color is selected, either fill or stroke, switch to brush


flxzt:
thanks! I wonder if this should be extended to the other non-color pens (selector, tools) as well? What are your thoughts?

flxzt:
I think this is not a change that should be included.
Changing the color should not trigger a switch, since it is also used in other context (like text, selections, shapes). Doing it only for specific tools (eraser, tools) might be unexpected, not what the user wants (one could argue: why not switch to another tool like shapes instead?) and not really good UX.

Issue #812, opened 2023-09-23
Typewriter formatting improvments.


FedericoCalzoni:
**Is your feature request related to a problem? Please describe.**  
If the text is bold and underlined to remove the underline it is necessary to remove the formatting with the "reset text attributes" button and press again the bold. 

**Describe the solution you'd like**  
Use the same behavior of other writing programs like libreoffice or word, removing the "reset text attributes" button. 

**Additional context**  
Shortcuts for formatting should be added. ctrl+b for bold, ctrl+i for italic and ctrl+u for underline

flxzt:
this is now implemented (see #829).

PR #829, opened 2023-10-10
#812 - Typewriter formatting improvments

xweskingx:
- Updates bold, italic, strikethrough, and underline buttons to act as a toggle for selected text 
- If the selected text is half toggled, then toggle that value for the entire selection


flxzt:
looks good, thank you!

PR #834, opened 2023-10-11
Port to new libadwaita specialized rows

oscfdezdz:


flxzt:
thanks!

The only thing I noticed is that you didn't update the widget variable names in the closures for the `connect_*_notify`/`connect_changed` callbacks. Otherwise it looks good to me.

Porting the custom widget `RnUnitEntry` to a style similar to spin rows should be looked into as well, but it doesn't need to be in this PR.

oscfdezdz:
> The only thing I noticed is that you didn't update the widget variable names in the closures for the `connect_*_notify`/`connect_changed` callbacks. Otherwise it looks good to me.
> 
> Porting the custom widget `RnUnitEntry` to a style similar to spin rows should be looked into as well, but it doesn't need to be in this PR.

They are now addressed!

crates/rnote-ui/data/ui/unitentry.ui
```diff
@@ -23,7 +20,9 @@
     </child>
     <child>
       <object class="GtkDropDown" id="unit_dropdown">
+        <property name="margin-start">12</property>
```
flxzt, line 23: A margin of 6 should be enough here

flxzt:
looks good! I like that it brings a more consistent look between different types of rows.

Issue #220, opened 2022-07-15
Suggestion for Chinese translations


BobH-Official:
I found that the completion of `Simplified Chinese`translation is way higher than the `Traditional Chinese` ( 90+% vs 70+%).
As `Simplified Chinese` and `Traditional Chinese` are the **SAME** language but in two different **WRITING FONTS**, I suggest that we just keep the Simplified Chinese active, and use the converter tool OpenCC([https://github.com/BYVoid/OpenCC](https://github.com/BYVoid/OpenCC)) to automatically generate the Traditional version of Chinese.
In this way, translators no longer need to maintain two **SAME** languages, and their work should be easier.

flxzt:
This is a good idea. Unfortunately weblate treats simplified and traditional as seperate languages, I don't think it is possible to convert them automatically there. Either a Github Action could be set up, or it could be done manually. Is it possible to convert .po files with `opencc`?. If yes, you could download the simplified chinese file from weblate, convert it and create a PR with the updated traditional chinese file. Would you be willing to do that?

BobH-Official:
It is possible to do it with `opencc`. Because now it is mid-night here in China, I will probably do it tomorrow.
But before converting, I would like to know what does the term [`perm.`](https://hosted.weblate.org/translate/rnote/repo/zh_Hans/?checksum=edee9ee7aff9e450&sort_by=-priority,position) actually means. I found it confusing, and cannot translate it.

flxzt:
No worries, there is no rush. :)
Ah that one probably needs a replacement by an icon or something. It is the abbreviation for permanent, it is the prefix of a checkbox.
![Bildschirmfoto vom 2022-07-17 18-37-27](https://user-images.githubusercontent.com/19841886/179413200-e6ce4dfb-67b9-4850-b21c-5b990d698dc7.png)

BobH-Official:
OK, I will leave it as is. And that makes the Han_s to be 100% complete (for now).

BobH-Official:
I have updated Chinese Traditional manually.
I used the config file `openCC\share\opencc\s2twp.json`.

flxzt:
Thanks! Can you provide the exact command for conversion? I believe it should be possible to automate it.

BobH-Official:
Yes, it can be done automatically. You can use openCC as either a `bin` or a `lib`. There are API for nodejs, python, cpp and etc. 
[...]

PR #838, opened 2023-10-14
fix linux chinese locale

detiam:
On Linux, there are no `zh_Hans` nor `zh_Hant` supported:
```shell
$ cat /usr/share/i18n/SUPPORTED | grep zh
lzh_TW UTF-8
zh_CN.GB18030 GB18030
zh_CN.GBK GBK
zh_CN.UTF-8 UTF-8
zh_CN GB2312
zh_HK.UTF-8 UTF-8
zh_HK BIG5-HKSCS
zh_SG.UTF-8 UTF-8
zh_SG.GBK GBK
zh_SG GB2312
zh_TW.EUC-TW EUC-TW
zh_TW.UTF-8 UTF-8
zh_TW BIG5
```
So the current Chinese translate actually not been applied and used under Linux.
The working way right now is to use `zh_CN`, `zh_SG` for Simplified Chinese and `zh_HK`, `zh_TW` for Traditional Chinese, like this pr will do, or this change can be done from Weblate side?

### DIFF
--- a/crates/rnote-ui/po/LINGUAS
+++ b/crates/rnote-ui/po/LINGUAS
@@ -9,7 +9,11 @@ pt_BR
 fr
 nb_NO
 zh_Hans
+zh_CN
+zh_SG
 zh_Hant
+zh_HK
+zh_TW
 pl
 tr
 hu
--- a/crates/rnote-ui/po/zh_CN.po
+++ b/crates/rnote-ui/po/zh_CN.po
@@ -1,0 +1,1 @@
+zh_Hans.po
--- a/crates/rnote-ui/po/zh_HK.po
+++ b/crates/rnote-ui/po/zh_HK.po
@@ -1,0 +1,1 @@
+zh_Hant.po
--- a/crates/rnote-ui/po/zh_SG.po
+++ b/crates/rnote-ui/po/zh_SG.po
@@ -1,0 +1,1 @@
+zh_Hans.po
--- a/crates/rnote-ui/po/zh_TW.po
+++ b/crates/rnote-ui/po/zh_TW.po
@@ -1,0 +1,1 @@
+zh_Hant.po


flxzt:
I am still confused about `zh_hans` and `zh_hant`. Aren't these "meta" locale that get matched to the specific locale that the user has set?

I *think* the best way to do this is to open a new languages in weblate.

detiam:
> I am still confused about `zh_hans` and `zh_hant`. Aren't these "meta" locale that get matched to the specific locale that the user has set?

It's supposed, but upstream hasn't implemented it, and it's more than one year after this issue was submitted now:
https://gitlab.gnome.org/GNOME/glib/-/issues/2610
> Does referring to zh_hans/zh_hant work for the added po files, did you verify that?

verified
image

> I would need to test if weblate does not overwrite the contents of the added po files when it applies upstream changes.

`po/LINGUAS` will be overridden, but I see we are manually handling this file in rnote right?

>  I _think_ generally the best way to do this is to open new languages in weblate.

`zh_CN` and `zh_SG`, `zh_HK` and `zh_TW` is basically one writing system with different timezone, time format and such, add them separately will add unnecessary workload of copying strings in weblate.

On Linux we also can soft link 
`/usr/share/locale/zh_Hans/LC_MESSAGES/rnote.mo` to `/usr/share/locale/{zh_CN,zh_SG}/LC_MESSAGES/rnote.mo`, `/usr/share/locale/zh_Hant/LC_MESSAGES/rnote.mo` to `/usr/share/locale/{zh_HK,zh_TW}/LC_MESSAGES/rnote.mo` to resolve this problem, but I don't know how to do it when `meson install`, and the case of macos and windows.

flxzt:
Softlinks would not be a good solution.

I'll try the following: I added a [language filter](https://docs.weblate.org/en/weblate-5.1.1/admin/projects.html#language-filter) in weblates settings to exclude the languages you are adding here.
I used this regex: `^(?!(zh_CN|zh_SG|zh_HK|zh_TW)$).+$`

I'll merge, hopefully weblate will just ignore the added languages.

flxzt:
seems to work.

A thing that is related to this is the translation between `zh_hans` vs `zh_hant`: There was some discussion in #220, essentially saying that it is possible to automatically generate the traditional chinese translation from the simplified. Currently I am doing this manually from time to time, but ideally this would be a CI workflow that is triggered whenever the simplified translation changes.

flxzt:
unfortunately weblate removes the added translations from `LINGUAS` again, as seen in the pending PR (#849)

detiam:
> unfortunately weblate removes the added translations from `LINGUAS` again, as seen in the pending PR (#849)
> 
> EDIT: the addon that is responsible for updating LINGUAS unfortunately does not have any configuration that could change this behaviour (see https://docs.weblate.org/en/latest/admin/addons.html#update-linguas-file). But I just checked, the translations/symlinks are still compiled to `.mo` files by meson's i18n module, even without an entry in `LINGUAS`

@flxzt  New released [Arch Linux package](https://archlinux.org/packages/extra/x86_64/rnote/) not contain `.mo` file for `zh_CN zh_SG zh_HK zh_TW`, and when I try to build myself
image
still not there, maybe it's windows and Linux behave different? Seems on Linux the locale code have to be added to `LINGUAS` file.

flxzt:
hm, okay. I guess then it is necessary.

flxzt:
yeah, that can be a workaround. It feels a bit hacky, but it's probably the only option to make it work currently.

detiam:
> yeah, that can be a workaround. It feels a bit hacky, but it's probably the only option to make it work currently. It could be in the same workflow that is getting added by #850

done

PR #847, opened 2023-10-23
Add description of flatpak and meson

TristanTrim:
When I started looking at this project I was unfamiliar with flatpak and meson. It would have been helpful to know this upfront.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -8,7 +8,7 @@ git submodule update --init --recursive
 ```
 
 # Building with Flatpak vs Meson
-This project can be compiled two different ways depending on your need: flatpak or meson.
+This project can be compiled in two different ways depending on your needs: flatpak or meson.
 
 Flatpak is a sandboxed environment/distribution used for building and running applications in a way that is more user friendly and cross platform. When using flatpak to build an application, flatpak creates a sandboxed environment tailered to exactly what the application needs. This makes it much easier to compile and run an application without issues.
 
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -7,6 +7,13 @@ cd rnote
 git submodule update --init --recursive
 ```
 
+# Building with Flatpak vs Meson
+This project can be compiled with two different build systems: flatpak or meson.
+
+Flatpak is a high level build system that allows easy building and targeting other operating systems by creating a container to build inside of, but for incremental development, this can be very slow, and meson should be used.
+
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -7,6 +7,13 @@ cd rnote
 git submodule update --init --recursive
 ```
 
+# Building with Flatpak vs Meson
+This project can be compiled with two different build systems: flatpak or meson.
+
+Flatpak is a high level build system that allows easy building and targeting other operating systems by creating a container to build inside of, but for incremental development, this can be very slow, and meson should be used.
```
flxzt, line 13: I wouldn't call it a build system, rather a sandboxed environment/distribution.

BUILDING.md
```diff
@@ -7,6 +7,13 @@ cd rnote
...
+
+Meson is the build system used inside of flatpak's containers. It consists of first configuring the build with meson, and then building it with ninja. Because it is building on the host machine, it may require more upfront work managing dependencies, but then compiling changes made to the codebase with ninja is very fast.
```
flxzt, line 15: I'd phrase this: "meson is the build system that Rnote uses for building the application. It is called when the flatpak is built. It is also possible to use meson directly on the host [..]" Also it's not really necessary to mention ninja, imo. Everything can be done through meson.

flxzt:
thanks! I appreciate the perspective where new contributors might get stuck/where more clarification is needed.

TristanTrim:
I rephrased some things. Please check it out : )

I'm far from knowledgeable about meson or flatpak, so I value you're corrections. I included the stuff about ninja because I found ninja compiled faster than using meson directly, but I realize you don't have to look very far into meson to find out about using ninja directly.

Flatpak on the other hand, I found quite confusing at first before I understood it's purpose is for platform independence, not for compiling during development. Right?

I also got the mistaken idea that since this was a rust application, I should be compiling with cargo. A more careful read of the BUILDING.md and I probably wouldn't have made that mistake, but I think this message helps others not go down the path I did without being too much bother for those already familiar with flatpak and meson.

BUILDING.md
```diff
@@ -8,11 +8,11 @@ git submodule update --init --recursive
 ```
 
 # Building with Flatpak vs Meson
-This project can be compiled with two different build systems: flatpak or meson.
+This project can be compiled two different ways depending on your need: flatpak or meson.
```
flxzt, line 11: a "in" is missing here and "need" -> "needs"

PR #850, opened 2023-10-27
Add action to update chinese translations after pushes

### LINKED ISSUE #220: Suggestion for Chinese translations

detiam:
resolve #220


flxzt:
thank you. Can this be a separate workflow that only triggers when the `zh_hans` file actually changed through [this condition](https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#running-your-workflow-only-when-a-push-affects-specific-files). It would also keep the CI workflow neat and clean

detiam:
@flxzt ok updated

flxzt:
alright, I tested it [here](https://github.com/flxzt/rnote-testing/actions/runs/6835465493/job/18589286936) but it fails in the last commit/push step.

flxzt:
with that addition it seems to be working ( see [here](https://github.com/flxzt/rnote-testing/commit/9b78ceb9809f5f0bee7ef320c6df3934621dfa3b) ). However the workflow removes `zh_hans` from `LINGUAS which is probably not what we want because weblate will always add it again and there may be system configuration that support these meta locales?. It shouldn't be a problem if there are "zh_hans zh_hant" **and** the additional locales "zh_CN  zh_SG zh_HK zh_TW" in the LINGUAS file

flxzt:
alright, this looks good to me now and seems to be doing what we want (see [here](https://github.com/flxzt/rnote-testing/actions/runs/6913076254)! Thank you!

detiam:
@flxzt updated, my system ssd broken so it's been later, sorry @_@

BobH-Official:
I just recieved email from weblate today, and noticed this pr. Thank you @flxzt for adoptting my suggestion #220. And thank you @detiam for implementing the code.

PR #854, opened 2023-10-30
Disable Qt5 and Qt6 in the poppler build configuration

silvasch:
These features don't seam to be required, and can cause issues when building the project.


flxzt:
thanks! this should speed up a first flatpak build as well.

PR #856, opened 2023-10-30
Don't open a new tab when the default "New Document" hasn't been modified

silvasch:
Previously, when opening rnote and loading a document  by opening a file the "default" (empty) canvas would remain open and the opened file would appear in a new tab. This PR changes this so that the default document gets replaced by the opened document when no changes to the default canvas were made, which I think is better.


silvasch:
I accidentally made changes for a new feature on the same branch, that's why I had to force-push, sorry.

flxzt:
that is a nice usability improvement, thanks!

Issue #867, opened 2023-11-03
Ability to open files in the default app from the workspaces sidebar


silvasch:
**Please submit only one feature request in one issue!**  
I think it would be good to add a menu option in the worspaces sidebar to open files in the default system app.

**Is your feature request related to a problem? Please describe.**  
Sometimes I want to just view a PDF instead of inserting it into the drawing area. I have to open my file manager each time and navigate to the PDF.

**Describe the solution you'd like**  
Add a menu option for opening non-rnote files with the default system app in the workspace sidebar, right under "open".

![Screenshot from 2023-11-03 09-48-16](https://github.com/flxzt/rnote/assets/128063006/26e6893d-1895-4b5f-b140-06fb2718b987)

flxzt:
good idea. The cli already brings in a crate that can provide this in a similar feature

PR #868, opened 2023-11-03
Open files in default app from workspaces

silvasch:
As specified in #867

### DIFF
--- a/crates/rnote-ui/src/workspacebrowser/filerow/actions/mod.rs
+++ b/crates/rnote-ui/src/workspacebrowser/filerow/actions/mod.rs
@@ -1,11 +1,13 @@
 // Modules
 mod duplicate;
 mod open;
+mod open_in_default;
 mod rename;
 mod trash;
 
 // Re-exports
 pub(crate) use duplicate::duplicate;
 pub(crate) use open::open;
+pub(crate) use open_in_default::open_in_default;
 pub(crate) use rename::rename;
 pub(crate) use trash::trash;
--- a/crates/rnote-ui/src/workspacebrowser/filerow/actions/mod.rs
+++ b/crates/rnote-ui/src/workspacebrowser/filerow/actions/mod.rs
@@ -1,13 +1,13 @@
 // Modules
 mod duplicate;
 mod open;
-mod open_in_default;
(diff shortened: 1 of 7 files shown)


silvasch:
This is now implemented, however, I'm not sure about one thing: The new menu option also exists for '.rnote' files. This doesn't really make sense, as you're already in the default app for that file type (When pressing it it just opens the file in a new window). Should I try to dynamically remove that option or should I just leave it?

flxzt:
I am not completely sure either, but I tend toward having it for rnote files as well. There might be scenarios where the default app isn't even rnote and it is probably confusing to have a menu where options only appear depending on the file type.

crates/rnote-ui/src/workspacebrowser/filerow/actions/mod.rs
```diff
@@ -1,11 +1,13 @@
 // Modules
 mod duplicate;
 mod open;
+mod open_in_default;
```
flxzt, line 4: It would be more clear if the module, function and action names are renamed to `open_in_default_app`. It's not much longer anyway.

crates/rnote-ui/src/workspacebrowser/filerow/actions/open_in_default.rs
```diff
@@ -0,0 +1,21 @@
...
+use gettextrs::gettext;
+use gtk4::{gio, gio::prelude::FileExt, glib, glib::clone};
+
+/// Create a new `open-in-default` action.
+pub(crate) fn open_in_default(filerow: &RnFileRow, appwindow: &RnAppWindow) -> gio::SimpleAction {
+    let action_open_in_default = gio::SimpleAction::new("open-in-default", None);
+    action_open_in_default.connect_activate(
+        clone!(@weak filerow, @weak appwindow => move |_action_open_in_default, _| {
+            if let Some(current_file) = filerow.current_file() {
+                if let Err(_) =  open::that(current_file.uri()) {
```
flxzt, line 13: Log this error with `log::debug("opening file .. with default app failed, Err: {e:?}" as well here

flxzt:
lgtm. thank you!

PR #875, opened 2023-11-12
RFC: consistent initial folder for export file chooser dialogs

pobrn:
The main change is the second commit:

```
This commit sets the initial folder of the file chooser
for all three types of export options. The following
locations are tried:

  1. last export path of the current canvas
  2. parent directory of the current canvas' output file
  3. current directory of the file browser in the sidebar

The first available will be used.
```

I have found myself wanting to see better defaults especially when I'm exporting a selection multiple times, or multiple selections in succession. So here is my attempt at implementing something.

### DIFF
--- a/crates/rnote-ui/src/canvas/mod.rs
+++ b/crates/rnote-ui/src/canvas/mod.rs
@@ -78,6 +78,8 @@ mod imp {
         pub(crate) empty: Cell<bool>,
         pub(crate) touch_drawing: Cell<bool>,
         pub(crate) show_drawing_cursor: Cell<bool>,
+
+        pub(crate) last_export_dir: RefCell<Option<gio::File>>,
     }
 
     impl Default for RnCanvas {
@@ -172,6 +174,8 @@ mod imp {
                 empty: Cell::new(true),
                 touch_drawing: Cell::new(false),
                 show_drawing_cursor: Cell::new(false),
+
+                last_export_dir: RefCell::new(None),
             }
         }
     }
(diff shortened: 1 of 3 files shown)


flxzt:
Thanks for the PR! I think this is a good change, it probably matches better what most users would expect.

crates/rnote-ui/src/canvas/mod.rs
```diff
@@ -705,6 +709,14 @@ impl RnCanvas {
         );
     }
 
+    pub(crate) fn get_last_export_dir(&self) -> Option<gio::File> {
```
flxzt, line 712: Typically the `get_` is emitted for getters in Rust

crates/rnote-ui/src/dialogs/export.rs
```diff
@@ -836,9 +824,22 @@ pub(crate) async fn dialog_export_selection_w_prefs(appwindow: &RnAppWindow, can
     }
 }
 
+fn get_initial_folder_for_export(appwindow: &RnAppWindow, canvas: &RnCanvas) -> Option<gio::File> {
```
flxzt, line 827: Add a function doc comment to explain what determines the initial folder

flxzt:
I'd change the behaviour for the default folder for the `filechooser_export_engine_state()` and `filechooser_export-engine_config()` functions as well, even if they are only used by developers

flxzt:
Looks good to me, thanks! I also like that the export location is remembered per canvas/tab which should be the most useful case most of the time.

crates/rnote-ui/src/dialogs/export.rs
flxzt: yes

flxzt:
I also only noticed it after merging. Oh well, not that critical

Issue #927, opened 2023-12-11
Add two more color slots


silvasch:
**Is your feature request related to a problem? Please describe.**  
There is an inconvinient number of color slots in rnote. The color picker contains nine rows of colors, but rnote only allows eight of them in the slots. Additionally, because of how filling shapes in rnote works, it's very convenient to have a transparent color as well.

**Describe the solution you'd like**  
Increase the number of color slots by two, for: blue, green, yellow, orange, red, purple, brown, white, black and transparent.

silvasch:
I'm not sure if I understand correctly, it's okay to add one?

Btw, this is not my first issue :), I'm the guy that added features in the last three updates. I just changed my profile pic.

flxzt:
I know :) Yeah I think adding a single one is fine, but more than that would make the minimum width too large. Alternatively we could look into using a FlowBox that splits the color buttons into two rows, but when I tried that I couldn't get it work properly in combination with the overlay container

silvasch:
> I know
I thought you didn't recognize me because you added the `good first issue` label.

> Yeah I think adding a single one is fine
Perfect, I actually use purple and brown daily, that's why I would like to. But I understand your concerns about the minimum width, and adding one slot already improves it for me a lot.

I'll try to implement it in tomorrow, but I might not have time.

silvasch:
I just thought of another solution; we could add a setting to define how many color slots you want. It could be a limited range (4-10?) to make sure that every setting works nicely.

flxzt:
Not a huge fan of that, feature creep like that can make the app too complicated in the long-term

silvasch:
Yeah that's true... I think it's better if we just leave it at what it's at now.

flxzt:
No single additional color? A grey or white would be useful for most, and that would be fine with me. :)

silvasch:
Ok, it just seemed to me that you're not sure about adding a slot at all.

silvasch:
I've added one color slot in the PR #936.
However, I'm not sure how I should set it's default color to white (because the `default_color` function does some magic calculations).

I could just leave it at the color calculated by this function, which is a blue between the two other blue tones, or I could rewrite the function to hardcode the default colors.

Personally I would hardcode the colors and choose the colors from the default gnome color palette:
![image](https://github.com/flxzt/rnote/assets/128063006/aae52021-6a7a-4064-a186-64ca98255a84)

My proposal would be

1 - Black
2 - Blue
3 - Green
[...]

PR #936, opened 2023-12-15
Additional color slot

silvasch:
This PR implements #927.


flxzt:
The code looks good, but the minimum window width needs to be increased to let's say 530 [here](https://github.com/flxzt/rnote/blob/cda23dcbef194e25d871ae9babdfddf577cc44fb/crates/rnote-ui/data/ui/appwindow.ui#L4) to avoid this bug:

Bildschirmfoto vom 2023-12-15 17-23-10

flxzt:
thanks! I hope this is still satisfactory for you, with the compromises that were made

silvasch:
No worries, I understand your concerns about adding too much.

PR #940, opened 2023-12-18
Fix wrong appmenu & canvasmenu position when open by shortcut

detiam:


flxzt:
@bertob is this good UX/ does this follow Gnome HIG, that the primary and secondary menu appear in an overlaid sidebar? The alternative could be: the keyboard shortcuts close the overlay in addition to showing the popovers from the menus in the "main" window

bertob:
No, this kind of behavior is not recommended, but I think it's basically trying to work around weirdness with the existing layout. I don't have a concrete solution off-hand, but if you want to bring the layout in line with the HIG the sidebar would probably need to be restructured. For example, maybe it could be a utility pane with a fixed header bar? Maybe all of the sidebar stuff could be moved elsewhere? Maybe the file manager is not needed at all, and the settings could just be a preferences dialog?

flxzt:
Ok thank you. But such a rework is another discussion. I think there is value in having the settings in the sidebar and having a workspace browser, even if it's not strictly in line with the other apps/the guidelines.

But to resolve this specific issue, maybe you could refactor the PR so that the shortcut closes the sidebar instead of showing the "wrong" menus (the current behaviour) @detiam ?

detiam:
@flxzt done! the reason I add appmenu on sidebar is based on commit message [here](https://github.com/flxzt/rnote/commit/48fe4c0e3ff80355056af8fa44d021e14e712eef) though

flxzt:
Sorry, that was unclear by me: I was only referring to the sidebar when it is an overlay.
So essentially:

Keep the menus as they are currently if the sidebar is not an overlay. Which means displaying the app-menu button in the sidebar and hiding the one in the main window.
But fix the menu locations when they are activated through the keyboard shortcut. Which can be done by checking if the sidebar is shown when the respective action is triggered and then popping up the correct app-menu & canvas-menu.

And if the sidebar is an overlay, close it before popping up the menus in the aciton (you implemented that already)

detiam:
> Sorry, that was unclear by me: I was only referring to the sidebar when it is an overlay. So essentially:
> 
> Keep the menus as they are currently if the sidebar is not an overlay. Which means displaying the app-menu button in the sidebar and hiding the one in the main window. But fix the menu locations when they are activated through the keyboard shortcut. Which can be done by checking if the sidebar is shown when the respective action is triggered and then popping up the correct app-menu & canvas-menu.
> 
> And if the sidebar is an overlay, close it before popping up the menus in the aciton (you implemented that already)

oh then this PR should be very simple, done

flxzt:
The condition that the sidebar is only closed when it is an overlay needs to be there for the canvas menu as well

flxzt:
I quickly fixed that and resolved the conflict (the file got renamed)

PR #943, opened 2023-12-21
chore: better screenshots

### LINKED ISSUE #933: Better appdata screenshots
The current appdata screenshots are not quite in line with the new Flathub [metadata quality guidelines](https://docs.flathub.org/docs/for-app-authors/appdata-guidelines/quality-guidelines), in particular:

- They're too tall (should be 1000x700 pixels or smaller)
- Window controls are non-standard (should be close-only)

This is becoming more important since we want to use app screenshots in featured banners and the like in the future :)

flxzt:
@bertob See: https://github.com/flxzt/rnote/tree/better-screenshots#screenshots , can you double-check if these adhere to the new guidelines?

fixes #933 when merged


bertob:
Yes, with one exception: In the first one the window seems to not have focus.

flxzt:
ok that's fixed now. Thanks!

Issue #938, opened 2023-12-16
Revise settings


silvasch:
**Is your feature request related to a problem? Please describe.**  
Currently rnote has a settings panel in the sidebar. This is kind of special as other applications usually have a dedicated settings window implemented using the `AdwPreferencesWindow` widget. Having a separate window has many advantages:
1. The preferences window has multiple pages that make it easier to find settings, as you can see a rough overview at a quick glance.
2. The `AdwPreferencesWindow` automatically implements a search function to quickly find settings.
3. The `Button Shortcuts` sections could be on a separate page and can then be further divided in `Drawing Pad Shortcuts`, `Stylus Shortcuts` and `Miscellaneous`.

**Describe the solution you'd like**  
Instead of the settings panel in the sidebar, implement a separate window using the `AdwPreferencesWindow`. This widget has `AdwPreferencesPage`'s as children, which are themselves divided into `AdwPreferencesGroup`'s. The settings would then be reordered as follows:

- General
  - UI
    - Show scrollbars
    - Inertial touch scrolling
    - Regular cursor
    - Enable drawing cursor
    - Drawing cursor
  - Auto Saving
    - Enable auto saving
    - Auto saving interval
- Canvas
  - Page Format
    - Format
    - Orientation
    - Width
    - Height
    - DPI
  - Document
    - Format border color
    - Color
    - Background pattern
    - Pattern
    - Pattern color
    - Pattern width
    - Pattern height
- Shortcuts
  - Stylus shortcuts
  - Miscellaneous
  - Drawing pad shortcuts

**Additional context**  
I've quickly implemented how the new settings window would look:
![image](https://github.com/flxzt/rnote/assets/128063006/c8abc326-0c8c-431b-90b7-a086563b2e35)

What do you think? Why did you decide to not use the standard `AdwPreferencesPage` from the beginning?

flxzt:
The main reason I impemented it as a sidebar is because I want to give users a live preview when document settings are changed. If it were a (modal) window, it would be cumbersome to make changes by: changing value -> closing window -> possibly reopen window to adjust and so on

silvasch:
> The main reason I implemented it as a sidebar is because I want to give users a live preview when document settings are changed.

That makes sense. So maybe the settings could be split into two parts: The "Page format" and "Document" settings stay in the sidebar, and the tab would be called "Canvas settings". All other settings don't require a live preview and could be in the new, separate window called "Preferences".

[...]

PR #954, opened 2023-12-30
Move `Document Layout` and `Developer` preferences

silvasch:
This PR moves the `Document Layout` and the `Developer` preferences to the settings panel.

# Status
- [x] Document layout
  - [x] Add UI
  - [x] Implement behavior (inconsistencies when loading documents)
  - [x] Remove previous UI
- [x] Developer options
  - [x] Visual debugging
    - [x] Add UI
    - [x] Implement behavior
    - [x] Remove previous UI
  - [x] Export engine state
    - [x] Add UI
    - [x] Implement behavior
    - [x] Remove previous UI
  - [x] Export engine config
    - [x] Add UI
    - [x] Implement behavior
    - [x] Remove previous UI


flxzt:
Thanks. Is it intentional that you didn't move the shortcuts section to its own page as discussed or is it outside of the scope of this PR for you?

The document-layout option looks good.

The developer options are now seen by normal users as well, which is not desirable. Also: it was an intentional choice not to restore the visual-debug option.
But to be really honest I still think the old menu was a better fit as a dev-menu where actions can be added quickly for testing purposes, why are you so eager to change it?

silvasch:
This PR implements everything how I mentioned it in [this comment](https://github.com/flxzt/rnote/issues/938#issuecomment-1870551113). I thought this was how we ended up.

> Is it intentional that you didn't move the shortcuts section to its own page as discussed or is it outside of the scope of this PR for you?

As mentioned above, this PR just implements everything mentioned in the comment.

> The developer options are now seen by normal users as well, which is not desirable.

I know what you mean, but I'd argue that the current way is visible for normal users as well. If you want to hide the developer options, I would hide the options completely and make them visible using a keyboard shortcut (either in the app menu or in the settings page).

> Also: it was an intentional choice not to restore the visual-debug option.

I thought visual debugging was already being restored, but turns out I never really tested it.

> But to be really honest I still think the old menu was a better fit as a dev-menu where actions can be added quickly for testing purposes, why are you so eager to change it?

I'd like to change it because I think they are in the wrong place. Having a developer menu is something special that rnote has (at least I've never seen developer options in any other app). The only place I can see them fit are as settings in the settings menu.

Here's what I'm gonna do:
1. I'm going to change this PR to only move the document layout options.
2. I'm going to create a second PR to move the shortcuts to a new panel.
3. I'm going to leave the developer menu as is right now and think about a better solution.

flxzt:
Sorry for the push, I accidentally did a chore on the wrong branch.


> Here's what I'm gonna do:

>    I'm going to change this PR to only move the document layout options.
>    I'm going to create a second PR to move the shortcuts to a new panel.
>    I'm going to leave the developer menu as is right now and think about a better solution.

Okay, that sounds good. That's what we can both agree on

flxzt:
can I refactor  this to completion @silvasch ? I'd like to do a release soon

silvasch:
Yes, you can. I have a lot to do at the moment, sorry.

flxzt:
thanks, and no worries

flxzt:
see #1006

Issue #961, opened 2024-01-07
Change keyboard shortcuts for `new document`


silvasch:
**Is your feature request related to a problem? Please describe.**  
The current shortcuts for window and tab-management are very different to how official gnome apps work. The current shortcuts are:
- `Ctrl-n`: Create new document, replaces currently open document
- No shortcuts for creating a new window
- No shortcuts for creating a new tab

**Describe the solution you'd like**  
Change them to how they are in other apps:
- `Ctrl-n`: Create a new rnote window
- `Ctrl-t`: Create a new document *in a new tab*

flxzt:
Agreed! The old one is from before tabs were implemented. The [Gnome HIG](https://developer.gnome.org/hig/reference/keyboard.html#basic-shortcuts) doesn't mention ctrl+t for new tab, but it seems to be convention among other apps so sounds good to me.

PR #964, opened 2024-01-09
New keyboard shortcuts for managing windows and tabs

silvasch:
Implements #961

### DIFF
--- a/crates/rnote-ui/src/app/mod.rs
+++ b/crates/rnote-ui/src/app/mod.rs
@@ -128,7 +128,7 @@ mod imp {
         }
 
         /// Initializes and shows a new app window
-        pub fn new_appwindow_init_show(&self, input_file: Option<gio::File>) {
+        pub(crate) fn new_appwindow_init_show(&self, input_file: Option<gio::File>) {
             let appwindow = RnAppWindow::new(self.obj().upcast_ref::<gtk4::Application>());
             appwindow.init();
             appwindow.present();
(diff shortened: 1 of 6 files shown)


crates/rnote-ui/src/appwindow/actions.rs
```diff
@@ -567,13 +565,6 @@ impl RnAppWindow {
             appwindow.handle_widget_flags(widget_flags, &canvas);
         }));
 
-        // New doc
```
flxzt, line 570: This action is still needed for the `New` menu item in the app-menu.

crates/rnote-ui/src/app/mod.rs
```diff
@@ -128,7 +128,7 @@ mod imp {
         }
 
         /// Initializes and shows a new app window
-        fn new_appwindow_init_show(&self, input_file: Option<gio::File>) {
+        pub fn new_appwindow_init_show(&self, input_file: Option<gio::File>) {
```
flxzt, line 131: use `pub(crate)` here

crates/rnote-ui/src/appwindow/actions.rs
silvasch: True. But I don't think it should use the old `new-doc` action, but should also create a new tab. I don't think replacing your current tab is the desired behaviour.

flxzt: Let's keep the action anyway, in case it is needed again at some point in the future.

flxzt:
looks good now. thanks

Issue #1000, opened 2024-02-03
Failed to build from source with meson


A-Walrus:
**Describe the bug**  
After building `rnote` from source with meson running it results in this error message and nothing else:

```
(rnote:27743): GLib-GIO-ERROR **: 19:13:32.718: Settings schema 'com.github.flxzt.rnote' does not contain a key named 'colorpicker-color-9'
```

**To Reproduce**  
1.  git clone the repo (commit `15312eff7b8e2987dc5f25423f5923c29dd470f9`)
2. Follow meson instructions
  - `meson setup --prefix=/usr _mesonbuild`
  - `meson compile -C _mesonbuild`
  - run the executable with `./_mesonbuild/target/release/rnote`

**Expected behavior**  
`rnote` opens correctly

**Desktop (please complete the following information):**  
 - OS: EndeavorOS
 - App Version: commit `15312eff7b8e2987dc5f25423f5923c29dd470f9`
 - Desktop Environment: Gnome 45.3
 - Display Server: Wayland

flxzt:
You need to install the app to the system with the latest gschema (a new colorpicker palette field got added relatively recently)

A-Walrus:
Thanks, trying to figure this out. Is the package you're referring to `gsettings-desktop-schemas`?
If so the latest stable release seems to be `45.0` which is what I have ([releases](https://gitlab.gnome.org/GNOME/gsettings-desktop-schemas/-/tags)).

Or am I looking at the completely wrong package :P
Thanks!

A-Walrus:
Thanks that solved it! I I think I'll add a quick note to the `BUILDING.md`  file to save the next person the trouble :)

flxzt:
That sounds good to me!

PR #1001, opened 2024-02-03
Add note about old gschema in BUILDING.md

A-Walrus:
Quick note about fixing older `gschemas`, see #1000.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -138,8 +138,13 @@ meson compile -C _mesonbuild
 
 The compiled binary should now be here: `./_mesonbuild/target/release/rnote`.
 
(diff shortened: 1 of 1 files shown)


BUILDING.md
```diff
@@ -138,6 +138,9 @@ meson compile -C _mesonbuild
 
 The compiled binary should now be here: `./_mesonbuild/target/release/rnote`.
 
+Note that if an older version of rnote has previously been installed, the old `gschema` file will still be used, which
```
flxzt, line 141: I'd add  a short explanation about what the gschema file is/does, mentioning that the issues can arise when it has been modified after when the build was last  installed. And explicitely mention the error you encountered, so it is easier to make the connection between the error and this note

flxzt:
thanks!

PR #1002, opened 2024-02-05
Mac os fullscreen temporary fix

### LINKED ISSUE #823: Window does not resize on mac and has no decoration
**Describe the bug**  

**To Reproduce**  

**Expected behavior**  

**Console Output**  
Run the app in the terminal with `flatpak --env=RUST_LOG=rnote=debug --env=RUST_BACKTRACE=1 run com.github.flxzt.rnote` and post the output.

**Screenshots**  

 - OS: [e.g. Fedora 38,  Windows 11 22H2, ...]
 - App Version: [e.g. Rnote v0.9.0]
 - Installation Source: [e.g. Flatpak, Archlinux Community Repo, ...]
 - Desktop Environment: [e.g. Gnome 44.2]

**Additional context**

Doublonmousse:
Tries to fix #823 by deferring a call to `maximize` after the window is presented.

In the current implementation, the whole logic for setting the default window size/height and maximize is separated into its own function, with conditional compilation flag to call it either before presenting the window on non mac os platform, or after for mac os.


crates/rnote-ui/src/app/mod.rs
```diff
@@ -128,8 +128,15 @@ mod imp {
...
         pub(crate) fn new_appwindow_init_show(&self, input_file: Option<gio::File>) {
             let appwindow = RnAppWindow::new(self.obj().upcast_ref::<gtk4::Application>());
+
             appwindow.init();
             appwindow.present();
+            // restore the window dimension and maximize after presenting it on mac os
+            // see issue 823 - https://github.com/flxzt/rnote/issues/823
+            #[cfg(target_os = "macos")]
+            if let Err(e) = appwindow.load_window_settings() {
+                tracing::error!("Failed to restore windows settings, Err: {e:}");
```
flxzt, line 138: The message should be `...{e:?}` so that the full error is printed verbosely

Doublonmousse: I copied more or less what was in `crates/rnote-ui/src/appwindow/mod.rs` so there are a couple of `...{e}` that could be changed to `...{e:?}` there, should I go and change everything to `...{e:?}` there ?

Doublonmousse:
I am not sure if there is any downside to doing in in that order on other os, I kinda tested it mainly on mac, and left it more or less the same for the other os, just in case

crates/rnote-ui/src/app/mod.rs
flxzt: yes that would be nice!

Doublonmousse:
Seeing how weird things can happen when changing order, I kinda want everything to remain the same on non mac-os platform.

The only thing that may happen (and even then I'm not sure I can differentiate this from the regular way gtk works on mac) is a slight offset between the mouse and the window titlebar when going from the `maximize` status to the windowed state on mac but this resolve on the second try, that may be bigger than without the change.

(I haven't tested if changing the order would create similar effects on  other platforms or not).

I'm compiling using the `rnote-macos-build.md` readme, and I'm not using dehesselle custom setup for now, so maybe double checking with a mac build before merging might be of order to be sure.

flxzt:
ok, that's a valid point - don't change what ain't broken.

code looks good, but then I'll hold back from merging until a new macos build from @dehesselle confirms the fix

dehesselle:
A pipeline with test builds is available here: https://gitlab.com/dehesselle/rnote_macos/-/pipelines/1166889790/builds

Doublonmousse:
Oh well that's not good ... The window is draggable but is stuck again in this "I am maximized and floating" state.

One difference between this build and what I compiled is that I did not manage to make the `gschema` work on mac using the final instruction `rnote-macos-build.md`, so none of `setup_settings_binds()`, `setup_periodic_save()`, `load_settings()` would have run before presenting the window (and I mocked the window settings in `load_window_settings()` manually instead). 

Though I'm not 100 % sure if that's what's causing the difference in behavior

dehesselle:
I've started a new pipeline. https://gitlab.com/dehesselle/rnote_macos/-/pipelines/1167617808

(It's not an effort, I can do this as often as necessary.)

Doublonmousse:
Ah still no. I'm going to try to remove the three calls to the `gsettings` bind temporarily to see if I can even reproduce the "non-buggy" strate

dehesselle:
New pipeline: https://gitlab.com/dehesselle/rnote_macos/-/pipelines/1167831675

Doublonmousse:
I'm on gtk4 `4.12.15` on mac and you're building on `4.12.4`, maybe that's where the difference is (partial bug fix) ?

One last (dirty) try with me mocking gsettings to be sure that's not where the difference lies (between my compiled version with gsettings not enabled and yours)

dehesselle:
Is that a typo? gtk4's latest stable release is `4.12.5`. I can update that so we can be sure.

dehesselle:
New pipeline with your latests changes and also gtk4 4.12.5: https://gitlab.com/dehesselle/rnote_macos/-/pipelines/1168185219/builds

dehesselle:
https://gitlab.com/dehesselle/rnote_macos/-/pipelines/1168267279

Doublonmousse:
There was one last call that did not happen on my local version because of a `gsetting` call that failed.

Just say it if it's too much. Thank you very much for all the build !

Normally this should be one-to-one with how my local version works (hopefully).
If this still doesn't work I will stop there and close the PR

Doublonmousse:
And of course I pushed a wrong version that did not run because I commented a little too aggressively ...

dehesselle:
https://gitlab.com/dehesselle/rnote_macos/-/pipelines/1168537588/

dehesselle:
https://gitlab.com/dehesselle/rnote_macos/-/pipelines/1169000111/

Doublonmousse:
Does the variable `window_rect` in this file sets the default size for the `.plist/gsetting` of the window ?
https://gitlab.com/dehesselle/rnote_macos/-/blame/master/resources/rnote_dmg.py?ref_type=heads#L101

I have weird offset effects depending on the versions but the maximize button always do something even though it's always incorrect (in the last version, maximizing seems to cycle between the default size I set and the size from `window_rect`, in another earlier build it would have the correct size but would placed incorrectly)

Apart from the `gsettings` part (that is now fully commented out) I wonder if this is where the difference occurs between the `meson` build and the packaged version 

Did the original bug prevent the window from being dragged around ? I can always drag the windows around on any on my (more or less) broken builds

dehesselle:
The `rnote_dmg.py` file is only used to build the dmg, it is independent from Rnote itself. It has parameters that say "when you open the dmg, where should Finder show the window". To be absolutely clear, I'm talking about this:
image

When you say `.plist/gsetting`, you are talking about `~/Library/Preferences/net.flxzt.Rnote.plist`, right? That is completely managed by GTK, my build/packaging does not interfere with that.

Do we need to compare more library versions? When you compare "meson build" and "packaged version", how are you building?

Doublonmousse:
When I say `meson build` I'm talking about the guide in https://github.com/flxzt/rnote/blob/main/misc/building/rnote-macos-build.md expect the last line (can't seem to make the `GSETTINGS_SCHEMA_DIR` export command work for rnote to find `gschema`).

For the packaged version, I'm talking about your build.

I thought that I ought to try to disable calls to the setting schema parameters to reduce the difference between the two build as much as possible but that is done now. 

I'm kinda wondering at this point if not having the `gsetting/plist` file vs having it creates the bug (although even without order matters as well) and I kinda "lucked out" on this.

I'm not sure there's any other fundamental difference between the two builds that would be responsible for the bug.

Doublonmousse:
`setup_periodic_save` and `save_to_settings` are the two last function present that have a relation to the setting schema, maybe I can try to comment that as well

Doublonmousse:
Do you want me to double check the versions of appstream, gtk4, poppler, desktop-file-utils and libadwaita to see if there is any difference ?

dehesselle:
Hm... thinking about it, window behavior and interfacing with macOS on such a low-level is GTK stuff. I think we can rule out other dependencies.

I've re-read the conversation in the issue. I was not able to reproduce the problem on my end. I can try that again, there are instructions on how a user was able to reproduce the original issue.

One other thing: you said you have problems with `GSETTINGS_SCHEMA_DIR`. I do not set that as well, what I do instead is setting `XDG_DATA_DIRS` as glib does support that to lookup its schema files.

New pipeline is here: https://gitlab.com/dehesselle/rnote_macos/-/pipelines/1169967724/

Doublonmousse:
Yeah it seems likely this is gtk specific.
Plus it's also very likely apple arm + recent os version specific. 

Maybe a last test would be to temporarily remove the setting `XDG_DATA_DIRS` (cause I don't see anything else that could explain the difference).
I need to re enable a seting before you launching a new build though for it

Doublonmousse:
The builds you made should make reproducing the issue easier though. 

Build 139 (the one just after updating to gtk 4.12.5) in particular opens as a maximized window and on my side it's draggable but not resizable and toggling maximize does not work (or does incorrect things)

dehesselle:
https://gitlab.com/dehesselle/rnote_macos/-/pipelines/1170014553

Doublonmousse:
I finally managed to install the `gschema` and for some reason I don't have any issue with the fix + the gschema installed. What I did was take the `com.github.flxzt.rnote.Devel.gschema.xml` file from the folder `/usr/local/share/glib-2.0/schemas` and put it in another folder, ran `glib-compile-schemas .` in that folder and ran rnote with the command

```
RUST_LOG=rnote=debug GSETTINGS_SCHEMA_DIR=(path to the folder with the xml file) rnote
```

I found out that `/usr/local/share/glib-2.0/schemas` is probably not the correct folder for gschema on arm, all of the schema files (and the `gschema.compiled` file) are in `/opt/homebrew/share/glib-2.0/schemas` (glib installed from the arm `brew` command). So I'm not sure why the bug appear on the packaged arm version but not on my side with the gsettings activated, though it seems to still be the same bug (maybe a good test would be to compile on that branch to check if the bug disappear). 

Compiling gtk and libadwaita from source with the branch from https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/6882, I can confirm that with the `gsettings` enabled, I don't have to invert the `maximise` and `present` commands, although there is a regression on that version https://gitlab.gnome.org/GNOME/gtk/-/issues/6436

Interestingly, even compiling from source, `gtk` installs into `/opt/homebrew/include/gtk-4.0`.

So this bug will probably be closed on the next `4.13.x` release (or on the nightlies soon) and we'll be able to revisit things once gtk `4.14` launches.  

As I did not manage to make the temporary fix work on the packaged arm version and the fix is incoming on the next gtk version, I'm closing this

PR #1003, opened 2024-02-06
Add `Ctrl+A` to select all strokes and re enable `Ctrl+T`

Doublonmousse:
- `Ctrl+A` to select all strokes
- `Ctrl+T` re enabled (added back to the `shortcuts.ui`, wouldn't work otherwise)

### DIFF
--- a/crates/rnote-ui/data/ui/shortcuts.ui
+++ b/crates/rnote-ui/data/ui/shortcuts.ui
@@ -233,6 +233,12 @@
                 </child>
                 <child>
                   <object class="GtkShortcutsShortcut">
+                    <property name="title" translatable="yes">Select All Strokes</property>
+                    <property name="accelerator">&lt;ctrl&gt;a</property>
+                  </object>
+                </child>
+                <child>
+                  <object class="GtkShortcutsShortcut">
                     <property name="title" translatable="yes">Duplicate Selection</property>
                     <property name="accelerator">&lt;ctrl&gt;d</property>
                   </object>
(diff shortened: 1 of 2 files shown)


Doublonmousse:
I'm a little confused, without this PR,
- `Ctrl+A` only works when opening the selection toolbar then clicking back on the canvas
- Same thing for `Delete`

I'm not sure this is intended. Should I add `Delete` as well as a global keyboard shortcut in the same way ? (It's also not listed in the keyboard shortcut page)

flxzt:
yeah, the problem is that the keyboard focus moves out of the canvas as soon as the user clicks somewhere else. I haven't found a good solution to that

Doublonmousse:
Yeah there isn't much to get from this PR as it doesn't seem to fix anything, that's why I closed it.

Maybe adding back `Ctrl+T` in the keyboard shortcut page would be okay ? Seems to play on whether or not the keyboard shortcut activates or not (?)

But it's also that by looking at #955 even though `Ctrl+A` `Ctrl+T` are implemented, the way they're implemented can make people think that the shortcuts don't exist.

flxzt:
Ctrl + T is already in the shortcuts window though and is a global shortcut?

I think they want Ctrl+Shift+T to reopen the last closed tab

grafik

Maybe these globally activatable, but pen-dependent shortcuts can be implemented differently - like adding a keyboard event controller to the appwindow and pass the pressed keys along with the modifiers into the engine.

Issue #974, opened 2024-01-11
Resize image content to fit page-width while pasting from clipboard


claws99:
I'm a teacher, I use lot of images from Internet (which are obviously of different resolutions). Copy-Pasting them is really cumbersome,  as the image goes out of boundaries of the page.

Solution would be to scale the image to say 30-40% of the page in visibility without compromising on resolution or aspect ratio.

flxzt:
That should be easy to implement, and a nice QoL improvement. But I'd say the original size should be used as long as it is smaller than the current viewport, and capped to the viewport when it is larger?

MisterE123:
> That should be easy to implement, and a nice QoL improvement. But I'd say the original size should be used as long as it is smaller than the current viewport, and capped to the viewport when it is larger?

That sounds reasonable

claws99:
> But I'd say the original size should be used as long as it is smaller than the current viewport, and capped to the viewport when it is larger?

agreed!

PR #1004, opened 2024-02-06
refined imported image size logic

Doublonmousse:
- for fixed layout : smallest ratio such that the image fits in the viewport both horizontally and vertically, and doesn't go out of bounds of the page on the right
- for infinite layout : smallest ratio such that the image fits in the viewport both horizontally and vertically

Maybe the logic can be tweaked a little. For the two infinite layouts I wonder if some additional logic could be added to change how images are put when the format borders are displayed ..

To fix : #974

### DIFF
--- a/crates/rnote-ui/data/ui/canvasmenu.ui
+++ b/crates/rnote-ui/data/ui/canvasmenu.ui
@@ -61,7 +61,7 @@
             <attribute name="action">win.block-pinch-zoom</attribute>
           </item>
           <item>
-            <attribute name="label" translatable="yes">Respect Borders when Pasting</attribute>
+            <attribute name="label" translatable="yes">Respect Borders When Pasting</attribute>
             <attribute name="toggle"/>
             <attribute name="action">win.respect-borders</attribute>
           </item>


Doublonmousse:
The svg part is not correct yet

Doublonmousse:
All fixed

crates/rnote-engine/src/strokes/vectorimage.rs
```diff
@@ -156,21 +160,37 @@ impl VectorImage {
...
 
+        // ratio to resize the image if needed
+        let resize_ratio = match resize {
+            None => 1.0,
+            Some(resize_struct) => calculate_resize(
+                resize_struct,
+                na::Vector2::new(svg_tree.size.width() as f64, svg_tree.size.height() as f64),
+                pos,
+            ),
+        };
```
flxzt, line 171: I think this calculation is better placed in the functions `generate_vector_image_from_bytes()`. Otherwise the parameters get a bit unclear - there's size, but it won't be the actual size if `resize` is also passed in. And then we don't need the `Resize` struct, we can pass in the arguments the calculation depends on with descriptive names directly into `calculate_size()`

Doublonmousse: I know it's a little weird.

I did it that way because I need to know the size of the svg and the image before calculating the ratio to make the image smaller. But that size only gets calculated in the `from_svg_str` and `from_image_bytes` respectively, after the string/bytes are read/parsed.

If I want to change this calculation to be done one step before in the `import.rs` file, I need to move some of the logic in these functions, so that I either read/parse the data from the bytes there, calculate the ratio, and pass on the calculated size whilst reverting the changes on the `from_.._bytes` function (but then we effectively parse the images twice, because there are other calls to the functions that will push `str` and `bytes` to them directly,  hence I can't modify then to send as an argument the `Image` or `Tree` types directly without adding other modifications elsewhere).

I agree that for now the `Resize` + `size` combo doesn't make sense. 
I wonder if I could remove the `resize` variable and change `size` for an enum of type

```rs
enum size_or_resize { //terrible name, do you have a better idea ?
     Fixed(na::Vector2<f64>) // (1)
     Resized(Resize)
}
```
as all the calls that are resized are calls where the `size` would be `None` and that is also true in the other direction. That way the `Resize` struct has a clear meaning and doesn't class with the `size` parameter in any way

(1) : I can remove here the Option so that we either Resize to fit of give a fixed size, or leave it to be able to call with `Fixed(None)` and have the image be pasted like it was before, with a 1-1 pixel mapping.

crates/rnote-engine/src/engine/import.rs
```diff
@@ -392,3 +420,12 @@ impl Engine {
         widget_flags
     }
 }
+
+/// checks if the layout is constrained in the horizontal direction
+fn is_fixed_layout(layout_type: &Layout) -> bool {
```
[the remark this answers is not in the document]
Doublonmousse: fixed

Doublonmousse:
I'm thinking of changing the `resize` logic for the infinite layout when the format borders are visible. To me, this signifies the users want to "respect" these borders as much as possible. So that preventing images to span across pages might be wanted. The logic would be 

- Try to force the image not to go over a page border that is on the right on the position the image is pasted.
  - If that results in an image size that is very small compared to the canvas view, nudge the image to the next page on the right and re apply the same logic 
  - If not stop and paste the image

flxzt:
That would be clever! We just have to make sure this logic is obvious to the user, otherwise the size might seem random and unpredictable.

crates/rnote-engine/src/strokes/vectorimage.rs
Doublonmousse: The position dependence is mostly there for the fixed vertical layout 

Basically I want it to prevent this : 
image

Still technically in the viewport (for the size) but slightly annoying to resize without changing mode, resizing then going back to a fixed layout.

I might be wrong (going to double check if I'm mistaken about this happening without the additional `layout_fixed` thing

Doublonmousse: For the other modes, position doesn't really matter if the size is roughly similar to the viewport's size

Doublonmousse: I've fixed the main thing with the `enum` and that cleans up code in the other session by quite a bit.

For positions, do you want me to change 
```rs
            (resize.max_viewpoint.x - pos_left_top_canvas.index(i)) / (current_width)
```
to
```rs
            (resize.max_viewpoint.x) / (current_width)
```

I kinda see the `position` thing as a kinda neat positionning thing. For relatively large images, just zoom to the image size you want and there you go it's going to be the size you want.  On the flipside, it won't do anything about smaller images.

For the `nudging` part, I kinda don't want to do it in this PR, because the idea is just a little too rough.
- If the canvas is really zoomed out, there's no reason to take these format borders into consideration, it's going to be way too small so there's another thing to check
- the threshold thing might be annoying, better to have it respect borders completly or not
- To make it clear to the user you'd have to add either toast messages the first time explaining it, add some graphic help somewhere to explain, and it would probably need a dedicated setting not to annoy people.

Maybe that's better to reserve this to a contextual or settings option somwhere :  "paste whilst respecting borders".

Doublonmousse:
I've addded a different paste command to respect borders (for all document layout) called with `Ctrl + shift + a`. Ideally that could also be triggered whilst doing a drag and drop operation (maybe `shift` + drag and drop of an image if this is done with the mouse and maybe a pen button if the image is dragged from a pen ?) but I don't know how to do that.

Also this only applies to image, not to strokes, text or elements from rnote's clipboard. Should any scaling be done on that (even only partially, when actually selecting borders to be respected ?)

So copying an image from a rnote document to another will not apply any scaling, even forcing borders to be respected. 
This applies to svg images as well as bitmap that are imported then copied in rnote's clipboard as I see 
```
Recognized clipboard content format: application/rnote-stroke-content
```
in this case (and the same pathway for images is not re executed)

Testing these things, I might have encountered an `.jpg` image that can't then be copied in rnote, failing with `Reading clipboard text failed while pasting clipboard as image/png, Err: Error { domain: gdk-pixbuf-error-quark, code: 3, message: "Unrecognized image file format" }`. I wonder if this is a extension/image format mismatch that makes these things occur though...

crates/rnote-ui/src/appwindow/actions.rs
```diff
@@ -782,19 +813,20 @@ impl RnAppWindow {
 
         // shortcuts for devel build
         if config::PROFILE.to_lowercase().as_str() == "devel" {
-            app.set_accels_for_action("win.visual-debug", &["<Ctrl><Shift>v"]);
+            app.set_accels_for_action("win.visual-debug", &["<Ctrl><Alt>v"]); //clashes with the special paste method
```
Doublonmousse, line 816: I would like to use `<Ctrl><Shift>v` for the second paste method that respect borders, should I leave this shortcut intact and change the one for the paste method or change this one (not sure this is the best keyboard shortcut to change the visual debug call to).

flxzt: It's fine to me to change the visual-debug shortcut. To be honest it can be any shortcut, it's just sometimes convenient to toggle it via keyboard for quick debugging. But not important which actual keys need to be pressed.

Doublonmousse: Ok then.
I think this PR is done, I want to do the other resizing part in another PR and stop at changing only the image import here.

The one thing that kinda works but could be done better is the drag and drop part. I changed things so that `shift + drag and drop` activate the other mode but that only works if the window is in focus to begin with. If you know how to do it correcly I can change the related code, I can also keep it as is, or remove it entirely (It's not the most beautiful thing ever ..)

Doublonmousse:
I think it looks good now. Let me know if there is anything to change.

The history of this PR is a little unclean. I wonder if if wouldn't be better to create a new clean branch and redo changes from there. I can't squash commits as I'd like because of intermediary merges

flxzt:
No worries about the commit history, I'll squash on merge anyway. I'll take a look soon!

crates/rnote-ui/src/canvas/imexport.rs
```diff
@@ -167,6 +170,7 @@ impl RnCanvas {
     pub(crate) async fn insert_stroke_content(
         &self,
         json_string: String,
+        resize: ImageSizeOption,
```
flxzt, line 173: `resize` -> `resize_option`

flxzt:
I think the better approach would be instead of adding additional actions and keyboard-shortcut, add a boolean option that toggles respecting-borders which is exposed in the canvas menu. This would also enable using this option without a keyboard.

crates/rnote-engine/src/strokes/resize.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+#[derive(Debug)]
+pub struct Resize {
+    /// width of a page
+    pub width: f64,
+    /// height of a page
+    pub height: f64,
+    /// if the layout has a fixed size vertically
+    pub isfixed_layout: bool,
```
flxzt, line 24: For consistency rename this to `layout_fixed_width` or similar

crates/rnote-engine/src/strokes/vectorimage.rs
```diff
@@ -1,10 +1,12 @@
 // Imports
 use super::content::GeneratedContentImages;
+use super::resize::{calculate_resize_ratio, ImageSizeOption};
 use super::{Content, Stroke};
 use crate::document::Format;
 use crate::engine::import::{PdfImportPageSpacing, PdfImportPrefs};
 use crate::{render, Drawable};
 use kurbo::Shape;
+use na;
```
flxzt, line 9: The existing codebase always uses `na::` directly for nalgebra types instead of importing it.

crates/rnote-engine/src/strokes/vectorimage.rs
```diff
@@ -137,7 +139,7 @@ impl VectorImage {
     pub fn from_svg_str(
         svg_data: &str,
         pos: na::Vector2<f64>,
-        size: Option<na::Vector2<f64>>,
+        size: ImageSizeOption,
```
flxzt, line 142: rename `size` -> `size_option`

crates/rnote-engine/src/strokes/bitmapimage.rs
```diff
@@ -100,15 +101,27 @@ impl BitmapImage {
     pub fn from_image_bytes(
         bytes: &[u8],
         pos: na::Vector2<f64>,
-        size: Option<na::Vector2<f64>>,
+        size: ImageSizeOption,
```
flxzt, line 104: `size` -> `size_option`

Doublonmousse:
I will change to a boolean option in the canvas menu. 
I first tried to add the option to every input method I could but didn't quite manage to get it for all options, so I think it's for the best.

Then I will try to test this on `main` on a new branch, to be sure it's working.

crates/rnote-engine/src/engine/import.rs
```diff
@@ -437,14 +465,40 @@ impl Engine {
...
+
+        self.store
+            .scale_strokes_with_pivot(&inserted_keys, na::Vector2::new(ratio, ratio), pos);
+        self.store.scale_strokes_images_with_pivot(
+            &inserted_keys,
+            na::Vector2::new(ratio, ratio),
+            pos,
+        );
+
+        // re generate view
```
Doublonmousse, line 493: This part is the one that's the most fragile.
When resizing of strokes occurs upon insertion (here only when pasting stroke content from an infinite layout to one with a finite width)
- strokes are pasted with no resize (that gives us the keys for the inserted strokes)
- geometry and view is regenerated
- then strokes are rescaled using the keys for the inserted strokes from before
- geometry and view is regenerated

This is one more geometry/view rerender than needed, and it also creates some transient graphical glitch showing phantom strokes that correspond to the original size, before anything gets actually resized..

Doublonmousse: fixed by putting all of that logic inside `insert_stroke_content` directly

Doublonmousse:
Alright, upon testing it seems to work fine (rescaling in the different modes, `respect_borders` for all reasonable imports of images).

Testing things a little more would be good (I'm thinking more of situations where the resizing logic is not wanted, or `respect_borders` apply where it shouldn't. I've excluded things like pdf import from being affected from these changes so it should be good).

flxzt:
I noticed these things in the code:

- You added `respect-borders` to the gschema xml but it is not bound in `appsettings.rs`.
- If I didn't miss anything, you can remove the `respect-borders` property from the `RnCanvasWrapper` entirely.
- Then you can also, instead of passing `respect-borders` to the `open_file_w_dialogs()` function as an argument, retrieve the boolean directly from inside the `try_open_file()` function. The only occurrence where this is explicitly set to false is while loading a rnote file in a new tab but there it is not used anyway.

I think otherwise the logic works well. Images always get resized to fit in the current viewport regardless of the respect-borders state but I think that's an intentional decision. Otherwise I haven't found unexpected behaviour while testing it briefly.

Doublonmousse:
- I'm going to fix the first point
- same for the other points. I was not sure if calling `respect_borders` would have to be set to `false` in some calls, but it's not the case and having it in the `try_open_file()` is also more logical (because it'd be used when called, unlike the current behaviour).

Fitting the image into the viewport is intentional (and could be changed easily if needed by changing some boolean logic inside the `resize.rs` file, adding a setting to toggle this could also be done if you want)

That's a relief you didn't find any unintentional behavior (I was kinda worried the image logic would bleed into places it was not supposed to like pasting rnote content or pdf)

I'm not sure I'm going to find enough time to go over these fixes this weekend, maybe a little bit later.

Doublonmousse:
Finally found some time to implement the changes ! Should be good to go now

crates/rnote-ui/src/appwindow/actions.rs
```diff
@@ -813,7 +818,7 @@ impl RnAppWindow {
                         }).collect::<Vec<PathBuf>>();
 
                         for file_path in file_paths {
-                            appwindow.open_file_w_dialogs(gio::File::for_path(&file_path), target_pos, true).await;
+                            appwindow.open_file_w_dialogs(gio::File::for_path(&file_path), target_pos, false).await;
```
flxzt, line 821: I think this slipped through in your latest refactor.

crates/rnote-ui/src/appwindow/mod.rs
```diff
@@ -104,6 +104,11 @@ impl RnAppWindow {
         self.imp().overlays.get()
     }
 
+    #[allow(unused)]
```
flxzt, line 107: Really minor nit, but I always add getter *and* setter's for gobject properties. (And move it slightly higher, below the other getters/setters in L86)

crates/rnote-ui/data/ui/canvasmenu.ui
```diff
@@ -60,6 +60,11 @@
             <attribute name="toggle" />
             <attribute name="action">win.block-pinch-zoom</attribute>
           </item>
+          <item>
+            <attribute name="label" translatable="yes">Respect Borders when Pasting</attribute>
```
flxzt, line 64: Also really minor, but when following Gnome HIG writing style, all word.len() >= 4 are capitalized

flxzt:
Alright, looks good now. Thank you! I am quite happy where this has arrived! :slightly_smiling_face:

Doublonmousse:
@silvasch I see you reverted this PR on your fork. Did I mess up something ? Do you want me to add a toggle to disable this all together or change the logic ?

silvasch:
No, sorry. I didn't know what the option does and didn't think it would show this here :)

I didn't try the feature yet, but I'm very excited for it because it's something that has bothered me for a long time!

Doublonmousse:
Well in its final form it applies to all image imports (and hopefully leave the rest intact). 
For image imports, it forces then to be
- smaller than the current view
- inside the page (not going over the right border) for fixed layouts
- inside the current page for all layout if the `Respect borders when pasting` option is activated

This is only on the first import, subsequent copy/pasting of the image inside the rnote document aren't affected.

There is also a small bonus : copy/pasting from an infinite document to a fixed layout document will resize strokes to be inside the page (for me it doesn't make sense to paste things that'd be largely "out of bounds" when doing something like that).

Well in any case that'd the whole changelog for this (maybe changing the commit history to this message would be better, cause the commit description is a little chaotic after squeezing)

PR #1016, opened 2024-02-12
feat: Add parabola

### LINKED ISSUE #982: Shapes: Add a Parabola
It is really difficult to make a good looking parabola using the current tools.

A new shape that creates a parabola in the shapes menu.

itsCryne:
This would resolve #982 

There may still be some rough edges as this is really my first time contributing to a bigger GUI project :innocent: 

Most code of the parabola is based upon the existing rectangle and quadratic bezier shapes.


flxzt:
Thanks! The code itself looks good at a first glance, but I don't think we need to introduce a dedicated `Parabola` Shape. We could use the already existing geometric primitives and the parabola builder could just emit a quadratic bezier curve.

Doublonmousse:
Wouldn't that (introduce a new shape) be a way to future-proof the implementation if we ever decide that we want to add better editing capacities (like editing the anchor points after the shape is constructed) in the future that are shape-specific ?

Right now this doesn't matter, but it could in the future

flxzt:
I don't really see the benefit for that. If you want to edit the shape later you could also implement it on the bezier curve, like being able to move the control points. I also want to keep the number of primitives low, otherwise we end up with a new shape for every mathematical function..

itsCryne:
Sorry for not communicating for so long - I got busy as the summer term started.
I'll try to find some time to change the implementation as suggested above in the next few weeks.

flxzt:
are you planning on finalizing this change @itsCryne ?

flxzt:
you can re-open when progress is made.

PR #1023, opened 2024-02-18
fix: failing large images export/clipboard copy #1022

### LINKED ISSUE #1022: Failing to copy to clipboard large files
**Describe the bug**  
Large images (svg and png) will fail to be copied to the clipboard if they are too large (~ 4 Mo is already too much)

**To Reproduce**  
1. paste a large image in a rnote document
2. try to copy it
3. try to paste it
5. See that the image is not pasted

**Expected behavior**  
Have the image be copied to the clipboard

**Console Output**  

```
read stream to rsvg loader failed.
```
or 
```
reading stream to rsvg loader failed.
```

 - OS: Mac os Sonoma
 - App Version: main commit 1fcb566
 - Installation Source: compile from source

**Additional context**  

This is a limitation of `librsvg` if the loader is not called with an unlimited size, see here : https://gitlab.gnome.org/GNOME/librsvg/-/issues/819

So calling 
```rs
rsvg::Loader::new()
            .with_unlimited_size(true)
```
instead of 
```rs
rsvg::Loader::new()
```
fixes the issue.

Doublonmousse:
Fixes #1022

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,3 +4,4 @@ Seio Inoue
 Moritz Mechelk
 PhilProg
 Silvan Schmidt
+Doublonmousse
--- a/crates/rnote-engine/src/render.rs
+++ b/crates/rnote-engine/src/render.rs
@@ -525,6 +525,7 @@ impl Svg {
         );
         let stream = gio::MemoryInputStream::from_bytes(&glib::Bytes::from(svg_data.as_bytes()));
         let handle = rsvg::Loader::new()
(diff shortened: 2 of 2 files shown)


flxzt:
Thanks a lot! Do you want to add yourself to `AUTHORS` ?

Doublonmousse:
Why not. Is keeping the pseudonym okay ?

flxzt:
yeah sure!

PR #1029, opened 2024-02-25
docs: added new drawing (cat of nature)

BeyondMagic:
`#NatureZa` translates directly into English as `#Nature`, it's to represent the cat's selfie in nature.


flxzt:
That's lovely :)

Issue #985, opened 2024-01-26
retain geometric shapes when exporting to xopp file


user032479798ADhiuh:
When exporting the document in xournal++ format, geometrical shapes (e.g. lines, rectangles) seem to be made part of the page background, meaning they cannot be edited or easily copied between documents. Currently, I need to export to svg and edit in a separate program (which can be time-consuming for larger diagrams as the default svg export has a far larger filesize than the xournal++ export).

**Expected behavior**  
Either the exported shapes should just correspond to xournal shapes, or if this is not possible, they should preferably just be converted to regular pen strokes, so that at least colors can be changed and drawings edited. 

**Desktop**
Linux, flatpak, latest version

Doublonmousse:
The first part seems to be a bug somewhere. The conversion is meant to export all shapes to bitmap images (afaik xournal++ does not support true svg import) that can be edited in xournal++ as images but they are consumed as part of the background instead.

I'm not sure yet where this comes, the difference between the rnote exported xml and a correct one where the image is not stuck in the background from xournal is not large, but I think I should be able to find the reason from that difference.

Matching xournal++ shapes should be feasible for a subset of shapes (not using the rough variant) that are simple enough (rectangles and lines seems relatively easy), I've yet to estimate if all of them are feasible though.

flxzt:
That's odd that an empty layer tag `<layer/> pushes the images into the background. I wonder if this is actually a bug in the file parsing in xournal++.

Doublonmousse:
I think it's actually how xournal handles background images.

If there are multiple layers with no background, only `<layer>` and `</layer>` are used like in one of the test file (in their clear xml variant here : 
https://github.com/xournalpp/xournalpp/blob/b0ce9574c17dbb31060abd9b215ed28db1e6f643/test/files/load/layer.xml#L6-L8
```xml
<layer>
<text font="Sans" size="12.00" x="171.75" y="122.25" color="black">l1</text>
</layer>
```

But the one that has an image background (https://github.com/xournalpp/xournalpp/blob/master/test/files/packaged_xopp/imgBackground/new.xopp) decompresses to 

```xml
<?xml version="1.0" standalone="no"?>
<xournal creator="Xournal++ 1.0.9">
<title>Xournal++ document - see https://github.com/xournalpp/xournalpp</title>
<page width="800.00" height="800.00">
<background type="pixmap" domain="attach" filename="attachments/bg_1.png"/>
<layer/>
</page>
</xournal>
```
[...]

PR #1030, opened 2024-02-26
only open a `layer` if the content to write is not empty

Doublonmousse:
For #985

Changing how `layer` elements are created to circumvent bugs that pushed all images down into the background instead of being editable in xournal.

I wonder if I should add to this PR some other modifications to export more shapes to their xournal equivalent.

I've glanced over the xournal code for line and rectangle as well as how these things are exported in the xournal file and it seems feasible to convert them to their xournal equivalent (in their non rough version).

### DIFF
--- a/crates/rnote-engine/src/fileformats/xoppformat.rs
+++ b/crates/rnote-engine/src/fileformats/xoppformat.rs
@@ -475,8 +475,7 @@ impl XmlWritable for XoppLayer {
     fn write_to_xml(&self, w: &mut xmlwriter::XmlWriter) {
         // only do something if we are sure the layer is not empty
         // Fix for #985
-        let is_empty = self.name.as_ref().is_none()
-            && (self.strokes.is_empty())
+        let is_empty = (self.strokes.is_empty())
             && (self.texts.is_empty())
             && (self.images.is_empty());
         if is_empty {
--- a/crates/rnote-engine/src/fileformats/xoppformat.rs
+++ b/crates/rnote-engine/src/fileformats/xoppformat.rs
@@ -475,10 +475,10 @@ impl XmlWritable for XoppLayer {
     fn write_to_xml(&self, w: &mut xmlwriter::XmlWriter) {
         // only do something if we are sure the layer is not empty
         // Fix for #985
-        let is_empty = (self.name.as_ref().is_some()
-            || (self.strokes.len() > 0)
-            || (self.texts.len() > 0)
-            || (self.images.len() > 0));
+        let is_empty = self.name.as_ref().is_some()
+            || (!self.strokes.is_empty())
+            || (!self.texts.is_empty())
+            || (!self.images.is_empty());
         match is_empty {
             false => {
                 tracing::trace!("empty layer, skipped")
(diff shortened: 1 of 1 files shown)


flxzt:
I think improving the shape export can be a follow-up because I'd imagine it is not so trivial with fills? Ideally we would have a method on all `Shape` types that decomposes them into bezier curve paths and a polygon with the fill color. We could then add a button that does that in our app as well, so users can "break" down shapes into paths and fills when they want.

crates/rnote-engine/src/fileformats/xoppformat.rs
```diff
@@ -473,22 +473,36 @@ impl XmlLoadable for XoppLayer {
...
-            image.write_to_xml(w);
-        }
-        w.end_element();
+        // only do something if we are sure the layer is not empty
+        // Fix for #985
+        let is_empty = self.name.as_ref().is_some()
+            || (!self.strokes.is_empty())
+            || (!self.texts.is_empty())
+            || (!self.images.is_empty());
+        match is_empty {
```
flxzt, line 482: Small nitpick, but i think for booleans a `if else` is better.

crates/rnote-engine/src/fileformats/xoppformat.rs
```diff
@@ -473,22 +473,36 @@ impl XmlLoadable for XoppLayer {
...
-        for text in self.texts.iter() {
-            text.write_to_xml(w);
-        }
-        for image in self.images.iter() {
-            image.write_to_xml(w);
-        }
-        w.end_element();
+        // only do something if we are sure the layer is not empty
+        // Fix for #985
+        let is_empty = self.name.as_ref().is_some()
```
flxzt, line 478: The logic is actually inverted to the variable name, it describes when the layer is not empty. Why is the name included here, can't there be layers without names? (my code comment on the layer says it's optional but I actually don't know, I don't remember how this works in xournal++)

Doublonmousse:
Well, the scope kinda changed with me figuring out the layer issue. It's more of a small usability tweak (prevent empty layers as that will only impede selection by the user until they change or remove the empty layer).

Unless you want to change the logic so that everything gets written into a single xournal layer (keeping the same order but only opening a single `layer` so that images are still below strokes) I don't think there's anything more to do here. It would make it easier for xournal users to have a single layer but at the same time this is a xournal issue.

crates/rnote-engine/src/fileformats/xoppformat.rs
flxzt: The layer name is still in there, please take it out of the condition because it does not indicate if a layer is empty.

flxzt:
I actually think the separation into different layers is something that is desirable so let's keep the current behaviour. Imho the problem lies in Xournal++ which should have more intuitive UI for accessing/modfying these layers.

Doublonmousse:
Okay.
Do you think it's useful to keep this PR and skip empty layers (as a small usability tweak) or should I close this ?

flxzt:
Yes definitely! I'll merge once you resolve the other issue and mark it ready.

flxzt:
Thanks!

Issue #990, opened 2024-01-29
"Opened file was modified on disk" pop up when saved


MisterE123:
**Describe the bug**  
Whenever the file is saved, via autosave or manual save, the following pup up appears after a second or so:
![image](https://github.com/flxzt/rnote/assets/61124264/9444874b-61be-4038-a4f7-80b14f129477)

**To Reproduce**  
1. be on windows
2. save the file

**Expected behavior**  
Just save the file, don't give that annoying pop up

**Desktop (please complete the following information):**  
 - OS: Windows 11
 - App Version: Rnote v0.9.4
 - Installation Source: Windows installer

ImTheCurse:
It only seems to happen to me when I save the .rnote files on oneDrive / google drive, but not on local save.

MisterE123:
I am saving the files on my onedrive filesystem

digitalsignalperson:
I'm on linux with ZFS and I'm seeing this "Opened file was modified on disk" on autosave. One time I clicked it and it reverted my file to be empty.

flxzt:
Can you run the app with trace logging enabled (Either `flatpak --env=RUST_LOG=rnote=debug run com.github.flxzt.rnote` or `RUST_LOG=rnote=debug rnote` ) and post the output, excluding the pointer events?

digitalsignalperson:
I realize I had two instances open. One was on a 3rd virtual desktop which I didn't notice. It had the file open when it was just empty. So it was overwriting the file with an empty copy. When I looked back in my ZFS snapshot history, the file sizes were alternating between 256KB and 743 bytes (empty). Today I opened the file and it was empty, and I realized there was the extra instance open. Unfortunately I didn't have the right snapshot for the last 6 hours or so so lost some notes.

Should I open a separate feature request for something like: "prevent same file being auto-saved in multiple instances"?

olbotta:
to me this happens when i save very quickly twice a file

MisterE123:
This may have been fixed on the latest version (0.11.0)

You could add possible close @flxzt

Doublonmousse:
@digitalsignalperson I'm opening another issue for this #1424 

@olbotta @ImTheCurse Does it still happen on 0.11 or 0.12? If so, is it with cloud storage? 
If this is only with cloud storage, I'll close this in favor of #461 for the cloud storage part

ImTheCurse:
> @digitalsignalperson I'm opening another issue for this #1424 
> 
[...]

Issue #894, opened 2023-11-20
Saving file on external google drive crashes app


tipptop3d:
**Describe the bug**  
When you try to save a document to a external drive (only tested for my case: google-drive://mymail@gmail.com/), the app crashes.

**To Reproduce**  
1. Open app
2. Draw something
3. Crtl-S
4. Navigate to external drive
5. Press Save
6. App crashes

**Expected behavior**  
File saved to the external drive

**Console Output**  
```zsh
$ flatpak --env=RUST_LOG=rnote=debug --env=RUST_BACKTRACE=1 run com.github.flxzt.rnote
 2023-11-20T15:09:44.839Z DEBUG rnote > ... env_logger initialized

(rnote:2): Gtk-CRITICAL **: 16:09:45.363: gtk_widget_class_set_accessible_role: assertion '!gtk_accessible_role_is_abstract (accessible_role)' failed

(rnote:2): GLib-GIO-CRITICAL **: 16:10:21.202: g_file_get_path: assertion 'G_IS_FILE (file)' failed
```

**Desktop (please complete the following information):**  
 - OS: Fedora 39
 - App Version: Rnote v0.9.1
 - Installation Source: Flatpak
 - Desktop Environment: Gnome 45.1

**Additional context**  
I try to sync my files between devices using gdrive. Saving files on Windows using the gdrive drive (G:) provided by installing gdrive is functioning flawless. On my linux device, the gdrive provided by logging in with google in GNOME is not functioning.

flxzt:
hm, I can't replicate it on my machine (I am on Fedora 39 too). Can you test if this occurs in other gtk4/libadwaita based apps?

tipptop3d:
gtk3 apps work just fine.
I also tried opening files on [celluloid](https://github.com/celluloid-player/celluloid), which works, only file/folder names are really weird.
![image](https://github.com/flxzt/rnote/assets/45404249/76febdf3-d371-4848-af9e-1fce51d1114b)
The same for [G4Music](https://github.com/neithern/g4music).
![Screenshot from 2023-11-23 15-54-23](https://github.com/flxzt/rnote/assets/45404249/37fa24dc-566d-4556-8241-85c4e8d5c360)
[Apostrophe](https://gitlab.gnome.org/World/apostrophe) couldn't open nor save files. If you have any other potential app in mind, feel free to tell me.

flxzt:
[...]

Issue #815, opened 2023-09-28
Popup warning about file change, prompting for a reload after saving in a folder with OneDrive sync activated


florentdup:
**Describe the bug**  

When saving a note in a folder with OneDrive synchronization enabled, a popup appears at the bottom telling you that a change was made and prompts you to reload the file. 

**To Reproduce**  

  1. Create a note.
  2. Save the note in a folder with OneDrive synchronization on.
    Popup prompting a reload appears.

Or alternatively, 
  1. Load an existing note saved in a folder where OneDrive is enabled
  2. Save the note.
    Popup prompting a reload appears.

**Expected behavior**
No Popup appear, since OneDrive doesn't modify the content of the file when uploading it to the cloud.

**Desktop**  
 - OS: [Windows 11 22H2]
 - App Version: [Rnote v0.8.2]
- Installation Source: [Windows release, rnote-win-installer-0.8.2-x86_64.exe ]

**Additional context**  
Maybe this bug, similar to issue 434 is due to the fact that OneDrive changes some kind of metadata or some kind of status when uploading the newly modified file to the cloud, but I believe OneDrive doesn’t modify the content of the file itself, so no modification should be detected by rnote.

flxzt:
I think you're right about this behaviour. We should start checking the monitored file by it's content to detect whether it was modified or not.

#461 might also be related.

PR #1033, opened 2024-02-27
refactor: improve canvas file watcher

flxzt:
using the `notify` and `async-fs` crates. They do not have the particularities of the glib file monitor and goutputstream files.

addresses #894, #990, #815

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1224,6 +1224,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "file-id"
+version = "0.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


flxzt:
Seems to be working fine on Windows as well, tested with Win 11 VM

flxzt:
Renames seem to only emit a `Event { kind: Modify(Name(From))`, this could be considered a regression because then the file becomes a unsaved New Document and the user gets notified that it is "moved or renamed". If it would be a [RenameMode::Both](https://docs.rs/notify/6.1.1/notify/event/enum.RenameMode.html) event it would be handled correctly. It seems to be a limitation of notify-rs.

But considering it still is more robust than the fragile handling of `.goutputstream` files with the current monitor, I still want to land this.

Issue #978, opened 2024-01-15
Error on startup - Dynamic link library entry point not found Neovim


Tfloow:
**Describe the bug**  
When launching the app on windows 11 64 bit, I get a pop-up window for the same error twice:

1. Entry Point _ZN3egl10CreateSyncEPvjPKx not found in the dynamic link library `C:\Program Files\Neovim\bin\libEGL.dll`

And sometimes leads to crash with the same error while using it.

**To Reproduce**  
1. _My guess:_ Have Neovim installed for windows
2. Launch the app on windows

**Expected behavior**  
No such pop-up and error

**Console Output**  
*Can't run the flatpak command because I am running windows*

**Screenshots**  
![image](https://github.com/flxzt/rnote/assets/88899420/2b61c361-459d-4d3e-9ad9-d7a76847c350)

**Desktop (please complete the following information):**  
 - OS: Windows 11 22H2
 - App Version: Rnote v0.9.4
 - Installation Source: Winget (windows store)
 - Desktop Environment: Windows
 - Display Server: X11
 - Input Source: Mouse, keyboard & graphic tablet

flxzt:
Reads like issue #938 - apparently the fix that closed that issue did not work fully

bluerosegarden:
Yep, when adding the [libegl.dll I found here](https://www.dll-files.com/download/4d2f81c874275b9515daf20caa0c21a7/libegl.dll.html?c=VTVoUkwrcWpYanUrVVNrajFrRDNyUT09) it worked great! I'm on windows however and installed using the installer.

I can't vouch for the whether or not the site is safe, but I scanned it with microsoft defender and it came up clean so I figured why not lmao

flxzt:
I removed the link.

flxzt:
@Tfloow please check if it is fixed for you with v0.10.1

Tfloow:
Yes it is now working I don't have the issue anymore. Thank you

flxzt:
Thanks for the confirmation!

PR #1041, opened 2024-03-06
fix: win error broken dyn deps on certain systems

flxzt:
addresses #978 

This adds libEGL*.dll and libGLES*.dll files to the installer build.
I verified it by building the installer on my windows system.
Just to be safe, this should suffice, right? @mdonoughe


mdonoughe:
LGTM

Issue #1017, opened 2024-02-13
Resizing increases line width of strokes


itsCryne:
**Describe the bug**  
Repeatedly resizing brush or shape paths leads to them increasing their thickness

**To Reproduce**  
1. Draw something using the Brush or Shaper
2. Select it
3. Repeatedly resize it

**Expected behavior**  
scale1 == scale2 <=> thickness1 == thickness2

**Console Output**  
```
 2024-02-13T14:28:39.745Z DEBUG rnote > ... env_logger initialized

(rnote:287344): Gtk-CRITICAL **: 15:28:40.110: gtk_widget_class_set_accessible_role: assertion '!gtk_accessible_role_is_abstract (accessible_role)' failed
 2024-02-13T14:28:40.128Z DEBUG rnote_engine::store > state has not changed, no need to record

(rnote:287344): Gtk-CRITICAL **: 15:29:13.510: The layout manager of type GtkBoxLayout 0x5653416400b0 does not create GtkLayoutChild instances
```

**Screenshots**  

https://github.com/flxzt/rnote/assets/51213332/0c87ab41-4dcc-424f-a5c3-51fa784ac476

**Desktop (please complete the following information):**  
 - OS: Manjaro
 - App Version: Tested with Rnote v0.9.4 & 551b1c5ed141d5089d851a11771387f832269e87
 - Installation Source: Arch package & built from source
 - Desktop Environment: GNOME 45.3
 - Display Server: Wayland
 - Input Source: Mouse and Keyboard, On-Screen-Tablet
 
**Additional context**  
\-

Doublonmousse:
The code that changes line width based on scale was probably implemented in #808.
I think the line width is changed in place when resizing occurs, so that it can get to these large sizes, instead of staying as a `width * (x+y)/2` throughout the resize action

Doublonmousse:
Basically, for each step of the resize movement, the scaling factor is applied with `width <- width * (x+y)/2` with `x` and `y` the scaling ratio for both axis with respect to the last one. 

Unless you lock the ratio, changing the ratio creates this cumulative effect.

Could the resize take as `scale` the ratio with respect to the initial size of the selection bound (so we need to save it) and apply the width with respect to the initial line width of the selected strokes (that has to be saved as well for as long as elements are selected) ?

This way, resizing will be more predictable and still keep the instant preview feature (the line stroke corresponds to the final one, unlike xournal)

PR #1050, opened 2024-03-19
put selection handle outside selection

Doublonmousse:
image

### DIFF
--- a/crates/rnote-engine/src/pens/selector/penevents.rs
+++ b/crates/rnote-engine/src/pens/selector/penevents.rs
@@ -321,8 +321,7 @@ impl Selector {
                             let offset_mean = offset_to_start.mean();
                             offset_to_start = start_extents * (offset_mean / start_mean);
                         }
-                        let min_extents = (Self::RESIZE_NODE_SIZE
-                            + na::Vector2::<f64>::from_element(Self::ROTATE_NODE_DIAMETER))
+                        let min_extents = (na::Vector2::<f64>::from_element(2.0f64))
                             / engine_view.camera.total_zoom();
                         let scale = (start_bounds.extents() + offset_to_start)
                             .maxs(&min_extents)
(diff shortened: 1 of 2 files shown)


Doublonmousse:
I've updated the minimum size (for now it's set at `2.0 / zoom_level`).
I don't see any of the issues I had with me setting it to 0.01 so that's good.

Having smaller minimum selection size does make #1017 more prominent though.
I feel like changing the stroke width calculation so that the following are true could help a little though (maybe a pattern matching could suffice)
- scale =(1,r) => scale_stroke = r
- scale =(r,1) => scale_stroke = r
- scale = (x,y) => scale_stroke = (x+y)/2

At least it would remove the issue for resize that are done on a single direction, leaving only the case of a series of resizes that are neither uniform nor done in a single direction without having to calculate the stroke width with respect to the initial size of the selection

crates/rnote-engine/src/pens/selector/penevents.rs
```diff
@@ -321,8 +321,7 @@ impl Selector {
                             let offset_mean = offset_to_start.mean();
                             offset_to_start = start_extents * (offset_mean / start_mean);
                         }
-                        let min_extents = (Self::RESIZE_NODE_SIZE
-                            + na::Vector2::<f64>::from_element(Self::ROTATE_NODE_DIAMETER))
+                        let min_extents = (na::Vector2::<f64>::from_element(2.0f64))
```
flxzt, line 324: unneeded parentheses here

flxzt:
That's a very interesting idea, let's discuss it further in #1017

crates/rnote-engine/src/pens/selector/penevents.rs
Doublonmousse: fixed

flxzt:
Thank you! :)

PR #1065, opened 2024-03-31
README.md: add archlinux installation source

Integral-Tech:


flxzt:
Thanks for the contribution, but for now I only want to advertise the flatpak since that one is controlled by me. My guess would be that arch users already know how to check and install the rnote archlinux package anyway

PR #1067, opened 2024-04-01
disable options for all shortcuts

### LINKED ISSUE #1060: Set pen action button to "none"
I think can be useful if you can set the pen buttons action to none
image
how you can see you can't disable the input pen action.

Yes, but not a rnote problem, i use a surface pro 8, with linux, however the driver for the pen is not completed (not working really well) and touching the back of the pen is equals to press a pen button, I think it could be useful for problem pens like this one.
  
Add a button "no action" or something like that, that just disable any action

**Additional context**  
I know it's not an rnote issue, but I think it could still be a useful function for many (certainly me 😆)

Doublonmousse:
Fixes #1060

### DIFF
--- a/crates/rnote-engine/src/pens/penholder.rs
+++ b/crates/rnote-engine/src/pens/penholder.rs
@@ -299,6 +299,7 @@ impl PenHolder {
                             widget_flags |= self.change_style_int(style, engine_view);
                         }
                     }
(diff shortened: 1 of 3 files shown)


flxzt:
that's a straight-forward improvement, thanks :slightly_smiling_face:

PR #1068, opened 2024-04-01
reset colors

### LINKED ISSUE #1056: Option to reset the colors
Add a button to reset the color palette to its default configuration.

Sometimes, whilst playing with different colors, we can end up with a color toolbar that is worse than the default on. For example, I changed it at some point and now I have two identical greens.

A button (not sure where to put it though) to reset the color selection to its default state.

- Recreate the color palette manually 
- have the windows that's used to change colors include the default color selection used in rnote so recreating the default choice of color is easier (not sure if this is the easiest way to do this, from the programming side)

Doublonmousse:
To fix #1056 

A few things I'm not too sure about 

- Is the position of the button correct ?

image

- should other color things (that are in the document settings like format border color, pattern color and background color) be affected ?

### DIFF
--- a/crates/rnote-ui/data/ui/settingspanel.ui
+++ b/crates/rnote-ui/data/ui/settingspanel.ui
@@ -387,6 +387,18 @@ gets disabled.</property>
                         </child>
                       </object>
                     </child>
+                    <child>
+                      <object class="AdwActionRow" id="reset_all_colors_row">
+                        <property name="title" translatable="yes">Reset Colors</property>
+                        <property name="subtitle" translatable="yes">Reset stroke/fill color choices</property>
+                        <child type="suffix">
+                          <object class="GtkButton" id="reset_all_colors_button">
+                            <property name="valign">center</property>
+                            <property name="label" translatable="yes">Reset Colors</property>
+                          </object>
+                        </child>
+                      </object>
+                    </child>
                   </object>
                 </child>
                 <!-- Button Shortcuts Group -->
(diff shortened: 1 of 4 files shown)


flxzt:
Shouldn't this be a reset for all color picker palette colors as described in issue #1056 ?

Doublonmousse:
I'm not 100 % sure. Given the current position, one would expect it only applies to color picker palettes of the `Document` part of settings.
But I don't want to mix things that affect the document (document background, grid color) with ones that don't (choice of pen colors).

Ideally this button should be in the color picker. But that means modifying the color picker (where almost all default colors are presented, though not in the exact same order)
image

The current position of the button is very similar to the `revert/apply` buttons for the page format. Maybe It'd be better to mirror this for the `Document` section and have a `reset/templates` button combo (with the `templates` part to populate maybe later)

Doublonmousse:
Maybe it'd be better to add a right click menu (that would also be triggered by a long touch press, and maybe pen long press) for each color in the color picker window to reset the colors.

(similar to how a right click in the color picker shows a `customize` option)


That could also be done for the individual color setters that are in the settings panel.

Doublonmousse:
Well I tried to detect right clicks and long press gestures on each `RnColorSetter` but only right click gesture detection works. Long press will only work once for each color setter then stop working.

Not sure how to do it correctly and where to put code between `RnColorSetter` and `RnColorPicker`, so if anyone has a better idea on how to do that (I would've added a dropdown menu with at least a reset option for each color), feel free to continue this PR.

Will close it otherwise

flxzt:
The long-press Gtk gesture has some issues, it's finicky to make it work. I've had similar issues with it in the past. I think the idea to implement the reset in this is very good in general, maybe we can revisit it at some point..

Doublonmousse:
Relevant thread (ending in "just reimplement the whole thing with a custom widget from scratch")

...

https://discourse.gnome.org/t/gtk4-need-button-pressed-and-released-signals/8506/18

flxzt:
Yeah in this case I can understand the reasoning why this can't be supported by the default GtkButton implementation. Subclassing and create a custom widget actually isn't that uncommon, we already do that in several places in the app

PR #1075, opened 2024-04-11
fix: `filters` not working for the save dialog

### LINKED ISSUE #1073: Problem at the moment of Export the archive,
When I save an archive, it exports as a '.archive' file, but it can't be opened with any program, including Rnote. This issue occurs most of the time, although occasionally it works. Additionally, today I encountered problems when saving the PDF archive. I'm working with Rnote Version 0.10.0 on Windows 11.

Doublonmousse:
To fix #1073 .

Makes the filter on the save dialog work. 

image

Before this the `file type` part was empty (filter not applied) leading to various potential errors, including the possibility to create files that have no extension.

### DIFF
--- a/crates/rnote-ui/src/dialogs/import.rs
+++ b/crates/rnote-ui/src/dialogs/import.rs
@@ -50,43 +50,31 @@ pub(crate) async fn filedialog_open_doc(appwindow: &RnAppWindow) {
 }
 
 pub(crate) async fn filedialog_import_file(appwindow: &RnAppWindow) {
-    // use a list of filters
-    let filters = gio::ListStore::new::<FileFilter>();
+    let filter = FileFilter::new();
+    filter.add_mime_type("application/x-xopp");
+    filter.add_mime_type("application/pdf");
+    filter.add_mime_type("image/svg+xml");
+    filter.add_mime_type("image/png");
+    filter.add_mime_type("image/jpeg");
+    filter.add_mime_type("text/plain");
+    filter.add_suffix("xopp");
+    filter.add_suffix("pdf");
+    filter.add_suffix("svg");
+    filter.add_suffix("png");
+    filter.add_suffix("jpg");
+    filter.add_suffix("jpeg");
+    filter.add_suffix("txt");
+    filter.set_name(Some(&gettext("Jpg, Pdf, Png, Svg, Xopp, Txt")));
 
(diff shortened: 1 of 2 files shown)


flxzt:
Thanks a lot! Adding `filters()` before using `default_filter()` is  missing in several places in the code, do you want to fix this everywhere?

Doublonmousse:
On it. For now I'm doing it kinda "naively" for the export part (meaning that there's a single filter shown for `.svg` if `svg` is selected in the previous export window, but there is no bidirectional link that would allow to change the file type in the file chooser window to say, a `.pdf` file and have the change propagated back into the application)

Doublonmousse:
I do get some 
```
(rnote:36736): Gtk-WARNING **: 13:05:01.848: gtk_file_chooser_add_filter() called on filter already in list
```
for the import file dialog somehow.

flxzt:
`ListStore`'s are also deprecated..

But I'd prefer to keep it a single filter anyway, because then it is immediatly clear what is importable - everything that is visibily listed in the filepicker

crates/rnote-ui/src/dialogs/import.rs
```diff
@@ -46,27 +50,43 @@ pub(crate) async fn filedialog_open_doc(appwindow: &RnAppWindow) {
...
+    filters_vec[4].add_suffix("jpg");
+    filters_vec[4].add_suffix("jpeg");
+    filters_vec[4].set_name(Some(&gettext("Jpg")));
+
+    filters_vec[5].add_mime_type("text/plain");
+    filters_vec[5].add_suffix("txt");
+    filters_vec[5].set_name(Some(&gettext("Txt")));
+
+    // append all these filters to the gio::ListStore
+    filters_vec.iter().for_each(|filter| filters.append(filter));
```
Doublonmousse, line 83: I'll revert this part (to use one filter here) and re check if the error persists.

Doublonmousse:
Is it ? On gtk-rs `gtk4::ListStore` is shown as deprecated, with `gio::ListStore` to use instead.

flxzt:
My bad, `gio::ListStore` is not

crates/rnote-ui/src/dialogs/import.rs
flxzt: Ok, sounds good!

flxzt:
Works now for me. Thanks for this fix :)

Issue #914, opened 2023-12-05
Make Zoom tool zoom in the direction of the pointer


dmaioli:
**Is your feature request related to a problem? Please describe.**  
Currently the Zoom tool zooms in the direction of the center of the page. This makes hard for a drawing tablet user to zoom on a specific object, because it is necessary either to zoom with the touchpad with the other hand, or to very carefully move the page such that the object of interest is in the center and then zoom. 

**Describe the solution you'd like**  
Make the Zoom tool zoom in the direction of the pointer, as suggested in #695 and #580 (marked as complete by mistake?). This allows to move the pointer on the object of interest and zoom in a very simple movement, and to be consistent with the touchpad zoom gesture, which follows the pointer.

**Describe alternatives you've considered**  
I have considered all the other zoom mechanism described in #539, but even with them a drawing tablet user that wants to zoom on a specific area is forced either to use the touchpad with the other hand or to carefully position the object of interest and the zoom.

flxzt:
The mentioned closed issue were referring to the zoom behaviour when holding <ctrl> + Scrolling with the mouse scroll-wheel.

But the zoom-tool could also be improved in the same way, I agree.

flxzt:
implemented with #1080

PR #1080, opened 2024-04-13
Made the zoom tool zoom towards the initially selected location

JonathanGodar:
As mentioned in #914 the zoom tool zooms towards the center of the screen. This PR makes the zoom tool zoom towards the selected start location. As mentioned in previous issues this makes navigating with a drawing tablet easier.

### Area of improvement
The "zoom formula" is copied from [this file](https://github.com/flxzt/rnote/blob/ba0c40ef25a8bf88d1f09bd46858017baf4fa88f/crates/rnote-ui/src/canvaswrapper.rs#L348). I think that there may be a need for a function in the camera to have a `zoom_towards` function. However I don't feel quite comfortable enough yet with the codebase to make that happen.

### DIFF
--- a/crates/rnote-engine/src/pens/tools.rs
+++ b/crates/rnote-engine/src/pens/tools.rs
@@ -428,7 +428,7 @@ impl PenBehaviour for Tools {
                                 - self.zoom_tool.start_surface_coord;
                             widget_flags |= engine_view
                                 .camera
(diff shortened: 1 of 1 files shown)


flxzt:
nice, thanks a lot!

I don't think it's too critical to re-use this behaviour in a method, but if you want you could of course follow-up with a refactor, that would definitely be appreciated :slightly_smiling_face:

Issue #658, opened 2023-05-01
Improvements for EPD


flxzt:
**Is your feature request related to a problem? Please describe.**  

The app in it's current state works very well for 2-in-1 tablets, good for external graphics tablets and decent for touch-only tablets.

There is a new, very exciting class of devices: open Epd tablets with fast-response times, such as the PineNote/ Bigme. I recently got that chance to test the app on a PineNote, and while a lot of feature work decently well, I noticed some papercuts where we could improve the experience significantly:

- the input lag is quite high, because I believe the event emission is locked to the display refresh rate. Epd's are special because they have full refreshes, which take ~0.5s, and partial refresh, which are much faster. Unfortunately it seems that the limit is full refresh here. There is a debian [available](https://github.com/m-weigand/pinenote-debian-recipes) configured for the pinenote which seems to include a patched xournalpp, which draws much more smoothly. We should look into if this can be done for rnote as well and if this is something that can be improved upstream either in gtk, gnome, libinput, .. (?) . Since we use the backlog for each event, the actual drawn stroke looks perfectly fine.  
- the pinenote has a "dithering" mode, where the default preset colors can be decently distinguished, because the dither pattern differ between them. Drawn strokes also look good with them, but it might be desirable to know the actual color, especially for export.
  - a possible solution: detect the rough "categories" of the color and display them as strings somewhere. The most intuitive would probably be the dimensions "hue, saturation, value" or "hue, saturation, brightness [hsv, hsb](https://upload.wikimedia.org/wikipedia/commons/thumb/1/16/Hsl-hsv_models_b.svg/600px-Hsl-hsv_models_b.svg.png?20110609185029).
- The instant refresh for dragging / zooming is not ideal for an Epd, because it updates the view a lot which results in constant display refreshes.
  - What we could do: Add a mode where moving the view is only indicated by either the bounds, corner, or its center. This would reduce the refresh noise significantly. Alternatively we could "limit" translations and scaling by tresholds (like: only every 50pt's or so / every 10% of zoom)

I am sure there are more things that could be improved, so I plan to add more here over time.

PhilDevProg:
The OSK from GNOME uses the [text_input_unstable_v3](https://wayland.app/protocols/text-input-unstable-v3) Wayland protocol. I don't know how the [Maliit keyboard](https://github.com/maliit/keyboard) used by KDE Plasma works though...

hmpthcs:
[...]

PR #1081, opened 2024-04-13
feat: add improve for Epd option, add active-color-label

flxzt:
This adds the following:

- A label that displays the current active color as a string in the format \<brightness\>,\<saturation\>,\<hue\> with easy to understand strings.
- An "optimize for E-Paper displays" option in order to toggle the visibility of this label. In the future more modifications to the app behaviour and UI additions can be enabled through it

Still a draft, because rnote-compose now needs to be translated with gettext as well. I'll see how to best integrate or refactor it

addresses #658


flxzt:
@Doublonmousse if you want/have enough time, please review!

crates/rnote-ui/po/rnote.pot
```diff
@@ -2318,3 +2332,114 @@ msgstr ""
...
+
+#: crates/rnote-ui/src/utils.rs:176
+msgctxt "part of string representation of a color"
+msgid "black"
+msgstr ""
+
+#: crates/rnote-ui/src/utils.rs:178
+msgctxt "part of string representation of a color"
+msgid "white"
+msgstr ""
```
Doublonmousse, line 2445: The rule for `gettext` to work correctly is to match the line the `pgettext` is called in the `pot` file right ?
I'm not 100 % sure it's the case here

line number in `utils.rs` vs `rnote.pot` (and associated name)

148/145/rose
150/146/red
153/147/orange
156/148/yellow
158/149/chartreuse-green
163/150/green
166/151/spring-green
169/152/cyan
172/153/azure
175/154/blue
178/155/violet
181/156/magenta
183/157/rose
184/158/invalid

187/161/desaturated
189/163/vibrant
190/164/invalid

193/167/dark
195/168/mid
197/169/bright
198/170/invalid

202/174/transparent
204/176/black
206/178/white

Doublonmousse (commented):
Alpha values are not shown in the name for the color ?

Testing with some colors, I do have some weird results on some slightly desaturated colors.

- #986A44, #63452C shows as mid orange 
- #FF7800 shows as white (it's a slightly desaturated orange)

I think it'd be good to plot where the limits are in the `Okhsv` space to double check this

crates/rnote-ui/po/rnote.pot
flxzt: I *think* the line number only helps for matching or even is only a help for translators to get additional context by looking at the source code, the msgid is what will be matched against in the end.
I don't know why it wrote the wrong line numbers, but it seems to be fixed with the latest execution of `meson compile rnote-pot` (though I only checked a few entries)

flxzt:
Good points, thanks.

I think I misunderstood the `value` field in the hsv colorspace while implementing this. I now fixed and improved the logic by:
- combining checks for value **and** saturation to detect white and black
- accounting for float inaccuracies by using `approx::rel_eq` with f32's epsilon for min/max values. It's relevant because the picked colors come from gdk, which only has f32 precision.
- hand-tweaked the ranges for colors. This is the part which will probably need further testing, because I don't have a calibrated screen. But at least the Gnome colors in the colorpicker are now always in the right category. Ideally there would a definitive (scientific) source for defining hue values as strings, but didn't find one yet.
- added alpha strings `transparent`, `translucent`, `slightly translucent` and `` (empty string == opaque)

Doublonmousse (approved):

PR #1089, opened 2024-05-04
feat: Open files/folder in the file explorer

### LINKED ISSUE #1053: Option to open Workspace in file manager
**Problem**

The Workspace file browser doesn't have much features like searching, renaming etc.

**Suggested solution**

Add an option to open the Workspace in the default file manager. I think this option is useful even if you add searching, renaming etc. features in the app.

**Optional**

You may add the option in the red marked location or wherever you find good.

**Additional context**

image

Doublonmousse:
To fix #1053.

This is a PR that's very similar to #868, using a slightly different crate (`opener`) that has a `reveal` call and a  MIT OR Apache-2.0 license.

The implementation detail for this is here https://docs.rs/opener/latest/opener/fn.reveal.html
The `open` crate doesn't have this option afaik but having two crates doing very similar things is slightly redundant

Tested only on linux aarch64 for now (but should be supported on all platforms)


Doublonmousse:
Tested on mac os (apple silicon) and windows as well. I haven't tested remote folders though.

flxzt:
We should probably only use one of `open` or `opener`, reads like `opener` is a bit more capable without adding many dependencies, so we could switch to it entirely.

You implemented it a bit differently than expected - the alternative would be that the feature is an additional action in the folder menu at the top that opens the current workspace folder, not an action on every list item. But I guess we could have even both?

The advantage of your implementation is that when opening a file/folder, it it already selected.

PR #1095, opened 2024-05-12
ui(AboutWindow): fix developers credits list

zefr0x:
### Before
screenshot_2024_05_13_02_25_21
### After
screenshot_2024_05_13_02_25_13

---

IIRC there is no way to format `@APP_AUTHORS@` as a comma separated strings using meson, but if there was it would be more efficient.


flxzt:
thank you!

PR #1102, opened 2024-05-18
change of inkstroke modeler

Doublonmousse:
On the rnote side : 
- cargo.toml change (for now using a custom branch/fork of the inkstroke modeler)
- API change (usize for n_steps, no PredictionParams and no tilt or orientation)

On the inkstroke modeler side
- rewrite to rust of the logic
This rewrite is mostly to have something that's easier to build/compile from rust on all platforms.

### DIFF
--- a/crates/rnote-compose/src/builders/penpathmodeledbuilder.rs
+++ b/crates/rnote-compose/src/builders/penpathmodeledbuilder.rs
@@ -201,7 +201,10 @@ impl PenPathModeledBuilder {
                         self.last_element = element;
                         tracing::error!("Updating stroke modeler with element failed, Err: {e:?}")
                     }
-                    _ => return,
+                    _ => {
+                        tracing::error!("Updating stroke modeler with element failed, Err: {e:?}");
+                        return;
+                    }
                 };
             }
         }
--- a/crates/rnote-compose/src/builders/penpathmodeledbuilder.rs
+++ b/crates/rnote-compose/src/builders/penpathmodeledbuilder.rs
@@ -186,17 +186,17 @@ n_steps exceeds configured max outputs per call."
         }
         self.last_element_time = now;
 
-        let modeler_input = ModelerInput::new(
+        let modeler_input = ModelerInput {
             event_type,
-            (element.pos[0] as f32, element.pos[1] as f32),
-            now.duration_since(self.start_time).as_secs_f64(),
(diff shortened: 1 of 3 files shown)


crates/rnote-compose/src/builders/penpathmodeledbuilder.rs
```diff
@@ -189,19 +186,17 @@ n_steps exceeds configured max outputs per call."
...
 
         match self.stroke_modeler.update(modeler_input) {
             Ok(results) => self.buffer.extend(results.into_iter().map(|r| {
-                let pos = r.pos();
-                let pressure = r.pressure();
+                let pos = r.pos;
+                let pressure = r.pressure;
                 Element::new(na::vector![pos.0 as f64, pos.1 as f64], pressure as f64)
             })),
             Err(e) => tracing::error!("Updating stroke modeler with element failed, Err: {e:?}"),
```
Doublonmousse, line 202: I feel like changing the error types in the modeler could make this part completely redundant.
Just call the update on the model, match on the results and different errors messages (maybe an enum type for errors), this way we don't test twice that the events are not a duplicate, that the time delta is positive and that the number of steps is not too large.

Doublonmousse:
Uh now I have issues when I pause the stroke and continue moving later

crates/rnote-compose/src/builders/penpathmodeledbuilder.rs
flxzt: Yeah, that's what I meant with strongly typed errors in `ink-stroke-modeler`. But I also want `riir` to land soon, it can be a follow-up

crates/rnote-compose/src/builders/penpathmodeledbuilder.rs
```diff
@@ -164,65 +164,65 @@ impl PenPathModeledBuilder {
...
+                        );
+                        self.restart(element, now);
+                    }
+                    ModelerError::Element {
+                        src: ElementError::Order { src: _ },
+                    } => {
+                        self.last_element = element;
+                        tracing::error!("Updating stroke modeler with element failed, Err: {e:?}")
+                    }
+                    _ => return,
```
flxzt, line 204: Other ModelerError's that might get added in the future should be logged as well, instead of silently returning

PR #1110, opened 2024-06-04
ui: migrate to libadwaita-1.5 and its adaptive dialogs

zefr0x:
I've used [`AdwAboutDialog`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.AboutDialog.html) and [`AdwAlertDialog`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.AlertDialog.html) in place of [`AdwAboutWindow`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.AboutWindow.html) and [`AdwMessageDialog`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.MessageDialog.html) respectively.

Old dialogs using [`GtkDialog`](https://docs.gtk.org/gtk4/class.Dialog.html) are still displayed in separate windows. We might look at the possibility of using [`AdwDialog`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.Dialog.html) as an alternative.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
   check:
     if: github.event.pull_request.draft == false
     runs-on: ubuntu-22.04
-    container: fedora:39
+    container: fedora:40
     steps:
       - name: Install dependencies
         run: |
--- a/.github/workflows/dist.yml
+++ b/.github/workflows/dist.yml
@@ -8,7 +8,7 @@ name: Generate Dist Archive
 jobs:
   dist:
     runs-on: ubuntu-22.04
-    container: fedora:39
+    container: fedora:40
     permissions:
       contents: write # needed for uploading release artifact
     steps:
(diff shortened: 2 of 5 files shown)


flxzt:
Looks good and straight forward, is the reason that this is a draft because you want to look into using AdwDialog instead of GtkDialog or will this be a possible follow-up?

zefr0x:
> is the reason that this is a draft because you want to look into using AdwDialog instead of GtkDialog or will this be a possible follow-up?

Yes, if you want, you can merge these changes now, but it may be a good idea to merge them all at once.

I didn't find time to look into AdwDialog, but i can say that actions related methods present in GtkDialog are not present in AdwDialog, so they may have to be implemented manually.

flxzt:
Hm, not yet sure why the CI fails. The used Fedora container should have libadwaita v1.5 available..

flxzt:
oh, of course.

flxzt:
> I didn't find time to look into AdwDialog, but from a fast reading, I can say that actions related functionalities present in GtkDialog are not present in AdwDialog, so they may have to be implemented manually.

This is why I think it can be a follow up since it requires more refactoring than the other replacements.

I'll merge this now, and open an issue for tracking GtkDialog -> AdwDialog porting

flxzt:
Thanks for this contribution!

Issue #785, opened 2023-09-01
Pen buttons not working as intended on windows 11 on Surface pro


Doublonmousse:
**Describe the bug**  
On windows 11, the primary button of the stylus is not working as intended.

When the button is pressed, this is not detected until the screen touches the screen, even when the pen is in proximity of the screen.

This leads to unwanted behavior : 
- When the temporary mode is selected, one would expect to be able to select objects as soon as the pen touches the screen with the button pressed. However, this is not possible. To be able to select something, release the button, lift the pen, then put the pen down again and then select (the select tool appears).

-When the toggle mode is selected, the button press is registered only when the pen touches the screen. But the select tool is not activated and neither the previous pen tool until the pen is lifted then put back on the screen.

**Expected behavior**  

The correct behavior is the one observed under linux.

On linux, button presses on the pen are registered if the pen is close enough to the screen. When using the primary button as a temporary toggle for the selector tool, the selector tool is active on the first contact of the pen with the screen with the button pressed.

**Screenshots**  

https://github.com/flxzt/rnote/assets/115779707/816bad8a-3b81-4b1a-b6e6-51ad21f68836

**Desktop:**  
 - OS: Windows 11 22H2
 - App Version:  Rnote v0.8.2
 - Installation Source: exe from the github release page
 - Surface pro 8

Doublonmousse:
After some debugging following #832 on windows, and knowing that gtk is the one handling pen events and reporting them back to rnote, I found that 
- gtk does not see that the button is pressed until the pen touches the screen
- This bug occurs both on gtk4 and xournal++ and saw the same behavior as rnote.

Looking at the event log when the pen approaches the screen with the side button pressed : 

```
2023-10-12T07:17:19.981Z DEBUG rnote::canvas::input > handle pen event element - element: Element { pos: [[322.5976257324219, 105.57954406738281]], pressure: 0.0 }, pen_state: Proximity, event_time_delta: 0ns, modifier_keys: [], pen_mode: Some(Pen)
 2023-10-12T07:17:19.989Z DEBUG rnote::canvas::input > canvas event MotionNotify - gdk_modifiers: ModifierType(0x0), is_stylus: true
 2023-10-12T07:17:19.989Z DEBUG rnote::canvas::input > handle pen event element - element: Element { pos: [[323.49139404296875, 104.00224304199219]], pressure: 0.0 }, pen_state: Proximity, event_time_delta: 0ns, modifier_keys: [], pen_mode: Some(Pen)
[...]

Issue #243, opened 2022-08-29
Stylus buttons do not get recognized with certain HP & Wacom pens


xkevio:
**Describe the bug:**
The two extra buttons on my stylus do not get recognized correctly within Rnote even though they are configured in the settings to either switch to the eraser or selection tool.

When pressing the primary stylus button, nothing happens. When pressing the secondary stylus button, it moves the canvas or rather stops using the brush.

I tested the input with `$ libinput debug-events` and that seems to only recognize me pressing the secondary button.
`$ libinput list-devices` also does not recognize the extra eraser device whereas `xinput` does.

The stylus in question is the **HP Pen (1MR94AA)** and I use the i3-gaps compositor on Manjaro Linux. `libinput` and `libwaycom` are installed. Weirdly, the primary stylus button works in Xournal++.

`xinput` recognizes my stylus as two pointers, one pen and one eraser:
```
$ xinput                                                                                  
⎡ Virtual core pointer                    	id=2	[master pointer  (3)]
⎜   ↳ Virtual core XTEST pointer              	id=4	[slave  pointer  (2)]
⎜   ↳ SYNA328B:00 06CB:CD50 Mouse             	id=10	[slave  pointer  (2)]
⎜   ↳ SYNA328B:00 06CB:CD50 Touchpad          	id=11	[slave  pointer  (2)]
⎜   ↳ ELAN2514:00 04F3:23DD                   	id=12	[slave  pointer  (2)]
⎜   ↳ ELAN2514:00 04F3:23DD Stylus Pen (0)    	id=19	[slave  pointer  (2)]
⎜   ↳ ELAN2514:00 04F3:23DD Stylus Eraser (0) 	id=20	[slave  pointer  (2)]
```
Furthermore, the settings when looked at with `xinput list-props [id]` look as follows for both:
```
$ xinput list-props 19                                                                
Device 'ELAN2514:00 04F3:23DD Stylus Pen (0)':
	Device Enabled (178):	1
	Coordinate Transformation Matrix (180):	1.000000, 0.000000, 0.000000, 0.000000, 1.000000, 0.000000, 0.000000, 0.000000, 1.000000
	libinput Send Events Modes Available (294):	1, 0
	libinput Send Events Mode Enabled (295):	0, 0
	libinput Send Events Mode Enabled Default (296):	0, 0
	Device Node (297):	"/dev/input/event11"
	Device Product ID (298):	1267, 9181
[...]

PR #1113, opened 2024-06-07
fix: workaround for windows ink

Doublonmousse:
This is a workaround for #785 

- force down state when the element is a stylus with a pressure value that is not zero. This way when a button is pressed before the pen is on the screen, this triggers proximity mode as rnote expects a left click when the pen touches the screen

- forces a pen state to up upon a button release as once again we have our button as a buttonrelease instead of the left click or BUTTON_PRIMARY release

Tested and working on my surface pro.
For now this is implemented using `cfg` flags only for windows builds.

From testing directly the pen input seen when using the Win32 API (what gtk is seeing) and WinRT, the original issue stems from limitations of the Win32 API. I'm reasonably confident then that the issue is a general issue on Windows, hence an os-specific workaround, so that applying it for all windows build shouldn't create issues.

Not sure this would make sense to apply elsewhere (or that this workaround may be needed outside windows).

I can add a toggle somewhere in the settings toolbar (hopefully in a way that's only shown on windows) if you want. I think it makes sense though to have the workaround enabled by default (xournal++ has a similar toggle that's disabled by default and isn't easy to find or understand that it fixes the issue when encountered).

### DIFF
--- a/crates/rnote-ui/src/canvas/input.rs
+++ b/crates/rnote-ui/src/canvas/input.rs
@@ -119,6 +119,11 @@ pub(crate) fn handle_pointer_controller_event(
                 if gdk_button == gdk::BUTTON_PRIMARY {
                     pen_state = PenState::Up;
                 } else {
+                    // Workaround for https://github.com/flxzt/rnote/issues/785
+                    // On window only one button release event is sent when the
+                    // pen leaves the screen, and if the button is pressed this
+                    // is not a gdk::BUTTON_PRIMARY
+                    #[allow(clippy::collapsible_else_if)]
                     if cfg!(target_os = "windows") {
                         pen_state = PenState::Up;
                     } else {
@@ -172,6 +177,11 @@ pub(crate) fn handle_pointer_controller_event(
         for (element, event_time) in elements {
             tracing::trace!("handle pen event element - element: {element:?}, pen_state: {pen_state:?}, event_time_delta: {:?}, modifier_keys: {modifier_keys:?}, pen_mode: {pen_mode:?}", now.duration_since(event_time));
 
+            // Workaround for https://github.com/flxzt/rnote/issues/785
+            // only one event is sent when the pen approaches the screen
+            // on Windows whereas rnote expects 2 (switch to proximity
+            // then down). This forces the pen to be down when on the
+            // screen
             #[cfg(target_os = "windows")]
             {
                 if element.pressure > 0.0 && is_stylus {
--- a/crates/rnote-ui/src/canvas/input.rs
+++ b/crates/rnote-ui/src/canvas/input.rs
@@ -119,12 +119,9 @@ pub(crate) fn handle_pointer_controller_event(
                 if gdk_button == gdk::BUTTON_PRIMARY {
                     pen_state = PenState::Up;
                 } else {
-                    #[cfg(target_os = "windows")]
-                    {
+                    if cfg!(target_os = "windows") {
                         pen_state = PenState::Up;
-                    }
-                    #[cfg(not(target_os = "windows"))]
-                    {
+                    } else {
                         pen_state = PenState::Proximity;
                     }
                 }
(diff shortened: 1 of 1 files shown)


crates/rnote-ui/src/canvas/input.rs
```diff
@@ -119,7 +119,14 @@ pub(crate) fn handle_pointer_controller_event(
                 if gdk_button == gdk::BUTTON_PRIMARY {
                     pen_state = PenState::Up;
                 } else {
-                    pen_state = PenState::Proximity;
+                    #[cfg(target_os = "windows")]
```
flxzt, line 122: Consider using [cfg!](https://doc.rust-lang.org/stable/std/macro.cfg.html) and a `if {} else {}`

flxzt:
I haven't had issues on windows so I can't test this, but it looks like a reasonable fix

> I can add a toggle somewhere in the settings toolbar (hopefully in a way that's only shown on windows) if you want.

Oh no please not. For fixes we should ensure it actually works and do a lot of testing instead of burdening the user with this.

Doublonmousse:
So, the original reason for the bug is this 
https://github.com/xournalpp/xournalpp/issues/2316#issuecomment-860083264

Now I have verified that this is what's happening on my end, and that it also would be possible to make it work correctly with some WinRT APIs instead of Win32 ones for the pens (I've tested this with the windows and windows-sys crate though doing this on gtk's code may be harder).

The thing I don't know is the following
- does it affect every windows PC ?
Did you not encounter any similar issue on Windows yourself ? I'd be interested to know on what hardware.

Seeing that there are technically different protocols (MPP, AES, USI with MPP being the most widely used) maybe there are slight differences.

I don't see a lot of windows-related input issues on this repo (https://github.com/flxzt/rnote/issues/243#issuecomment-1732542608 is on windows but buttons seems to register, the pen uses AES by default + extra drivers)

From my experience (me, some people that tried rnote, or some other reports on reddit) this issue affects at least all surface products.

crates/rnote-ui/src/canvas/input.rs
```diff
@@ -119,7 +119,11 @@ pub(crate) fn handle_pointer_controller_event(
                 if gdk_button == gdk::BUTTON_PRIMARY {
                     pen_state = PenState::Up;
                 } else {
-                    pen_state = PenState::Proximity;
+                    if cfg!(target_os = "windows") {
```
flxzt, line 122: A code comment here with the reason for this windows-specific quirk would be nice (and I see that clippy will complain about a collapsible if/else, you can turn it off with `#[allow(clippy::collapsible_else_if)]`

crates/rnote-ui/src/canvas/input.rs
```diff
@@ -168,6 +172,13 @@ pub(crate) fn handle_pointer_controller_event(
         for (element, event_time) in elements {
             tracing::trace!("handle pen event element - element: {element:?}, pen_state: {pen_state:?}, event_time_delta: {:?}, modifier_keys: {modifier_keys:?}, pen_mode: {pen_mode:?}", now.duration_since(event_time));
 
+            #[cfg(target_os = "windows")]
+            {
+                if element.pressure > 0.0 && is_stylus {
+                    pen_state = PenState::Down;
+                }
+            }
+
```
flxzt, line 181: Here too - let's document it also in code

crates/rnote-ui/src/canvas/input.rs
Doublonmousse: fixed

crates/rnote-ui/src/canvas/input.rs
Doublonmousse: fixed

flxzt:
let's try it, I think this is ready and can be merged. The proper fix to avoid this workaround would be in Gtk to provide a consistent event stream between all platforms and tablet technologies

Doublonmousse:
Yes, it would have to be done in gtk https://gitlab.gnome.org/GNOME/gtk/-/issues/6723 probably with a similar approach to this https://github.com/Doublonmousse/winit/tree/winit-pen though I'm not sure I'm ready to deal with potential issues porting this to gtk (according to this https://github.com/microsoft/WindowsAppSDK/issues/233). 

Maybe I'm worrying for nothing and things have changed between 2020 and now and it would be possible to do this without too much issues (not sure gtk uses any WinRT API but if they do or start to do, it'll be possible to fix relatively easily).

flxzt:
Alright, good to have this documented for future reference.

The workaround itself lgtm.

PR #1115, opened 2024-06-08
fix: filters with native file pickers and forcing file extensions

### LINKED ISSUE #1114: file dialog filters not working as expected
**Describe the bug**  

After #1075, filters still don't work as expected.

On windows, applying the filters make gtk not even use the windows native file picker https://gitlab.gnome.org/GNOME/gtk/-/issues/6717.

On linux, the filter appears but you can still save a file without the file extension. The filename portion is the only part selected so it's harder to achieve but is still possible.
image

**Expected behavior**  
Have the filter work in a way that makes it impossible to have a `filename` saved without any extensions and using the native file picker on all platforms.

Libreoffice works around this by adding themselves the `.odt` if the extension is not there in the file picker
Screenshot 2024-06-08 105938

One issue on the gtk repo talks about this issue as well https://gitlab.gnome.org/GNOME/gtk/-/issues/5367

To have something that uses the native file picker everywhere, no mime type nor custom filters can be used. Now the question is whether using [add_pattern](https://gtk-rs.org/gtk3-rs/stable/latest/docs/gtk/struct.FileFilter.html#method.add_pattern) over [add_mime_type](https://gtk-rs.org/gtk3-rs/stable/latest/docs/gtk/struct.FileFilter.html#method.add_mime_type) will work

 - OS: Windows 11 23H2, fedora 40
 - App Version: rnote main (post #1075)
 - Installation Source: from source
 - Desktop Environment: KDE 6 on linux
 - Display Server: wayland on linux

Doublonmousse:
this allows for the filepicker to stay as the native one on windows and, this makes the extension automatically apply on both windows and linux even if the file name doesn't include the extension.

To fix #1114

Some notes 
- I think only having one of `filter.add_pattern` and `filter.add_suffix` is enough for it to work
- for import, we have to remove the mimetype for the filter to apply with the native file picker on windows but it's a little more restrictive on imports when used on linux. If you want I can set a conditional compilation flag to keep the mimetypes on linux (or at least on non windows targets)
- for the export of doc pages I actually removed the previous fix, but in this case adding a mimetype worked on windows. Also, as it is fundamentally a choice of folders, I don't understand why jpg/pdf/svg filters are added there, seems like an error.
- Tested on windows and linux but not yet on mac os.

### DIFF
--- a/crates/rnote-ui/src/dialogs/export.rs
+++ b/crates/rnote-ui/src/dialogs/export.rs
@@ -30,6 +30,9 @@ pub(crate) async fn dialog_save_doc_as(appwindow: &RnAppWindow, canvas: &RnCanva
     } else {
         filter.add_mime_type("application/rnote");
     }
+    if cfg!(target_os = "macos") {
+        filter.add_suffix("rnote");
+    }
     filter.set_name(Some(&gettext(".rnote")));
 
     // create the list of filters
@@ -292,6 +295,9 @@ fn create_filedialog_export_doc(
             } else {
                 filter.add_mime_type("image/svg+xml");
             }
+            if cfg!(target_os = "macos") {
+                filter.add_suffix("svg");
+            }
             filter.set_name(Some(&gettext("Svg")));
         }
         DocExportFormat::Pdf => {
@@ -300,6 +306,9 @@ fn create_filedialog_export_doc(
             } else {
                 filter.add_mime_type("application/pdf");
             }
+            if cfg!(target_os = "macos") {
+                filter.add_suffix("pdf");
+            }
             filter.set_name(Some(&gettext("Pdf")));
         }
         DocExportFormat::Xopp => {
@@ -307,6 +316,9 @@ fn create_filedialog_export_doc(
                 filter.add_pattern("*.xopp");
             } else {
                 filter.add_mime_type("application/x-xopp");
+            }
+            if cfg!(target_os = "macos") {
+                filter.add_suffix("xopp");
             }
             filter.set_name(Some(&gettext("Xopp")));
         }
@@ -596,8 +608,6 @@ fn create_filedialog_export_doc_pages(
 
     filedialog.set_initial_folder(get_initial_folder_for_export(appwindow, canvas).as_ref());
 
-    // does it work with mimetype but not the previous thing ?
-
     let filter = FileFilter::new();
     // We always need to be able to select folders
     filter.add_mime_type("inode/directory");
@@ -608,6 +618,9 @@ fn create_filedialog_export_doc_pages(
             } else {
                 filter.add_mime_type("image/svg+xml");
             }
+            if cfg!(target_os = "macos") {
+                filter.add_suffix("svg");
+            }
             filter.set_name(Some(&gettext("Svg")));
         }
         DocPagesExportFormat::Png => {
@@ -615,6 +628,9 @@ fn create_filedialog_export_doc_pages(
                 filter.add_pattern("*.png");
             } else {
                 filter.add_mime_type("image/png");
+            }
+            if cfg!(target_os = "macos") {
+                filter.add_suffix("png");
             }
             filter.set_name(Some(&gettext("Png")));
         }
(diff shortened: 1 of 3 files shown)


flxzt:
I would like to keep the MIME-type based filtering on linux so this is probably better implemented as a windows-specific workaround

crates/rnote-ui/src/dialogs/export.rs
```diff
@@ -578,32 +600,15 @@ fn create_filedialog_export_doc_pages(
...
-            filter.add_suffix("jpeg");
-            filter.set_name(Some(&gettext("Jpeg")));
-        }
-    }
-
-    let filter_list = gio::ListStore::new::<FileFilter>();
-    filter_list.append(&filter);
-    filedialog.set_filters(Some(&filter_list));
-
+    // note : we revert https://github.com/flxzt/rnote/pull/1075
```
flxzt, line 608: I think some got accidentally deleted here?

Doublonmousse: Tell me if I'm wrong but for me this opens a folder chooser, so file filters aren't used

flxzt: Right, I think my reasoning when implementing this was that all files that could possibly be overwritten by the export would be visible. Picking fails when not a folder is selected so it should be fine to add these the files filter

Doublonmousse: Ok, I'll try to put the code back and adapt it (I think I need some more `cfg` directives to make it work)

crates/rnote-ui/src/dialogs/export.rs
```diff
@@ -421,7 +451,7 @@ pub(crate) async fn dialog_export_doc_pages_w_prefs(appwindow: &RnAppWindow, can
                         }
                     }
                     Err(e) => {
-                        tracing::debug!("Did not export document pages (Error or dialog dismissed by user), Err: {e:?}");
+                        tracing::debug!("mac suffix consistency VERSION Did not export document pages (Error or dialog dismissed by user), Err: {e:?}");
```
flxzt, line 454: What's with this message?

Doublonmousse: Oh sorry, that's me changing debug message to be sure that I actually test the right version. I've had issues before with the install not working or the `rnote` command not launching the version I just compiled after compiling it and installing. It was not really meant to be pushed, I removed it

flxzt: ok, no problem!

crates/rnote-ui/src/dialogs/export.rs
```diff
@@ -872,21 +915,42 @@ fn create_filedialog_export_selection(
...
-            filter.add_suffix("jpg");
-            filter.add_suffix("jpeg");
+            if cfg!(target_os = "windows") {
+                filter.add_pattern("*.jpg");
+                filter.add_pattern("*.jpeg");
+            } else {
+                filter.add_mime_type("image/jpeg");
+            }
+            if cfg!(target_os = "macos") {
+                filter.add_suffix("jpg");
```
flxzt, line 952: Here the `jpeg` suffix is missing

crates/rnote-ui/src/dialogs/export.rs
```diff
@@ -583,27 +613,40 @@ fn create_filedialog_export_doc_pages(
...
+                filter.add_suffix("png");
+            }
             filter.set_name(Some(&gettext("Png")));
         }
         DocPagesExportFormat::Jpeg => {
-            filter.add_mime_type("image/jpeg");
-            filter.add_suffix("jpg");
-            filter.add_suffix("jpeg");
+            if cfg!(target_os = "windows") {
+                filter.add_pattern("*.jpg");
```
flxzt, line 639: Here `jpeg` is missing as well (and in :644)

flxzt: This one is still missing.

flxzt:
alright, looks good now!

Issue #1048, opened 2024-03-15
Add settings for "Vertical space" tool in "Infinite" document layout


anefil:
When document layout is set to "Infinite" or "Semi infinite", the tool should have options to work only on one vertical line of pages.

**Is your feature request related to a problem? Please describe.**  
I have a document with every homework assignment written in one column. I would like to use the tool for editing one vertical line of pages only.

**Describe the solution you'd like**  
Add tool-specific settings, like on the brushes or on the eraser.

**Describe alternatives you've considered**  
Using box selection provides difficulties, as I need to zoom out and accurately  select the pages. It is unpleasant

![image](https://github.com/flxzt/rnote/assets/95078518/f35033a8-b79e-464c-8509-92e4001f2448)

Doublonmousse:
Fixed by #1116

Issue #878, opened 2023-11-13
Option for creating vertical whitespace per column of pages


tipptop3d:
Currently the vertical whitespace tool moves everything below the line, but i can imagine that sometimes, if you write over multiple columns, you would only want to adjust the current column.

flxzt:
to clarify - you're talking about selecting two vertical positions and then moving only the content between them up and down?

tipptop3d:
Not really. I like your "semi infinite" layout to split my work horizontally onto multiple "columns", as i still need to export them to a pdf.  Maybe my sketch helps ;D I mean only moving objects that are below the red line, not over the complete width, constrained by the format borders.
![rnote](https://github.com/flxzt/rnote/assets/45404249/82c27b8d-f8a6-4992-8d2f-6b5845617bef)

flxzt:
ahh I see. That would be quite useful indeed, I agree.

tipptop3d:
I thought more of a checkbox or different mode under the tools menu, that if selected moves only a single column, but yes, your solution would allow a more flexible/wider selection. I would also be fine with that.

Doublonmousse:
Somehow similar to #1048

Doublonmousse:
I was thinking of always displaying the borders (with some external padding) and clicking on that padding would change the spacing, whilst clicking between the borders would activate the tool.

I'd say changing for this tool should display the borders in a way where they're visible and clicking again should reset them to make the borders be in view again.

Doublonmousse:
Partially fixed by #1116.

The borders are the page borders, there isn't a more flexible way to choose the borders

PR #1116, opened 2024-06-08
feat: new options for the vertical space tool

Doublonmousse:
Adds the following options to the vertical space bar
image

For #1048 and #878

- [x] make the two last options work
- [x] adapt the gui part of the tool to only go as far as the page widths selected


Doublonmousse:
I think I've done everything except doing the more complex tool with borders that can be edited. I wonder if it's not better to do this later in a separate PR.

flxzt:
I like this a lot from a functionality POV.

Some suggestions for improvements in the UI:
I don't like the location of the added options.
It would be better to place this right next to the tool
and rearrange the tools sidebar with visible separators between the individual tools.
This would also be consistent to the other sidebar pages.

There are also some minor inconsistencies like the option button size which I think should be the smaller variant.

Also: the option names could improved imho. What about instead of "Shape Configuration" -> " "Vertical-Space Tool Options", "Respect Vertical/Horizontal Borders" -> "Limit movement to Vertical/Horizontal Page Borders" ?

Doublonmousse:
Screenshot 2024-08-21 at 00 17 29

I've went for the linked style of other toolspages for now.. I think it's a little better at showing that there are three different tools than before.

It doesn't really fix the fact that the settings icon isn't related to the vertical space tool though.
Ideally I'd like to keep the linked aspect of the three tools whilst making the settings icon appear tied to the vertical space tool

flxzt:
What do you think about something like this?

grafik

Possibly with separators between the individual tools as well

Doublonmousse:
Yes, that works well !

flxzt:
alright, I'll add my commit then.

Oh and I am not sure the force snapping option is actually needed, since it is already covered by the "snap positions" option. It could be confusing when deactivating one of them, why it seemingly is not respected

Doublonmousse:
Yeah, makes sense.
It's more of a habit I took since https://github.com/flxzt/rnote/discussions/721

I think it makes sense to have a second snap mode where displacement are multiple of the grid size as opposed to snapping elements to the grid (usecase : select some text and change its position so that it stays aligned to the grid). But, same thing as before, it makes more sense to do that as a global setting than this

flxzt:
Yeah. I am a bit concerned for that option because of feature creep of more "niche" use cases.

Can be a follow-up?

Doublonmousse:
Yes, I agree it makes sense to remove the "force snapping" option from the current PR.

And I'll see if I want to do a follow up for snapping options later (but there is probably less niche things to add to snapping before I do that)

crates/rnote-compose/src/verticaltoolconfig.rs
```diff
@@ -0,0 +1,10 @@
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+/// configuration for the vertical tool
+pub struct VerticalToolConfig {
+    /// horizontal limit
+    pub horizontal_border: bool,
```
flxzt, line 7: I'd rename these fields to `limit_movement_<horizontal/vertical>_borders`

crates/rnote-engine/src/store/stroke_comp.rs
```diff
@@ -640,6 +640,64 @@ impl StrokeStore {
             .collect::<Vec<StrokeKey>>()
     }
 
+    pub(crate) fn keys_between(
```
flxzt, line 643: Can this function be split up into two separate ones that are only responsible for x or y? And let the keys be supplied by the user for composing them together to get the same result.

This would get rid of the complexity (match statement) in the function body and would make it more re-usable.

crates/rnote-ui/data/ui/penssidebar/._toolspage.ui
flxzt, line 1: I think this file was added by mistake

crates/rnote-ui/src/penssidebar/toolspage.rs
```diff
@@ -15,6 +18,16 @@ mod imp {
...
+        #[template_child]
+        pub(crate) verticaltool_menubutton: TemplateChild<MenuButton>,
+        #[template_child]
+        pub(crate) verticaltool_popover: TemplateChild<Popover>,
+        #[template_child]
+        pub(crate) verticaltool_popover_close_button: TemplateChild<Button>,
+        #[template_child]
+        pub(crate) verticalspace_respect_vertical_bordersrow: TemplateChild<adw::SwitchRow>,
+        #[template_child]
+        pub(crate) verticalspace_respect_horizontal_bordersrow: TemplateChild<adw::SwitchRow>,
```
flxzt, line 30: I'd rename these fields in the following way: `verticaltool` -> `verticalspace` and respect_<horizontal/vertical>_bordersrow` -> `limit_movement_<horizontal/vertical>_borders_row`

crates/rnote-compose/src/lib.rs
```diff
@@ -31,6 +31,8 @@ pub mod style;
 pub mod transform;
 /// other misc utilities
 pub mod utils;
+/// vertical tool options
+pub mod verticaltoolconfig;
```
flxzt, line 35: Why is this in rnote-compose? I think it should be in rnote-engine under `pensconfig` (and renamed to verticalspacetoolconfig`

crates/rnote-engine/src/store/stroke_comp.rs
Doublonmousse: I've not done exactly that for now but I got rid of the match statement.
It's replaced with multiple filters using the bools for the state of the verticalspace tool to skip unneeded selection using the lazily evaluated or operator.

I've not tried to chain the filters, that'd need for the `stroke_components` to be public (or at least the iterator over them through a getter method) + two external functions defined on the `impl Iterator<Item  = (Strokekey,...)>` (can't be a method on self).

But at that point, I can also get the iterator from the new getter, and do the filters inside `toolspage.rs` (and maybe lazily evaluate some of the values I'm for now passing as parameters to `keys_between`)

Doublonmousse:
The CI check is giving warnings for a line that doesn't exist ?
```
 warning: unused import: `tracing::error`
  --> crates/rnote-engine/src/store/stroke_comp.rs:15:5
   |
15 | use tracing::error;
   |     ^^^^^^^^^^^^^^
   |
   = note: `#[warn(unused_imports)]` on by default
```
But there's nothing 
https://github.com/Doublonmousse/rnote/blob/vertical_tool/crates/rnote-engine/src/store/stroke_comp.rs#L15

crates/rnote-engine/src/store/stroke_comp.rs
flxzt: Check out `filter_keys_intersecting_bounds()`, that's what I had in mind.

Wouldn't something like this be possible?

```rust
let keys = store.keys_unordered();
if limit_movement_vertical_border {
    let filtered = store.keys_between_vertical_position(&keys);
    // ...
} else {
    let filtered = store.keys_below_y()
    // ...
}

flxzt:
I noticed it as well. Seems to be a false positive lint.

One additional point, which I am not sure about:

Should strokes that only intersect the limits be included as well or not?

crates/rnote-engine/src/store/stroke_comp.rs
flxzt: But I won't block merging this if it isn't done this way, so your call

Doublonmousse: I'll try to do the method used in `filter_keys_intersecting_bounds()` and see how that compares. 

One question I have for this (and other related operations), is that I'm not sure we're using the Rtree everywhere for queries (there's a `keys_in_bounds` method that seems like a good candidate for doing this instead, unless there's issue if we have semi infinite bounds ?)

Doublonmousse:
I think it's better to have all selection tools work when the full stroke is in bounds, and add a global option for partial selection like #155 for all tools.

Do you think adding a graphical indicator when limiting the selection horizontally to show where the selection stops is a good idea ? I also wondered if I should do something in that case to make it harder to accidentally overlap strokes onto the next page (like some snapping or indicator when the selection would touch the next horizontal border)

crates/rnote-engine/src/store/stroke_comp.rs
flxzt: yeah that's true `keys_in_bounds` could be used instead!

flxzt:
At least from my testing it was intuitive enough without an indicator.

Snapping sound like a good idea, I can see it being useful often

Doublonmousse:
I'm not sure snapping is a super good idea. I don't know how to deal with potentially two snapping modes at the same time (General snapping to grid + the added one when crossing the border)

Maybe it makes more sense to limit the amplitude of the tool so that the selected portion stays in the same page instead. This way I don't have the issue with multiple snapping modes, and we don't end up with strokes that become unmovable by the vertical space tool when using the limiter on horizontal borders (because of the decision to only include strokes that are fully in bounds).

crates/rnote-engine/src/store/stroke_comp.rs
flxzt: Ok, this is fine for me now

flxzt:
Couldn't there be both? Just like content already also snaps to the page borders when `Snap positions` is enabled.

flxzt:
But implementing snapping is also not a blocker for me

Doublonmousse:
Oh I see, that's indeed possible to have two snapping modes.

That would make placing strokes between pages when using the horizontal limit less of an issue.

Maybe that can be a follow up ? I wonder if it's better to select strokes that intersect the bounds in this PR in the meantime

flxzt:
> Maybe that can be a follow up ? I wonder if it's better to select strokes that intersect the bounds in this PR in the meantime

Of course.

I just tested it again, and including intersecting strokes feels more intuitive, so I'd say yes.

But I found one remaining issue: When both limits are enabled in the UI, the vertical option actually is not respected.
There seems to be a logic error in the implementation

Doublonmousse:
I think it's working now

flxzt:
Looks good now, nice!

PR #1117, opened 2024-06-08
 feat: Open files/folder in the file explorer

### LINKED ISSUE #1053: Option to open Workspace in file manager

Doublonmousse:
Reopening #1089 after a github mess...

To fix #1053 

I switched to opener and added the action to open the workspace folder in the folder submenu.
This is a PR that's very similar to https://github.com/flxzt/rnote/pull/868, using a slightly different crate (opener) that has a reveal call and a MIT OR Apache-2.0 license.

The implementation detail for this is here https://docs.rs/opener/latest/opener/fn.reveal.html
The open crate doesn't have this option afaik but having two crates doing very similar things is slightly redundant

Tested only on linux aarch64 for now (but should be supported on all platforms)

image

One thing to verify is that opener does not have the same [open](https://docs.rs/open/5.1.4/open/index.html)::[that_detached](https://docs.rs/open/5.1.4/open/fn.that_detached.html#) option, only [opener](https://docs.rs/opener/latest/opener/index.html)::[open](https://docs.rs/opener/latest/opener/fn.open.html#) and I don't know if that makes a difference or not

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -358,6 +358,17 @@ dependencies = [
 ]
 
 [[package]]
+name = "bstr"
+version = "1.9.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "05efc5cfd9110c8416e471df0e96702d58690178e206e61b7173706673c93706"
+dependencies = [
+ "memchr",
+ "regex-automata 0.4.6",
+ "serde",
+]
+
+[[package]]
 name = "bumpalo"
 version = "3.16.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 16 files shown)


flxzt:
> One thing to verify is that opener does not have the same [open](https://docs.rs/open/5.1.4/open/index.html)::[that_detached](https://docs.rs/open/5.1.4/open/fn.that_detached.html#) option, only [opener](https://docs.rs/opener/latest/opener/index.html)::[open](https://docs.rs/opener/latest/opener/fn.open.html#) and I don't know if that makes a difference or not

Interesting - the [docs](https://docs.rs/opener/0.7.1/opener/fn.open.html) and the implementation suggest that it doesn't take care of being non-blocking nor being detached from the rnote process. I think on my Gnome installation it ends up being okay because it [launches through dbus](https://github.com/Seeker14491/opener/blob/ef8c5e4f3c458f656fba477965ecd7e30d26c9f5/opener/src/freedesktop.rs#L26) but for other systems/file-managers this might be an issue. Need to test it more with different distributions.

flxzt:
I may have changed my mind - looks like opener in general does not block, but there are no guarantees. And since we definitely don't want to block the `open::that_detached` function seems like the better choice

Doublonmousse:
Okay, then I'll revert the change and use open instead.

I'd still like to have the reveal feature and I'm not 100% convinced of the one used currently. At least on windows it opens a new explorer no matter what even if another window revealing the same file is open. Pretty sure that's not how that works in vscode.

Issue #1120, opened 2024-06-15
Port `GtkDialog`s to `AdwDialog`s


flxzt:
as discussed #1110 porting `GtkDialog` to `AdwDialog` needs additional effort because actions need to re-implemented manually.

PR #1124, opened 2024-06-16
ui: migrate `GtkDialog` to `AdwDialog`

zefr0x:
Closes: #1120
Related: #1110

- In the UI i used ~`GtkBox`~ [`AdwToolbarView`](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.ToolbarView.html) and `AdwHeaderBar` to replace the actions functionality that was in `GtkDialog`. (its a minor change, but Github's diff algorithm doesn't handle .xml files properly because of the change in the indentation)

- Rather than hiding the dialog when the file selection window is opened, the `sensitive` property of the dialog will be switched to `false`. It will be returned to `true` when the file selection window is closed.

- The import related dialogs are more complicated since we need to return a value.

For now just file-selection, print, and keyboard-shortcut dialogs opens in a separate window. I think we don't have a choice for the first two, but the last one is possible (if there was `AdwShortcutsDialog`).

---

P.S. I'm not confident with my async skills so I might've done some thing wrong :)

### DIFF
--- a/crates/rnote-ui/src/dialogs/export.rs
+++ b/crates/rnote-ui/src/dialogs/export.rs
@@ -1,5 +1,4 @@
-// gtk4::Dialog is deprecated, but the replacement adw::ToolbarView is not suitable for a async flow
-#![allow(deprecated)]
+//adw::ToolbarView is a replacement for adw::Dialog but not suitable for an async flow
 
 // Imports
 use crate::canvas::{self, RnCanvas};
@@ -7,9 +6,7 @@ use crate::RnStrokeContentPreview;
 use crate::{config, RnAppWindow};
 use adw::prelude::*;
 use gettextrs::gettext;
-use gtk4::{
-    gio, glib, glib::clone, Builder, Button, Dialog, FileDialog, FileFilter, Label, ResponseType,
-};
+use gtk4::{gio, glib, glib::clone, Builder, Button, FileDialog, FileFilter, Label};
 use num_traits::ToPrimitive;
 use rnote_compose::SplitOrder;
 use rnote_engine::document::Layout;
@@ -89,7 +86,7 @@ pub(crate) async fn dialog_export_doc_w_prefs(appwindow: &RnAppWindow, canvas: &
     let builder = Builder::from_resource(
         (String::from(config::APP_IDPATH) + "ui/dialogs/export.ui").as_str(),
     );
-    let dialog: Dialog = builder.object("dialog_export_doc_w_prefs").unwrap();
+    let dialog: adw::Dialog = builder.object("dialog_export_doc_w_prefs").unwrap();
     let button_confirm: Button = builder.object("export_doc_button_confirm").unwrap();
     let with_background_row: adw::SwitchRow =
         builder.object("export_doc_with_background_row").unwrap();
@@ -101,10 +98,12 @@ pub(crate) async fn dialog_export_doc_w_prefs(appwindow: &RnAppWindow, canvas: &
     let export_file_label: Label = builder.object("export_doc_export_file_label").unwrap();
     let export_file_button: Button = builder.object("export_doc_export_file_button").unwrap();
     let preview: RnStrokeContentPreview = builder.object("export_doc_preview").unwrap();
+    let export_doc_button_cancel: Button = builder.object("export_doc_button_cancel").unwrap();
+    let export_doc_button_confirm: Button = builder.object("export_doc_button_confirm").unwrap();
 
     let initial_doc_export_prefs = canvas.engine_ref().export_prefs.doc_export_prefs;
     let doc_layout = canvas.engine_ref().document.layout;
-    dialog.set_transient_for(Some(appwindow));
+    dialog.present(appwindow);
 
     // initial widget state with the preferences
     let selected_file: Rc<RefCell<Option<gio::File>>> = Rc::new(RefCell::new(None));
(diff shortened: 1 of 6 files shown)


oscfdezdz:
> In the UI i used `GtkBox` and `AdwHeaderBar` to replace the actions functionality that was in `GtkDialog`.

I would suggest using [AdwToolbarView](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/main/class.ToolbarView.html) instead of `GtkBox` which is a more specific widget for this type of use because of the properties and styles it has.

crates/rnote-ui/src/dialogs/import.rs
```diff
@@ -266,21 +268,69 @@ pub(crate) async fn dialog_import_pdf_w_prefs(
...
+            import_pdf_button_cancel.emit_clicked();
+
+            glib::Propagation::Stop
+        }))),
+    ));
+    dialog.add_controller(controller);
+
+    // Wait for a response from the dialog
+    loop {
+        if let Some(res) = rx.next().await {
```
flxzt, line 332: Is the loop really needed here? When the channel resolves to `None` the sender side has been closed, and we should return as well.

crates/rnote-ui/src/dialogs/import.rs
```diff
@@ -310,17 +362,66 @@ pub(crate) async fn dialog_import_xopp_w_prefs(
...
+            import_xopp_button_cancel.emit_clicked();
+
+            glib::Propagation::Stop
+        }))),
+    ));
+    dialog.add_controller(controller);
+
+    // Wait for a response from the dialog
+    loop {
+        if let Some(res) = rx.next().await {
```
flxzt, line 423: Same here

crates/rnote-ui/src/dialogs/import.rs
zefr0x: Is there any type of error that I should return in this case?

flxzt: I guess an anyhow Error with a descriptive message should suffice. It only ends up being printed as a log message anyway

crates/rnote-ui/src/dialogs/export.rs
```diff
@@ -682,7 +695,7 @@ pub(crate) async fn dialog_export_selection_w_prefs(appwindow: &RnAppWindow, can
     export_file_button.connect_clicked(
         clone!(@strong selected_file, @weak export_file_label, @weak button_confirm, @weak dialog, @weak canvas, @weak appwindow => move |_| {
             glib::spawn_future_local(clone!(@strong selected_file, @weak export_file_label, @weak button_confirm, @weak dialog, @weak canvas, @weak appwindow => async move {
-                dialog.hide();
+                dialog.set_sensitive(false);
```
flxzt, line 698: Shouldn't this be [set_visible](https://gtk-rs.org/gtk4-rs/stable/0.8/docs/gtk4/prelude/trait.WidgetExt.html#method.set_visible) ?

zefr0x: From a UX standpoint in this step we are still interacting with the dialog so it shouldn't be hidden. Also even if we did hide it we can't interact with the canvas anyway (we will have a non-intractive main window without an indication).

By having `sensitive` set to `false` we have an indication that we can't interact with the dialog or the canvas while we are selecting a file.


It might make since with the old `gtk::Dialog` since it was a separate window (it was like we did replace the dialog with a file-selection window). But with `adw::Dialog` it is bad for UX.

flxzt: IIRC it was an issue when the modal gtk dialog had the file picker stacked on top of it. But good point, adw::Dialog is not a separate Window anymore

crates/rnote-ui/src/dialogs/import.rs
```diff
@@ -296,11 +347,13 @@ pub(crate) async fn dialog_import_xopp_w_prefs(
...
     );
-    let dialog: Dialog = builder.object("dialog_import_xopp_w_prefs").unwrap();
+    let dialog: adw::Dialog = builder.object("dialog_import_xopp_w_prefs").unwrap();
     let dpi_row: adw::SpinRow = builder.object("xopp_import_dpi_row").unwrap();
     let xopp_import_prefs = canvas.engine_ref().import_prefs.xopp_import_prefs;
+    let import_xopp_button_cancel: Button = builder.object("import_xopp_button_cancel").unwrap();
+    let import_xopp_button_confirm: Button = builder.object("import_xopp_button_confirm").unwrap();
 
-    dialog.set_transient_for(Some(appwindow));
+    dialog.present(appwindow);
```
flxzt, line 356: Move this down to right before awaiting the channel receiver, so the handlers are already set up before the dialog is presented.

crates/rnote-ui/src/dialogs/import.rs
```diff
@@ -122,8 +122,10 @@ pub(crate) async fn dialog_import_pdf_w_prefs(
         builder.object("pdf_import_page_borders_row").unwrap();
     let pdf_import_adjust_document_row: adw::SwitchRow =
         builder.object("pdf_import_adjust_document_row").unwrap();
+    let import_pdf_button_cancel: Button = builder.object("import_pdf_button_cancel").unwrap();
+    let import_pdf_button_confirm: Button = builder.object("import_pdf_button_confirm").unwrap();
 
-    dialog.set_transient_for(Some(appwindow));
+    dialog.present(appwindow);
```
flxzt, line 128: Same here

crates/rnote-ui/src/dialogs/mod.rs
```diff
@@ -440,9 +440,15 @@ pub(crate) async fn dialog_edit_selected_workspace(appwindow: &RnAppWindow) {
...
+    let edit_selected_workspace_button_cancel: Button = builder
+        .object("edit_selected_workspace_button_cancel")
+        .unwrap();
+    let edit_selected_workspace_button_apply: Button = builder
+        .object("edit_selected_workspace_button_apply")
+        .unwrap();
 
     preview_row.init(appwindow);
-    dialog.set_transient_for(Some(appwindow));
+    dialog.present(appwindow);
```
flxzt, line 451: Same here

flxzt:
LGTM, thank you for your contribution. :)

Issue #1061, opened 2024-03-29
Constant crashes when attempting to draw on canvas since upgrade to GNOME 45


fotnite-vevo:
**Describe the bug**  
Nearly any action (attempting to draw on the canvas, attempting to move copied in photo) will potentially crash the app, and occasionally crash GNOME Shell. Only occurs when running app under Wayland (never crashes under XWayland)

**To Reproduce**  
1. Attempt to draw on canvas, attempt to move object, etc.

**Expected behavior**  
Rnote doesn't crash

**Console Output**  
```
2024-03-29T23:05:52.845016Z DEBUG rnote: .. tracing subscriber initialized.
rnote: ../cairo/src/cairo-pattern.c:3831: _cairo_pattern_get_extents: Assertion `status == CAIRO_STATUS_SUCCESS' failed.
fish: Job 1, 'RUST_LOG=rnote=debug RUST_BACKT…' terminated by signal SIGABRT (Abort)
```

**Desktop (please complete the following information):**  
 - OS: Arch Linux
 - App Version: Rnote v0.10.1 (also applies to prior versions)
 - Installation Source: Arch Linux Community
 - Desktop Environment: GNOME 46.0
 - Display Server: Wayland
 - Input Source: Mouse and Keyboard, Tablet

fotnite-vevo:
Alright, I did a bit more testing, and it seems to be **related to fractional scaling.** I usually have fractional scaling set to 1.25x, I still get crashes on 1.5x, but not on 1x, 1.75x, or 2x. but Not sure if it's an upstream bug or not but it seems like it might be.

Doublonmousse:
Gnome 46 means running on newer gtk (4.14) and libadwaita (1.5) versions.

gtk 4.14 made `ngl` the default renderer and that seems like the most likely suspect (see https://gitlab.gnome.org/GNOME/gtk/-/issues/?sort=created_date&state=opened&search=ngl&first_page_size=20)

Maybe follow https://gitlab.gnome.org/GNOME/gtk/-/issues/6574#note_2063669 and set `GSK_RENDERER=gl` or to another renderer to test if ngl is the issue (and report issues upstream to the gtk gitlab).

It will be a while before we can seriously look at gtk 4.14 though (probably need to wait until the new versions are available on all platforms
- mac : brew, up to date
- windows : msys, not up to date but seems close
[...]

PR #1132, opened 2024-07-03
Force the renderer to be `gl`

Doublonmousse:
Workaround for #1061, forcing the renderer to always be `gl`


flxzt:
I usually don't like hardcoding things like that, but given the issues that are coming in the bugtracker this might be the most sensible thing to do currently.

Since some users also get crashes on various distros, should we add this for linux as well?

Doublonmousse:
Oh right, I did it on the wrong line. 
Now it's applied for all os (with a note refering to the issue)

PR #1134, opened 2024-07-04
fix: two-finger long-press gesture causing crashes on certain systems

### LINKED ISSUE #810: Crash/Freeze using it on a touchscreen with a stylus (Hyprland)
**Describe the bug**  
When using RNote on my HP Envy Convertible with an HP Stylus Pen, the program often randomly crashes or becomes partly unresponsive. I have uploaded a video to demonstrate both cases, since it's like a 50/50 chance of which one happens:

**To Reproduce**  
I cannot reliably reproduce the error, it happens somewhat randomly. The most reliable way for me is to bring the stylus close to the screen periodically and zoom/pan around. This video shows both cases happening (sorry for the sketchy recording, I had to hold my phone while doing this to record haha):

https://github.com/flxzt/rnote/assets/55491906/c6687be1-d6a0-4d1a-9b65-4291c98f961a

The first time, the program just crashes to desktop, the second time it freezes, not allowing me to pan/zoom anymore (I can still tap on the buttons though). The latter seems like some faulty palm rejection not detecting the stylus is not there anymore. When I try similar things in e.g. Firefox by just panning around and putting the stylus near the screen, Firefox has no problem with it and continues as usual when the stylus is removed.

**Expected behavior**  
The program staying responsive and usable.

**Console Output**  
When running with flatpak + the env flags:

```
(rnote:2): Gdk-WARNING **: 23:22:46.096: Failed to read portal settings: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface “org.freedesktop.portal.Settings” on object at path /org/freedesktop/portal/desktop

(rnote:2): Gtk-WARNING **: 23:22:47.866: Creating a portal monitor failed: GDBus.Error:org.freedesktop.DBus.Error.UnknownMethod: No such interface “org.freedesktop.portal.Inhibit” on object at path **/org/freedesktop/portal/desktop
```

When running as a Arch package, I additionally get a Segmentation fault printed to the console.

 - OS: Arch Linux
 - App Version: 0.8.2
 - Installation Source: Flatpak, Arch package
 - Desktop Environment: Hyprland 0.29.1

flxzt:
this possibly fixes #810 - but it really needs testing since I am not experiencing the issue myself. I also am cautious because the removed workaround initially also solved crashes caused by Gtk.

@Ge-lx if you have the time, please take a look

### DIFF
--- a/crates/rnote-ui/src/canvaswrapper.rs
+++ b/crates/rnote-ui/src/canvaswrapper.rs
@@ -589,31 +589,6 @@ mod imp {
                         .handle_pressed_shortcut_key(ShortcutKey::TouchTwoFingerLongPress, Instant::now());
                     canvaswrapper.canvas().emit_handle_widget_flags(widget_flags);
                 }));
-
-                self.touch_two_finger_long_press_gesture.connect_end(
-                    clone!(@weak obj as canvaswrapper => move |gesture, event_sequence| {
-                        // Only deny the sequence that is actually handled.
-                        // Because this gesture is grouped with the zoom gesture, denying all
-                        // sequences within the group ( by calling `set_state()` ) might result in a segfault in certain cases
-                        if let Some(es) = event_sequence {
-                            // setting event sequences states directly is deprecated,
-                            // but it is not clear how to refactor it while not regressing the fix 4c33594 for #595
-                            #[allow(deprecated)]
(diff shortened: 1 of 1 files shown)


flxzt:
Let's go with it - I am optimistic that it will fix the broken event sequence state that causes a crash on hyprland.

PR #1138, opened 2024-07-10
Added build instructions for Arch-based distros and for Nix

JonathanGodar:
I had a hard time building on arch and think it would be nice to have  some pointers to the differences when building on arch-based distros.

However, I no longer have arch installed so I can not check if the instructions are complete. The instructions I wrote are based on what I remembered to be troublesome and also the PKGBUILD on the AUR, thus it should be accurate or at least more helpful then nothing?

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,5 @@
 .idea/
 .toggletasks.json
 *.flatpak
+.direnv
+
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -137,6 +137,23 @@ PKGBUILD](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=rnote-git) to
 see what dependencies you need to install. (The names assigned to the "depends"
 and "makedepends" variables).
 
+### Using the Nix package manager
+
+If you are running linux or darwin and have installed [Nix](https://nixos.org) then you can install all dependencies using Nix:
+
+```bash
+# This drops you in a shell with all the required dependencies available
+# Make sure your CWD is the root of the project.
+nix --extra-experimental-features 'nix-command flakes' develop
+```
+
+If you do not want to be dropped into a bash session you can use [direnv](https://github.com/direnv/direnv). Install direnv and then type
+
+```bash
+# In project root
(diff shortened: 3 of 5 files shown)


JonathanGodar:
Sorry forgot to check the diff after committing. I am not sure what caused it but something reformatted the BUILDING.md file. I can revert all the formatting changes if you would like

JonathanGodar:
I added instructions for using Nix. This also means that I added three files to the project root:
- flake.nix | This contains the nix code to specify which dependencies are needed
- flake.lock | Locks the dependencies to the correct version
- .envrc | Is needed to make direnv work. 

As with the AUR instructions the flake.nix takes its dependencies from nixpkgs, more info in the flake.nix files comments.

----
I hope that it is okay that both of the changes are in the same PR. This is because I did some restructuring to the BUILDING.md file. If you would only like the nix or the arch build instructions, ping me and I'll make a commit where I remove the one that is not wanted.

dehesselle:
Your instructions for Arch are bad advice. You definitely should not use the `meson-arch` wrapper here. That wrapper uses `/usr` as install prefix and is useful when you build packages. Software that you build and install without package manager is supposed to be installed in `/usr/local`.

And generally speaking, contributing untested changes/code is usually considered a "no go".

flxzt:
Thanks for the PR. I do have some suggestions: instead of referring to the AUR for the needed dependencies, list them directly in the README. I'd assume package maintainers actually also look in the build instructions for writing their package spec files.

> I can revert all the formatting changes if you would like

Yes, that would be nice. It can be a separate PR if you think the formatting is not ideal/should be improved

I also agree with Rene wrt to `meson-arch`, it seems to be a tool specifically for arch packaging and I don't see why meson couldn't be used directly.

And yes, please test it. If only in a VM or a container.

I can't speak to or verify the nix part of this PR, someone else would need to do that before it can be merged

JonathanGodar:
Thanks to both of you for you feedback!

I was abit of

JonathanGodar:
Thanks to both of you for your feedback!

Since I never got it to work on arch without using the arch-meson wrapper I'll remove the arch part and maybe submit in the future if I feel like getting an arch vm. 

I'll remove the Arch part and revert the formatting changes once I get some time and finish another PR I'm working on.

PR #1139, opened 2024-07-11
logic changes for selection cancellation

### LINKED ISSUE #414: Define exit behavior of the selection tool
**Description**
After using the selection tool, using the stylus anywhere on the canvas (besides the selected object) draws a stroke with the pen tool. This is quite useful to quickly return to drawing, but I find this behavior unintuitive, because there is no indication in the UI, that this will happen. In addition this behavior is different from Xournal++. In Xournal++, the first use of the pen after selecting an object doesn't draw anything, it just switches to the pen tool.  
This isn't really a bug, the question is more about, should the current behavior stay or is the behavior from Xournal++ better. Personally I like the Xournal++ behavior more, because of the UI feedback and because many users of this application probably come from Xournal++ like I did.

**To Reproduce**
1. Select something 
2. Use the pen to click on canvas to "exit" the selection tool and return to the pen tool 
Result: A stroke has been drawn where you clicked on the canvas

**Expected behavior**
No stroke created and pen tool selected like in Xournal++

Doublonmousse:
Fixes #414

- This fixes the small dots that appears on selection cancellation with the pen (whilst allowing for strokes to still be written if they're larger than small dots)

A couple remarks : 
- ~~shapes need an additional up/down cycle : we could change it to follow the same logic as the brush but this could also be done by not changing the logic of the shape part and only disallowing shapes to be too small (in that case we can create shapes that aren't even visible but are still part of the document)~~
- the eraser and typewriter both need an additional pen up/down to activate
- tools aren't affected (they won't change anything if they're activated on such a short pen down/up cycle)
- ~~This does not change behavior when a selection is deleted.~~


Doublonmousse:
I've added a condition to only create shapes if they're large enough. As things like arrows would still appear on a single pen event, I'll leave the current behavior for shapes after a selection cancellation (that is we need an extra up/down cycle)

Doublonmousse:
I've added the cancellation of the selection tool upon deleting a selection when the selection tool is temporary

crates/rnote-engine/src/pens/shaper.rs
```diff
@@ -153,18 +159,33 @@ impl PenBehaviour for Shaper {
...
+                            .iter()
+                            .map(|x| x.bounds())
+                            .reduce(|acc, x| acc.merged(&x))
+                            .unwrap_or(Aabb::new_invalid())
+                            .volume()
+                            > engine_view
+                                .pens_config
+                                .shaper_config
+                                .get_stroke_width()
+                                .powi(2);
```
Doublonmousse, line 173: This has to be removed. As I decided to have an additional pen up/down cycle, this doesn't serve any purpose. It also creates issues with constraints (a line that's horizontal or vertical won't satisfy the condition)

crates/rnote-ui/src/appwindow/actions.rs
```diff
@@ -406,7 +406,19 @@ impl RnAppWindow {
...
+                // trigger an event for a KeyboardPress::Delete
+                let (_, widget_flags) = canvas.engine_mut().handle_pen_event(
+                    rnote_compose::PenEvent::KeyPressed { keyboard_key: KeyboardKey::Delete, modifier_keys: vec![] },
+                    None,
+                    Instant::now()
+                );
+                widget_flags
+            } else {
+                canvas.engine_mut().trash_selection()
+            };
```
flxzt, line 421: Shouldn' this logic be in the engine in the `trash_selection()` function?

crates/rnote-engine/src/pens/brush.rs
```diff
@@ -229,6 +230,28 @@ impl PenBehaviour for Brush {
...
+                        // behind
+                        let volume = engine_view
+                            .store
+                            .get_stroke_ref(*current_stroke_key)
+                            .unwrap()
+                            .bounds()
+                            .volume();
+                        if engine_view.store.get_cancelled_state()
+                            && volume
+                                < 4.0
```
flxzt, line 244: What about naming this value by adding a constant variable to make it more readable?

crates/rnote-engine/src/pens/selector/penevents.rs
```diff
@@ -180,9 +180,16 @@ impl Selector {
...
                         } else {
                             // when clicking outside the selection bounds, reset
+                            tracing::debug!("cancelling the selection");
                             engine_view.store.set_selected_keys(selection, false);
                             self.state = SelectorState::Idle;
 
+                            // we create a new engine variable that indicates that the pen event is one from the same
+                            // event sequence that the one that cancelled the selection
+                            // this is done by setting this variable to true here and
+                            // removing it on a pen up event (and doing more things if this variable is true on a pen up)
```
flxzt, line 190: this comment should probably be modified and explain the cancelled_state variable in the engine struct definition, rather than having it here.

crates/rnote-ui/src/appwindow/actions.rs
Doublonmousse: yes, it's only called from here anyway. I changed it and it may also make it easier to merge as well

crates/rnote-engine/src/pens/brush.rs
Doublonmousse: Yes, done. I've added a little docstring to it as well

crates/rnote-engine/src/pens/selector/penevents.rs
Doublonmousse: I modified the comment here and moved the rest in the engine struct definition.

I also did a little renaming (the getters/setters name weren't matching the variable name in the engine struct definition)

Doublonmousse:
Putting back in draft as this needs a little work before it can be merged.
I'm wondering if the condition can be slightly modified to include time delay as well. It will cancel a dot after a selection but sometimes this dot can be large enough to stay (if the pen slides a bit instead of a clean hit)

PR #1141, opened 2024-07-11
detect multiple touch inputs

Doublonmousse:
Beware : quite experimental and not fully tested.

In particular, the current version creates a lot of 
```
(rnote.exe:11404): GLib-GObject-CRITICAL **: 23:44:59.380: g_boxed_copy: assertion 'src_boxed != NULL' failed

(rnote.exe:11404): GLib-GObject-CRITICAL **: 23:44:59.380: g_boxed_free: assertion 'boxed != NULL' failed
```
And seems to increase drastically the compilation time.

This replaces `event.device().unwrap().num_touches()` by another approach, using the fact that each touch sequence is associated to a unique (and constant) sequence. Hence we can use `event.event_sequence().as_ptr() as usize` to distinguish different touches and follow it.

This prevents any zip-zap when there are multiple fingers on the screen when `Draw with Touch` is enabled.

Apparently, the `EventSequence` type is touch specific (so maybe it would be possible to add this into #672 ? Though I don't know if you'd segfault if the function is called on a non touch event).

### DIFF
--- a/crates/rnote-ui/src/canvas/input.rs
+++ b/crates/rnote-ui/src/canvas/input.rs
@@ -1,5 +1,6 @@
 // Imports
 use super::RnCanvas;
+use adw::glib::subclass::types::ObjectSubclassIsExt;
 use gtk4::{gdk, glib, graphene, prelude::*, Native};
 use rnote_compose::penevent::{KeyboardKey, ModifierKey, PenEvent, PenState, ShortcutKey};
 use rnote_compose::penpath::Element;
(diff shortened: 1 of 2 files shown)


Doublonmousse:
Gonna close this I think for now.

I wanted to try to do something else : implement a counter and increment/decrement with `TouchBegin/TouchCancel/TouchEnd` events but I can't get it to fully work. 
I guess these events would have to be always counted and I'm not sure it's the case with the event propagation right now (I'm trying it in the same function).

It _can_ prevent the zigzag issue but it's far from perfect (the counter can sometimes increase over time, going from 2 touches back to one will immediately resume drawing, sometimes this also happen after the end of a zoom gesture).

It's kinda annoying though that `num_touches()` doesn't really give you this info ...

PR #1143, opened 2024-07-12
fix: prevent accidental resizes when interacting with buttons in the titlebar

### LINKED ISSUE #1142: Accidental resizes when clicking buttons in the titlebar
### Discussed in https://github.com/flxzt/rnote/discussions/1140

<sup>Originally posted by **xkmanlordx** July 12, 2024</sup>
First of big fan.

One minor annoyance though: whilst using my tablet to take notes, when trying to press the save button, or open the sidebar, due to slight imprecision, the window resizes instead of clicking the symbol i was aiming for. 
Maybe introducing a delay for the resize feature could help, although I can see how that could also be annoying, perhaps the hit-boxes could be made slightly larger, but I have no programming knowledge, so I'm just spitballing.

Either way, great app, many thanks for putting it out there.</div>

Doublonmousse:
Prevent events that occur on buttons to propagate to the titlebar. This makes it impossible to accidentally resize the window when clicking or touching on a button in the titlebar.

Fixes #1142

### DIFF
--- a/crates/rnote-ui/data/ui/mainheader.ui
+++ b/crates/rnote-ui/data/ui/mainheader.ui
@@ -59,7 +59,7 @@
           </object>
         </child>
         <child type="end">
-          <object class="GtkBox">
+          <object class="GtkBox" id="right_buttons">
             <property name="spacing">3</property>
             <child>
               <object class="GtkButton">
--- a/crates/rnote-ui/data/ui/mainheader.ui
+++ b/crates/rnote-ui/data/ui/mainheader.ui
@@ -59,7 +59,7 @@
           </object>
         </child>
         <child type="end">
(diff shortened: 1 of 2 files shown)


crates/rnote-ui/data/ui/mainheader.ui
```diff
@@ -59,7 +59,7 @@
           </object>
         </child>
         <child type="end">
-          <object class="GtkBox">
+          <object class="GtkBox" id="right_buttons">
```
flxzt, line 62: I'd name the box `right_buttons_box` for consistency

Doublonmousse: fixed

flxzt:
Looks good! :)

PR #1144, opened 2024-07-12
change `RnUnitEntry` to not grab scroll focus

Doublonmousse:
`SpinButton` changed to `adw::SpinRow` to not grab scroll focus.

This prevents accidentally large changes in values when scrolling with the mouse the settings window.

### DIFF
--- a/crates/rnote-ui/data/ui/settingspanel.ui
+++ b/crates/rnote-ui/data/ui/settingspanel.ui
@@ -183,37 +183,23 @@ gets disabled.</property>
                       </object>
                     </child>
                     <child>
-                      <object class="AdwActionRow" id="format_width_row">
+                      <object class="RnUnitEntry" id="format_width_unitentry">
                         <property name="title" translatable="yes">Width</property>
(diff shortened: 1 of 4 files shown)


flxzt:
Hm, not sure if SpinRow should be used like that, I think it should always be a direct child of a ListBox or other list containers.

Maybe the SpinButton itself can be adapted to not grab focus while scrolling?

Also: I'd like to keep that different step increments are used depending on the selected unit, that would need to be set on the adjustment.

Doublonmousse:
I think I can make the step increments work on the SpinRow using https://docs.rs/libadwaita/latest/libadwaita/struct.SpinRow.html#method.set_adjustment

To use `SpinButton`, I would have to add controllers to catch scroll events. But then I don't know how I could pass events to the scrolled window (cause stopping the propagation of scroll events before the button will only make scrolling stop if the mouse gets over the button, which technically prevents the core issue but introduces a new one).

Doublonmousse:
You're right that this is usually the full row. Maybe I can subclass the selection element inside of the `SpinRow` instead of what I'm currently doing

PR #1146, opened 2024-07-14
improv: unit entry ignoring scrolls entirely

flxzt:


crates/rnote-ui/src/unitentry.rs
```diff
@@ -59,6 +59,20 @@ mod imp {
             self.configure_spinner(self.unit.get(), self.dpi.get());
             self.value_spinner.set_value(10.0);
 
+            // Disable scrolling entirely,
+            // fixing unexpected scrolling issue when the unit entry is inside a scrolled window
```
Doublonmousse, line 63: Should we add a link to https://gitlab.gnome.org/GNOME/libadwaita/-/blob/main/src/adw-spin-row.c#L483 (to remember that this is following the same tactic Adw::SpinRow is, for reference ?)

flxzt: that's a good idea, will add it to the comment!

Doublonmousse:
looks good

PR #1151, opened 2024-07-20
Optimizations for compress_to_gzip and decompress_from_gzip

anesthetice:
## compress_to_gzip :
compression level reduced to 5 from 6 (default)
minuscule increase in file sizes, better save times
2024-07-20-234720_hyprshot

## decompress_from_gzip
Vec::with_capacity() optimization
roughly ~15% faster

### DIFF
--- a/crates/rnote-engine/src/fileformats/rnoteformat/mod.rs
+++ b/crates/rnote-engine/src/fileformats/rnoteformat/mod.rs
@@ -30,10 +30,24 @@ fn compress_to_gzip(to_compress: &[u8]) -> Result<Vec<u8>, anyhow::Error> {
 
 /// Decompress from gzip.
 fn decompress_from_gzip(compressed: &[u8]) -> Result<Vec<u8>, anyhow::Error> {
-    let mut decoder = flate2::read::MultiGzDecoder::new(compressed);
-    let mut bytes: Vec<u8> = Vec::new();
-    decoder.read_to_end(&mut bytes)?;
+    // Optimisation for the gzip format, defined by RFC 1952
+    // vector's capacity defined by the size of the uncompressed data
+    // size given in little endian format, last 4 bytes of "compressed"
+    //
+    //   ISIZE (Input SIZE)
+    //     This contains the size of the original (uncompressed) input data modulo 2^32.
+    let mut bytes: Vec<u8> = std::panic::catch_unwind(|| {
+        let mut decompressed_size: [u8; 8] = [0; 8];
+        decompressed_size[0..4].copy_from_slice(&compressed[compressed.len() - 4..]);
+        Vec::with_capacity(usize::from_le_bytes(decompressed_size))
+    })
+    .unwrap_or_else(|_| {
+        tracing::warn!("Could not determine the size of the uncompressed data");
+        Vec::new()
+    });
 
+    let mut decoder = flate2::read::MultiGzDecoder::new(compressed);
+    decoder.read_to_end(&mut bytes)?;
     Ok(bytes)
 }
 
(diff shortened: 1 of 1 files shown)


RayJW:
I don't have time to test out this theory for at least another month, but maybe if you can find some time, since you already are playing around with compression. 1. is there any reason Rnote doesn't use something more modern and performant like zstd which usually has better compression and speed at ~level 3 than gzip, and if there is a reason then 2. maybe some benchmarks with pigz would be worth it to see if the parallelism is worth implementing? Anyway, thanks for the interesting work on this! :)

crates/rnote-engine/src/fileformats/rnoteformat/mod.rs
```diff
@@ -23,18 +23,31 @@ use std::io::{Read, Write};
...
 
 /// Decompress from gzip.
 fn decompress_from_gzip(compressed: &[u8]) -> Result<Vec<u8>, anyhow::Error> {
+    // Optimization for the gzip format, defined by RFC 1952
+    // capacity of the vector defined by the size of the uncompressed data
+    // given in little endian format, by the last 4 bytes of "compressed"
+    //
+    //   ISIZE (Input SIZE)
+    //     This contains the size of the original (uncompressed) input data modulo 2^32.
+    let mut bytes: Vec<u8> = std::panic::catch_unwind(|| {
```
flxzt, line 39: TIL about that, very cool. But can this be refactored to not use catch unwind? I think it's not that great to use as a general try-catch mechanism, it should be possible to use fallible methods instead

flxzt:
Thanks a lot for doing this work! The graphs very clearly visualize the trade-offs between the different compression levels. I also agree with using level 5, making compression faster with only a negligible increase in file size

crates/rnote-engine/src/fileformats/rnoteformat/mod.rs
anesthetice: You're right, for some reason I didn't want this part of the code to fail for any reason, despite the fact that an error would be returned a couple of lines later if the file format was invalid, refactored :+1:

crates/rnote-engine/src/fileformats/rnoteformat/mod.rs
```diff
@@ -23,18 +23,31 @@ use std::io::{Read, Write};
...
 /// Decompress from gzip.
 fn decompress_from_gzip(compressed: &[u8]) -> Result<Vec<u8>, anyhow::Error> {
+    // Optimization for the gzip format, defined by RFC 1952
+    // capacity of the vector defined by the size of the uncompressed data
+    // given in little endian format, by the last 4 bytes of "compressed"
+    //
+    //   ISIZE (Input SIZE)
+    //     This contains the size of the original (uncompressed) input data modulo 2^32.
+    let mut bytes: Vec<u8> = {
+        let mut decompressed_size: [u8; 8] = [0; 8];
```
flxzt, line 40: Instead of hardcoding this as 8 bytes and using `usize::from_le_bytes` I think [u32::from_le_bytes](https://doc.rust-lang.org/stable/core/primitive.u32.html#method.from_le_bytes) should be used and then be cast to usize. Otherwise this would fail to compile on 32 bit architectures because then `usize::from_le_bytes` expects an array with size 4 (see [docs](https://doc.rust-lang.org/stable/core/primitive.usize.html#method.from_le_bytes))

anesthetice: Good catch, on it

flxzt:
looks good to me now, thanks for your contribution :)

Issue #639, opened 2023-04-22
Pages menu, scroll with pages 


Intranox:
I really like this software and it really includes a lot of things. One feature that could be useful would be to have the ability to have a menu where you can jump directly to a page or enter the page number and go directly to one of them. It depends on the use sometimes documents get quite long and jumping to specific points in the document through the pages with preview could help a lot

vBazilevich:
Another useful related feature: pages re-ordering

Intranox:
> Another useful related feature: pages re-ordering

Yes, this is a good idea. Another interesting idea could be to add a page between two already created pages

PR #1153, opened 2024-07-22
page indicator for fixed layouts

Doublonmousse:
Partial fix for #639 

[Screencast_20240722_132148.webm](https://github.com/user-attachments/assets/45cc0397-1ea5-4a7c-bfc3-eb0f6f06519c)

This does not include any page menu, only a page number indicator to show the current page number and change the page. For now this is only implemented for fixed layouts (infinite layouts would need a little more work as page positions and number depend on a little more than the current canvas size, like which pages include strokes. Having a page number might be a little more difficult as well).

There's probably a few things to improve (these are suggestions)
- [ ] make it work for infinite layouts (?)
- [ ] make it async ? (there's no need to halt some other calculations for the page number to be calculated, not sure it currently is the case)
- [ ] show the total number of page (I would like to have a 1/xx inside the spinbutton but that might end up being hard to do with gtk. Having the total number of page on the right would be a little too much wasted space)

### DIFF
--- a/crates/rnote-ui/src/appwindow/mod.rs
+++ b/crates/rnote-ui/src/appwindow/mod.rs
@@ -12,6 +12,7 @@ use adw::{prelude::*, subclass::prelude::*};
 use gettextrs::gettext;
 use gtk4::{gdk, gio, glib, Application, IconTheme};
 use rnote_compose::Color;
+use rnote_engine::document::Layout;
 use rnote_engine::ext::GdkRGBAExt;
 use rnote_engine::pens::pensconfig::brushconfig::BrushStyle;
 use rnote_engine::pens::pensconfig::shaperconfig::ShaperStyle;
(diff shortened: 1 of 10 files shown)


crates/rnote-ui/src/appwindow/mod.rs
```diff
@@ -492,6 +492,38 @@ impl RnAppWindow {
...
+                // give the previous page when the go_to_page function is called
+                // hence a cascade of calls are made until the first page is shown
+                // maybe we should set the page number based on the page visible
+                // at the center of the viewport instead
+                let page_number = ((canvas.engine_ref().camera.offset().y + 0.01)
+                    / (canvas.engine_ref().document.format.height()
+                        * canvas.engine_ref().camera.zoom()))
+                .ceil()
+                .max(1.0);
+                self.main_header().page_spin().set_value(page_number);
```
Doublonmousse, line 517: I should probably only trigger the `set_value` if the page number is actually different from the actual value
+ use `block_signal`/`unblock_signal`  on the `page_spin` so that the function to change the page isn't called twice (thanks @Clemapfel for the idea)

Doublonmousse: Seems like doing this is harder than it seems because [glib](https://gtk-rs.org/gtk-rs-core/git/docs/glib/index.html)::[signal](https://gtk-rs.org/gtk-rs-core/git/docs/glib/signal/index.html)::[SignalHandlerId](https://gtk-rs.org/gtk-rs-core/git/docs/glib/signal/struct.SignalHandlerId.html#) does not implement Clone, hence it's difficult to save the value and use it in another function to block/unblock things

Doublonmousse: https://github.com/gtk-rs/gtk4-rs/issues/1817

Doublonmousse:
Gonna close this as this stayed as a draft, is pretty incomplete, and would need to be rebased anyway.

PR #1170, opened 2024-08-07
zstd with parallelism instead of gzip

anesthetice:
https://github.com/anesthetice/rnote-compression-benchmarking/

From the compression methods tested, zstd seems to be a better choice for the  rnote file format:

- Better compression ratio
- Faster reads + writes
- Simple multi-threading

median out of 16 samples

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3404,6 +3404,7 @@ dependencies = [
  "unicode-segmentation",
  "usvg",
  "xmlwriter",
+ "zstd",
 ]
 
 [[package]]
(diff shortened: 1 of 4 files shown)


anesthetice:
Implementation went smoother than what I expected, rnote can still decompress gzip files and will compress them using zstd the next time the file is saved. One thing left to discuss I think is, should we expose the compression level to users?

anesthetice:
Here's a rough before and after comparison of a couple files (different to the ones used for rnote-comp-benchmarking)

Before:
2024-08-08-190025_hyprshot

After:
2024-08-08-190046_hyprshot

RayJW:
Amazing work, that is exactly what I was talking about. Thanks, for looking into this! Looks like it's definitely worth considering. :)

anesthetice:
converted this to a draft, to focus on an updated file format

anesthetice:
closing this to de-clutter

Issue #384, opened 2022-11-29
stroke straightening when holding pen still after stroke


olbotta:
Some note-taking apps have this cool feature to make really straight lines in a simple way:

when you draw a stroke, and you stand still with the pen without lifting it from the screen, after ~1s a straight stroke is drawn from the stroke starting point to the point you are holding the pen (and the original stroke gets deleted).

Usually this does not work when the initial stroke is too "wavy".

tad-lispy:
That would make it lovely for quickly drawing wireframes.

ignassew:
Samsung notes also have this for shapes: If you draw a (circle, square, hexagon, star, etc) and hold still for some time, it will transform it into a clean shape.

ignassew:
Similar issue:  #120

SrEstegosaurio:
Having my math fractions straight would be lovely. Hope someone picks this up.

Doublonmousse:
The first part would be to detect correctly the gesture of holding the pen still after the stroke.

For pens/touch/graphical tablet, a simple algorithm would be possible without changing too much the logic (have some timeout and check whether the events have moved outside of some radius over a lap of time or not).

The annoying part  would be that this won't work at all for mouse input (if you hold the mouse still, no events are fired, whereas there would be events for pen/graphical tablet/touch as there is always some noise in the input).

So fixing that would mean having an extra timeout that's reset on every event and that fires an event (maybe an additional pen event ?) if the pen is held down with no event coming up, in a way that doesn't mess things up (we can't have two events at the same time mutating the engine)

Issue #120, opened 2022-02-20
Add shape recognition


SimonBrandner:
**Is your feature request related to a problem? Please describe.**

While making notes, people often draw simple shapes such as lines, circles, triangles, squares... Switching to a different tool often requires a lot of unnecessary hand movement and takes time that is then wasted.

**Describe the solution you'd like**

I'd like a shape recognition feature to be added to aid in drawing simple shapes. This makes drawing/taking notes a much simpler, quicker and generally more enjoyable experience.

**Describe alternatives you've considered**

**Additional context**

Other apps have this:

+ Xournal++ 

https://user-images.githubusercontent.com/25768714/154840426-ab31c779-968a-44eb-b36d-48234f3d9fdd.mp4

+ Microsoft Whiteboard

https://user-images.githubusercontent.com/25768714/154840428-d81d9507-fa69-48d1-8e0d-0ceba0ad6ad0.mp4

PhilDevProg:
This should be available in two tools in my opinion:
- A switch in the settings menu of the pen tool
- A button in the selection menu to recognize already drawn shapes

RayJW:
I think this should be doable by using something like the [$-family](https://depts.washington.edu/acelab/proj/dollar/index.html) of algorithms, or maybe work based off of this group of algorithms like [Penny Pincher](https://www.sciencedirect.com/science/article/pii/S0097849315001788). Either algorithm was already considered lightweight when the research was published, but should have negligible impact anyway considering the evolution of performance since their publishing. All of these algorithms showed a recognition rate of 95% - 99% which should be plenty for trivial shapes like arrows and lines. 

I think regardless of the implementation, the shape database would have to be created. Although, one could use their dataset from the research project, it would contain a bunch of shapes irrelevant for Rnote increasing the chance of recognition issues.

szymongul:
bump

PR #1175, opened 2024-08-12
feat: long hold pen for lines/shape recognition

Doublonmousse:
This is still a draft but I want to be sure that the general structure is correct before going further.
At the moment it always transform the pen path into a line between the start and end position if the gesture is detected.

For now what I'm worried about
- [ ] does the current implementation works without mutation issues on the engine ?
- [ ] I'm not fully sure the use of the `OneOffTask` is completly correct
- [ ] tweak the settings and test it with touch/pen input (does it work correctly).

Then there is a lot of functionality missing 
- [ ] toggle the setting in the UI (and have this propagate into the rest of the code so no power is lost detecting a gesture for nothing)
- [ ] emit shapes with the correct stroke width (and extend it further than just lines, for now I'm not entirely sure how that could be done, there's probably new methods to add to the shaper to be able to do that from a single function call)
- [ ] implement the shape recognizer. For now there is no shape recognizer as I was only searching for a way to make the gesture work. At least the general form of the recognizer should be a `PenPath -> Option<Shape>' so it either recognizes something and returns it or it doesn't. 

Issues : #384 and maybe #120 if the last point is developped further than just lines/not lines

### DIFF
--- a/crates/rnote-engine/src/engine/mod.rs
+++ b/crates/rnote-engine/src/engine/mod.rs
@@ -17,7 +17,7 @@ pub use strokecontent::StrokeContent;
 
 // Imports
 use crate::document::Layout;
-use crate::pens::{Pen, PenBehaviour, PenStyle};
+use crate::pens::{Pen, PenStyle};
 use crate::pens::{PenMode, PensConfig};
 use crate::store::render_comp::{self, RenderCompState};
 use crate::store::StrokeKey;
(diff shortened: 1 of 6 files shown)


PeterWang-dev:
Any progress on this PR?

walndyn:
I would also like to know if this PR is still in work or depracted?

This is the only thing that keeps me from switching from OneNote

Doublonmousse:
> I would also like to know if this PR is still in work or depracted?
> 
> This is the only thing that keeps me from switching from OneNote

Well no, nothing has changed since last time. Still at the very first step/line (does the long hold gesture works ?).
Given the time frame since the last commit it will have to be rebased anyway. I think other commits/PR will get merged first before I go again on this one

walndyn:
Alright. Thank you for your fast reply :)

Issue #1173, opened 2024-08-11
New file format (stabilization)


flxzt:
Let's track what Rnote's new file format should look like here.

There are already some ideas and implementation for improvements floating around.

Ideally backwards compatibility is kept which I think is doable.

Improvements are:
- A new compression format : see #1170
- Move the file format version outside of the compressed data (first bytes of the file) to improve forward compatibility
- Look into using another de-serialization format than json : bincode is very promising because it greatly improves speed, file size and memory consumption.
 Things to consider are: how to continue making upgrade path's relatively easy to implement - currently json's untyped `Value`'s are used, that won't work with other formats. I can see two possibilities: limit additions and changes to what is backwards compatible from the perspective of the serde derived `Serialize` and `Deserialize` trait, and if that's not possible start implementing both traits manually. This can be quite tedious in some cases though.
- Storing stroke coordinates more effiently : the [ink-serialized-format](https://en.wikipedia.org/wiki/Ink_Serialized_Format) does something interesting - using derivatives to reduce the data for strokes with specific shapes (?) (@Doublonmousse knows more)
- Partial file loading : On large documents it might be preferable not having to load the entire document into memory when viewing/editing only some pages. Is is possible to encode that into the format similar to how we use an R-Tree for storing stroke bounding boxes?

Doublonmousse:
For the ink serialized format, there is no floats used to store ink strokes, there's only a minimum fixed size (like 1/1000th of a cm) and quantization is used. Hence everything is using ints. The other thing used is that instead of saving the data as 
```
a b c etc...
```
It's saved as 
```
a, (b-a), c - (a + 2 (b-a)) etc...
```
Hence first element, first derivative (difference) and then only second derivatives. The general idea is that the second derivatives are usually very small, hence additional compression can be done on top of this (huffman or bitpacking as we expect low values that can be represented on very few bits).
The code is actually public : https://source.dot.net/#PresentationCore/MS/Internal/Ink/InkSerializedFormat/ under the MIT License so maybe I'll look into having that in rust, as a way to have a little more interoperability with other apps, including onenote when copying/pasting between apps.

I'd be great though to still have an easy way for people to obtain a human-readable or json version of the file and/or the file specification. Having vendor lock in for files because of undocumented binary files, or documented but without readily available readers that most people can use is not something I want, having been victim of this myself and the sheer insanity of trying to get the data out.

[...]

PR #1177, opened 2024-08-14
More versatile rnote file format, issue 1173

anesthetice:
## proposal
Untitled-2024-08-14-1317

## todo

- [ ] agree on a format standard
- [ ] implement proper forward-compatibility
- [ ] configuration
- [ ] conversions to alternative save

https://github.com/flxzt/rnote/issues/1173

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3404,6 +3404,7 @@ dependencies = [
  "unicode-segmentation",
  "usvg",
  "xmlwriter",
+ "zstd",
 ]
 
 [[package]]
@@ -4876,6 +4877,34 @@ name = "zeroize"
 version = "1.8.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "ced3678a2879b30306d323f4542626697a464a97c0a07c9aebf7ebca65cd4dde"
+
+[[package]]
+name = "zstd"
+version = "0.13.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fcf2b778a664581e31e389454a7072dab1647606d44f7feea22cd5abb9c9f3f9"
+dependencies = [
+ "zstd-safe",
+]
+
+[[package]]
+name = "zstd-safe"
+version = "7.2.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "54a3ab4db68cea366acc5c897c7b4d4d1b8994a9cd6e6f841f8964566a419059"
+dependencies = [
+ "zstd-sys",
+]
+
+[[package]]
+name = "zstd-sys"
+version = "2.0.13+zstd.1.5.6"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "38ff0f21cfee8f97d94cef41359e0c89aa6113028ab0291aa8ca0038995a95aa"
+dependencies = [
+ "cc",
+ "pkg-config",
+]
 
 [[package]]
 name = "zune-core"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -86,6 +86,7 @@ winresource = "0.1.17"
 xmlwriter = "0.1.0"
 # Enabling feature > v20_9 causes linker errors on mingw
 poppler-rs = { version = "0.23.0", features = ["v20_9"] }
+zstd = { version =  "0.13", features = ["zstdmt"] }
 
 [patch.crates-io]
 # once a new piet (current v0.6.2) is released with updated cairo and kurbo deps, this can be removed.
--- a/crates/rnote-engine/Cargo.toml
+++ b/crates/rnote-engine/Cargo.toml
@@ -55,6 +55,7 @@ tracing = { workspace = true }
 unicode-segmentation = { workspace = true }
 usvg = { workspace = true }
 xmlwriter = { workspace = true }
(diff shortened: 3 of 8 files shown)


anesthetice:
I just tested serializing with default bincode, the resulting file (bincode, using zstd) is larger than the original (json, using gzip), furthermore I could not test v2.0.0-rc.3 as they have moved away from serde it seems and are using their own Decode and Encode trait

anesthetice:
@flxzt could you please take a quick look at this draft when you have a moment? It would be good to get your thoughts on this approach and whether I should continue.. :eyes:

anesthetice:
2024-08-19-145446_hyprshot

After giving it more thought, I think it's better to not show "Serialization Method" or "Compression Method" to users, and to just default to Bitcode and Zstd. However I do see the merit in allowing users to choose a compression level, and implementing a way to convert files and lock the methods used (e.g. for debugging purposes, you could convert a Rnote file encoded with bitcode and compressed with zstd to an uncompressed JSON-encoded format and when the file is re-saved it will not be converted back to the default (bitcode + zstd))

anesthetice:
saveprefs

anesthetice:
1. simple use case, opening multiple legacy files and saving them into the new format (loading legacy files takes longer but it's very quick once in the new format) 
2. using rnote-cli to mutate multiple files (gzip(5) + json to zstd(17) + bitcode) 
3. demo of the method lock, you can work on a file encoded with json and uncompressed as long as it is locked

https://github.com/user-attachments/assets/0c90cbcd-1895-4a1c-878e-b293690f1aed


https://github.com/user-attachments/assets/9d9511a3-7cc8-4714-95d6-77efcf47e49c


https://github.com/user-attachments/assets/0ded9a6d-8ca0-41d3-a5cb-b9f4fb029b37

anesthetice:
### demo 1 - backwards compatibility, size ratio, persistent compression level
https://github.com/user-attachments/assets/a58a541b-6cfc-4a49-a398-aaff1937cdb6

### demo 2 - rnote-cli mutate (atomic file saving is also noticeable)
https://github.com/user-attachments/assets/0e0d9f27-8876-4b18-9d3a-a9cd879f1737

anesthetice:
Getting some weird issues with bitcode relating to the packing that aren't present with json, might have to reconsider using bitcode sadly..

RayJW:
> Getting some weird issues with bitcode relating to the packing that aren't present with json, might have to reconsider using bitcode sadly..

What exactly are those issues, if I may ask? I see you're using bitcode with serde which seems to have [issues](https://github.com/SoftbearStudios/bitcode/issues/47) for the time being. Not sure if that is related, though. It might be that some `serde_json` types cause problems according to the [wiki](https://github.com/SoftbearStudios/bitcode/wiki/Serde).

RayJW:
> > > Getting some weird issues with bitcode relating to the packing that aren't present with json, might have to reconsider using bitcode sadly..
> > 
> > 
> > What exactly are those issues, if I may ask? I see you're using bitcode with serde which seems to have [issues](https://github.com/SoftbearStudios/bitcode/issues/47) for the time being. Not sure if that is related, though. It might be that some `serde_json` types cause problems according to the [wiki](https://github.com/SoftbearStudios/bitcode/wiki/Serde).
> 
> I merged this PR alongside my "Styled Lines" PR: #1210 in a branch of Rnote I personally use, however opening previous "versions" of my files serialized using bitcode yielded a "Error: invalid packing", I tried debugging this with a forked version of bitcode that uses [crates.io/crates/serde_path_to_error](https://crates.io/crates/serde_path_to_error), (forgot to save the output sorry) but didn't find a "fix".
> 
> The bigger problem is that from what I can tell, bitcode is sadly but understandably not as forward-compatible as JSON. I still plan on keeping bitcode as a serialization option in this PR, for users willing to use rnote-cli and the method "lock", but JSON will probably go back to being the default.

That's disappointing but also very understandable. Thank you for working on this either way! What about the other options you explored, like Bincode or Postcard? I personally would prefer those over JSON because I don't really care about disk space at all, but the speed improvements are very interesting for the daily workflow. Because breaking forward compatibility would kinda suck.

finnbear:
> using bitcode with serde which seems to have [issues](https://github.com/SoftbearStudios/bitcode/issues/47) for the time being.

Just so you know, the linked issue would allow you to call `bitcode::{encode, decode}` on a `struct` where one field only implements `serde` traits.

This does not affect what you're using, `bitcode::{serialize, deserialize}`, at all :)

> The bigger problem is that from what I can tell, bitcode is sadly but understandably not as forward-compatible as JSON

`bitcode` is not self-describing, which has forwards-compatibility implications depending on your schema. It also [doesn't support all `serde` features](https://github.com/SoftbearStudios/bitcode/wiki/Serde#unsupported-features).

anesthetice:
https://github.com/anesthetice/rnote-testing

anesthetice:
Just a quick update, the PR is in my opinion mostly ready (the cli, documentation, and error handling could still be improved), I've been using it myself for a while after all. But there is still a lot of work to be done before it can be merged, as we want to stabilize a Rnote format (encoding/serialization) alongside any major changes to the file format, which I plan to work on during the summer

anesthetice:
Moved to #1578, as I wanted to implement some major changes, and it would take longer to solve every merge conflict that just cleanly restart at a new branch.

Issue #1063, opened 2024-03-30
Memory leak with embedded PDF.


RayJW:
**Describe the bug**
Memory leak while trying to load PDF embedded into Rnote document.

**To Reproduce**  
1. Importing the PDF already takes multiple minutes on my relatively performant laptop.
2. Once it is loaded, viewing the PDF works fine but uses a lot of memory until reaching the third page, Rnote now takes over 10 Gb's of memory and reliably gets killed because it's taking down my other processes.

**Expected behavior**  
This obviously shouldn't happen, and I think it's fair to talk about a memory leak, although this is a Rust application I suspect the PDF library might not be. This PDF is only 3 pages, 2 Mb and works fine in all other PDF applications (Firefox, Evince, and Okular in my case).

**Console Output**  
```
2024-03-30T19:17:47.148322Z DEBUG rnote: .. tracing subscriber initialized.
2024-03-30T19:18:23.031656Z ERROR rnote_engine::store::render_comp: Generating stroke images failed stroke while regenerating rendering in viewport `Aabb { mins: [-123.06843938144867, 720.3065966056062], maxs: [917.8854437199712, 1943.2992129587522] }`, stroke key: StrokeKey(3v1), Err: read stream to rsvg loader failed.

Caused by:
    cannot load more than 1000000 XML elements

Stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: __clone
2024-03-30T19:19:06.073044Z ERROR rnote_engine::store::render_comp: Generating stroke images failed stroke while regenerating rendering in viewport `Aabb { mins: [-94.61117261757015, 1369.3988513674444], maxs: [888.9508093859195, 2309.5973792545474] }`, stroke key: StrokeKey(3v1), Err: read stream to rsvg loader failed.

Caused by:
    cannot load more than 1000000 XML elements

Stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: <unknown>
   9: __clone
[...]

PR #1178, opened 2024-08-15
Remove unmaintained "atty" crate

RayJW:
I was trying to see if I can get to the bottom of #1063 when I saw the security advisory for the dependency atty which has an (arguably harmless) open [vulnerability](https://rustsec.org/advisories/RUSTSEC-2021-0145.html). Since it is only used a single time and the functionality is available natively in Rust since 1.70 (which Rnote has been using for a while) I quickly got rid of it.


flxzt:
Looks good, thank you! Easy to forget such code hygiene tasks.

PR #1188, opened 2024-08-25
Improve Regex for duplication logic

### LINKED ISSUE #1179: Duplicate only allows one "-" in the file name
**Describe the bug**  
If a file has a hyphen in the name e.g. `test - test.rnote` and I select to duplicate it, the new file removes the existing hyphen and adds one at the end with the new number e.g. `testtest - 1.rnote`.

**To Reproduce**  
1. Create a file with a hyphen e.g. `Exam 1 - Task 1.rnote`
2. Duplicate the file for the second task (right click on the three dots, duplicate)
3. Check the name of the new file, mine is named `Exam 1Task 1 - 1.rnote` removing the first hyphen.

**Expected behavior**  
I would expect that the duplicate only adds an `- 1.rnote` to the existing file, without "changing" there the name.

**Console Output**  
Run the app in the terminal with `flatpak --env=RUST_LOG=rnote=debug --env=RUST_BACKTRACE=1 run com.github.flxzt.rnote`
and post the output.
```sh
❯ flatpak --env=RUST_LOG=rnote=debug --env=RUST_BACKTRACE=1 run com.github.flxzt.rnote
2024-08-20T07:52:40.050378Z DEBUG rnote: .. tracing subscriber initialized.
```
I think this log does not really help in this issue, or I did not found the proper one.

**Screenshots**  

 - OS: Fedora 40 - Workstation Edition
 - App Version: Rnote v0.11.0
 - Desktop Environment: Gnome 46
 - Display Server: Wayland
 - Input Source: Mouse and Keyboard

**Additional context**  
I was able to reproduce the described bug also on a Windows 10 machine, also running Rnote v0.11.0.

RayJW:
This PR fixes #1179 and in the process I also fixed some minor things for consistency. The solution also works without the loop now according to my testing. If you don't like any of the smaller rearrangements, feel free to let me know, and I will get rid of them. I also added / changed some of the comments to make it a bit easier to understand everything, if that's fine with you.

### DIFF
--- a/crates/rnote-ui/src/workspacebrowser/filerow/actions/duplicate.rs
+++ b/crates/rnote-ui/src/workspacebrowser/filerow/actions/duplicate.rs
@@ -12,18 +12,23 @@ use tracing::debug;
 
 /// The regex used to search for duplicated files
 /// ```text
-/// - Look for delimiter
+/// - Look for original filename lazily
 /// |         - Look for `<delim>1`/`<delim>2`/`<delim>123`/...
-/// |         |        - Look for the rest after the `<delim><num>` part
+/// |         |        - Look for the file extension after the `<delim><num>` part
 /// |         |       |       - At the end of the file name
 /// |         |       |       |
-/// |        \d*      |       $
+/// |         (?P<delim{}\d+)$
 /// |                 |
-/// DELIM       (?P<rest>(.*))
+/// ^(.*?)            (\.\w+)
 /// ```
 static DUP_REGEX: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(&(crate::utils::FILE_DUP_SUFFIX_DELIM_REGEX.to_string() + r"^(.*?)( - \d+)?$"))
-        .unwrap()
+    // Incorporate FILE_DUP_SUFFIX_DELIM_REGEX dynamically into the regex pattern
+    let regex_pattern: String = format!(
+        r"^(.*?)(?P<delim>{}\d+)?(\.\w+)?$",
+        crate::utils::FILE_DUP_SUFFIX_DELIM_REGEX
+    );
+
+    Regex::new(&regex_pattern).unwrap()
 });
 
 /// Create a new `duplicate` action.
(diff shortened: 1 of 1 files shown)


Doublonmousse:
I feel like adding a few more tests would be a good idea.
This way we can more clearly see what's the new intended behavior and keep it working going forward

RayJW:
> I feel like adding a few more tests would be a good idea. This way we can more clearly see what's the new intended behavior and keep it working going forward

I totally agree, I'm actually trying to add tests at the moment. But I can't run the current tests because jpeg-decoder doesn't compile for some reason. Once I get that running, I wanted to add tests for the issue that arose in #1179. Until I get that working, I thought the rest could already be reviewed properly.

Doublonmousse:
What's the compile error message ?

RayJW:
>    Compiling gettext-sys v0.21.3
> error: Unrecognized option: 'diagnostic-width'
> 
> error: could not compile `jpeg-decoder` (lib)
> warning: build failed, waiting for other jobs to finish...
> 
>  *  The terminal process "/home/user/Projects/GitHub/rnote/.flatpak/cargo.sh 'test', '--package', 'rnote', '--bin', 'rnote', '--', 'workspacebrowser::filerow::actions::duplicate::tests', '--show-output'" terminated with exit code: 101.

Doublonmousse:
What version of `rustc` is installed ? Can you update it ?
The only related issue I could find was this : https://github.com/rust-lang/rust/issues/118385 and it seems it's more related to the terminal or rust installation than a compile error from a dependency

RayJW:
Okay, so the error only occurs when I'm trying to run the tests with the rust-analyzer extension. When manually running `cargo test --package rnote --bin rnote -- workspacebrowser::filerow::actions::duplicate::tests --show-output` it works fine. Sorry, I'm only now getting into Rust, so all the testing and stuff is still quite new to me. It seems my change broke duplication of folders, the last test case is now failing, so I will look into that and add new tests for the added functionality.

crates/rnote-ui/src/workspacebrowser/filerow/actions/duplicate.rs
```diff
@@ -12,18 +12,23 @@ use tracing::debug;
...
 /// |                 |
-/// DELIM       (?P<rest>(.*))
+/// ^(.*?)            (\.\w+)
 /// ```
 static DUP_REGEX: Lazy<Regex> = Lazy::new(|| {
-    Regex::new(&(crate::utils::FILE_DUP_SUFFIX_DELIM_REGEX.to_string() + r"\d*(?P<rest>(.*))$"))
-        .unwrap()
+    // Incorporate FILE_DUP_SUFFIX_DELIM_REGEX dynamically into the regex pattern
+    let regex_pattern: String = format!(
+        r"^(.*?)(?P<delim>{}\d*)?(\.\w+)?$",
```
flxzt, line 27: Why not keep naming the capture groups? I think it improves readability when using the regex

flxzt:
alright, this looks good now. Thank you! :)

PR #1192, opened 2024-08-27
Add confirmation prompt to deletion action

### LINKED ISSUE #1136: prompt for confirmation when removing folders
**Describe the bug**  
I deleted the 'Documents' directory from the Default Workspace, corresponding to '~/Documents' linux directory.

The directory was permanently deleted, and not even a user dialog warning about this was issued.

The deleted directory wasn't moved to the Trash bin, it was permanently deleted.

I tried recovering it with the Linux tool 'testdisk' without success.

**To Reproduce**  
1. Go to the Workspace
2. Delete a directory

**Expected behavior**  

The deleted directory should have been moved to the Trash bin, or at least warn me that it was going to be permanently deleted.

**Console Output**  

I can't provide the output because I turned off the computer to try to recover the deleted data with 'testdisk' Linux utility. I'm afraid that turning it on again will overwrite the unallocated disk sectors that were originally occupied by the deleted data.

**Desktop**  
 - OS: Kubuntu 22.04
 - App Version: Rnote v0.10.2
 - Desktop Environment: KDE Plasma 5.24 
 - Display Server: Wayland

RayJW:
This is my WIP attempt to resolve #1136, but it isn't working at the moment. Clearly, I don't understand Rust and the whole borrowing logic enough for me to resolve this at the moment. I tried many different things but can't get the logic "Rusty" enough. At the moment, the issue is that the `&appwindow` reference in trash.rs doesn't live long enough. I tried creating a clone, among other things, but that didn't help either.

Other than that, I tried to copy the style of the other dialogues as much as possible to keep the whole architectural idea the same.

I'll keep working on this sometime, but in the meantime, if someone knows how to properly do this, feel free to point me in the right direction or open a PR.

### DIFF
--- a/crates/rnote-ui/data/ui/dialogs/dialogs.ui
+++ b/crates/rnote-ui/data/ui/dialogs/dialogs.ui
@@ -206,7 +206,7 @@ Changes which are not saved will be permanently lost.</property>
 
   <object class="AdwAlertDialog" id="dialog_trash_file">
     <property name="heading" translatable="yes">Trash File</property>
-    <property name="body" translatable="yes">Are you sure you want to delete this file?</property>
+    <property name="body" translatable="yes">Are you sure you want to move this file to the trash?</property>
     <property name="default-response">cancel</property>
     <property name="close-response">cancel</property>
     <responses>
--- a/crates/rnote-ui/data/ui/dialogs/dialogs.ui
+++ b/crates/rnote-ui/data/ui/dialogs/dialogs.ui
@@ -207,7 +207,7 @@ Changes which are not saved will be permanently lost.</property>
   <object class="AdwAlertDialog" id="dialog_trash_file">
     <property name="heading" translatable="yes">Trash File</property>
     <property name="body" translatable="yes">Are you sure you want to delete this file?</property>
-    <property name="default-response">trash</property>
+    <property name="default-response">cancel</property>
     <property name="close-response">cancel</property>
     <responses>
       <response id="cancel" translatable="yes">Cancel</response>
--- a/crates/rnote-ui/data/ui/dialogs/dialogs.ui
+++ b/crates/rnote-ui/data/ui/dialogs/dialogs.ui
@@ -203,4 +203,15 @@ Changes which are not saved will be permanently lost.</property>
       <object class="RnIconPicker" id="edit_selected_workspace_icon_picker"></object>
     </child>
   </object>
+
+  <object class="AdwAlertDialog" id="dialog_trash_file">
+    <property name="heading" translatable="yes">Trash File</property>
+    <property name="body" translatable="yes">Are you sure you want to delete this file?</property>
+    <property name="default-response">trash</property>
+    <property name="close-response">cancel</property>
+    <responses>
+      <response id="cancel" translatable="yes">Cancel</response>
+      <response id="trash" appearance="destructive" translatable="yes">Trash</response>
+    </responses>
+  </object>
 </interface>
--- a/crates/rnote-ui/src/dialogs/mod.rs
+++ b/crates/rnote-ui/src/dialogs/mod.rs
@@ -560,11 +560,11 @@ pub(crate) async fn dialog_trash_file(appwindow: &RnAppWindow, current_file: &gi
 
     match dialog.choose_future(appwindow).await.as_str() {
         "trash" => {
-            glib::spawn_future_local(clone!(@strong current_file, @weak appwindow => async move {
+            glib::spawn_future_local(clone!(@weak appwindow, @strong current_file => async move {
                 current_file.trash_async(
                     glib::source::Priority::DEFAULT,
                     None::<&gio::Cancellable>,
-                    clone!(@strong current_file, @strong current_file => move |res| {
+                    clone!(@weak appwindow, @strong current_file => move |res| {
                         if let Err(e) = res {
                             appwindow.overlays().dispatch_toast_error(&gettext("Trashing file failed"));
                             error!("Trash filerow file `{current_file:?}` failed , Err: {e:?}");
--- a/crates/rnote-ui/src/dialogs/mod.rs
+++ b/crates/rnote-ui/src/dialogs/mod.rs
@@ -8,6 +8,7 @@ use crate::canvas::RnCanvas;
 use crate::canvaswrapper::RnCanvasWrapper;
 use crate::config;
 use crate::workspacebrowser::workspacesbar::RnWorkspaceRow;
+use crate::workspacebrowser::RnFileRow;
 use crate::{globals, RnIconPicker};
 use adw::prelude::*;
 use gettextrs::{gettext, pgettext};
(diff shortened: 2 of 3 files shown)


Doublonmousse:
I'd say to look at what's done here : 
https://github.com/flxzt/rnote/blob/42f16dadb3d370aad72926e1017e26124dc1ff6e/crates/rnote-ui/src/dialogs/mod.rs#L108

Wrap what's done inside the "trash" option inside of the `glib::spawn_future_local(clone! ...` part, to have weak references instead of strong ones (if you don't you have lifetime issues)

crates/rnote-ui/src/dialogs/mod.rs
```diff
@@ -552,6 +553,38 @@ pub(crate) async fn dialog_edit_selected_workspace(appwindow: &RnAppWindow) {
     dialog.present(appwindow);
 }
 
+pub(crate) async fn dialog_trash_file(appwindow: &'static RnAppWindow, filerow: &RnFileRow) -> () {
```
Doublonmousse, line 556: You probably don't need the `&'static` here

Doublonmousse: Sucess : 
Screenshot 2024-08-27 at 22 30 07

RayJW: Wow, thanks for the hint. I added that when trying to fix another issue and totally forgot about it. Let me fix this up and then mark this ready for review. Tested it and it seems to work fine :)

crates/rnote-ui/data/ui/dialogs/dialogs.ui
```diff
@@ -203,4 +203,15 @@ Changes which are not saved will be permanently lost.</property>
       <object class="RnIconPicker" id="edit_selected_workspace_icon_picker"></object>
     </child>
   </object>
+
+  <object class="AdwAlertDialog" id="dialog_trash_file">
+    <property name="heading" translatable="yes">Trash File</property>
+    <property name="body" translatable="yes">Are you sure you want to delete this file?</property>
+    <property name="default-response">trash</property>
```
flxzt, line 210: The default selected response should be to cancel

crates/rnote-ui/src/dialogs/mod.rs
```diff
@@ -552,6 +553,38 @@ pub(crate) async fn dialog_edit_selected_workspace(appwindow: &RnAppWindow) {
...
+                let Some(current_file) = filerow.current_file() else {
+                    return;
+                };
+                current_file.trash_async(
+                    glib::source::Priority::DEFAULT,
+                    None::<&gio::Cancellable>,
+                    clone!(@weak filerow, @strong current_file => move |res| {
+                        if let Err(e) = res {
+                            appwindow.overlays().dispatch_toast_error(&gettext("Trashing file failed"));
+                            debug!("Trash filerow file `{current_file:?}` failed , Err: {e:?}");
```
flxzt, line 574: This should be an error event

crates/rnote-ui/src/dialogs/mod.rs
```diff
@@ -552,6 +553,38 @@ pub(crate) async fn dialog_edit_selected_workspace(appwindow: &RnAppWindow) {
     dialog.present(appwindow);
 }
 
+pub(crate) async fn dialog_trash_file(appwindow: &RnAppWindow, filerow: &RnFileRow) -> () {
```
flxzt, line 556: Instead of passing the filerow, directly pass the file. Then the implementation of the dialog is decoupled from the filerow. You'll need to pass the file as a `@strong` reference to all closures then

flxzt (commented):
Do we actually need to prompt when trashing in all cases?

It's a bit redundant because `trash` should already be recoverable
(in contrast to *deletion*).

Maybe it's best to only display the dialog for folders that are not empty to avoid mistakes like the one from the original issue?

crates/rnote-ui/data/ui/dialogs/dialogs.ui
RayJW: “This property determines what happens when the Return key is pressed while the alert is shown.”, on enter any file manager or application I tested always confirms deletion, and this is also the behaviour I would expect. Because otherwise, how does one confirm without mouse input? On escape, however, we definitely would want to cancel.

RayJW:
> Do we actually need to prompt when trashing in all cases?
> 
> It's a bit redundant because `trash` should already be recoverable (in contrast to _deletion_).
> 
> Maybe it's best to only display the dialog for folders that are not empty to avoid mistakes like the one from the original issue?

I was thinking about this as well. However, the error in the issue was mainly the user deleting something they didn't want to delete and not knowing that it just gets moved to the trash. I could see the exact same issue being reported with someone accidentally deleting a really important file, so I don't think only adding it for folders is the user friendliest solution.

crates/rnote-ui/data/ui/dialogs/dialogs.ui
flxzt: At least the Gnome [Human Interface Guidelines](https://developer.gnome.org/hig/patterns/feedback/dialogs.html#dialogs) suggest to make cancelling the default for destructive actions (last paragraph).

Although admittedly Rnote is already not following that consistently with the "Clear Document" button.

The keyboard focus should be on the default action button so it would take an additional right nav-key

RayJW: Okay, fair enough, it's your call, I don't have an issue with changing it. It is however weird because it's the same for Nautilus. If you shift + del to permanently delete something and press enter, it confirms the deletion.

crates/rnote-ui/src/dialogs/mod.rs
RayJW: I was thinking the same, that's however how the current trash action handled it, which is why I kept this behaviour. I'll change that.

crates/rnote-ui/data/ui/dialogs/dialogs.ui
flxzt: I don't have a strong opinion about it either but I usually stick to the HIG . I'll change the `Clear Document` action to it as well.

Curious that Nautilus as the core file mangement app doesn't do that - especially when the action actually is permanent and possibly very destructive.

crates/rnote-ui/src/dialogs/mod.rs
flxzt: thanks!

crates/rnote-ui/data/ui/dialogs/dialogs.ui
RayJW: Alright, I'll change it as well. I agree, I wasn't aware that's how the HIG handles it, I just thought about the default behaviour in file managers and stuck to that. But it's changed now :)

crates/rnote-ui/src/dialogs/mod.rs
RayJW: Done, thanks for the tip. I initially tried getting it to work like this, but had issues with the weak reference.

crates/rnote-ui/data/ui/dialogs/dialogs.ui
flxzt: Alright, thanks :)

flxzt:
> I was thinking about this as well. However, the error in the issue was mainly the user deleting something they didn't want to delete and not knowing that it just gets moved to the trash. I could see the exact same issue being reported with someone accidentally deleting a really important file, so I don't think only adding it for folders is the user friendliest solution.

Alright, that makes sense. Let's keep it for all files and folder.

>Edit: I was also thinking maybe the wording should be something along the lines of “Are you sure you want to move this file to the trash?” to make it clearer that it's not actually permanently deleted.

Yes that reads better to me!

I just noticed an unrelated issue with the dialog, it seems to be working only once for an individual file row. When it is cancelled and then tried to be opened a second time it just does not appear

RayJW:
> > Edit: I was also thinking maybe the wording should be something along the lines of “Are you sure you want to move this file to the trash?” to make it clearer that it's not actually permanently deleted.
> 
> Yes that reads better to me!

Alright, fixed it ^^

> I just noticed a different issue with the dialog, it seems to be working only once for an individual file row. When it is cancelled and then tried to be opened a second time it just does not appear

That's weird. You're right, I did some testing, but I have no idea what could be the cause. It should get correctly destroyed when it gets cancelled IMO.

crates/rnote-ui/src/workspacebrowser/filerow/actions/trash.rs
```diff
@@ -1,30 +1,19 @@
...
-        }),
-    );
+    action.connect_activate(clone!(@weak appwindow, @weak filerow => move |_, _| {
+        let Some(current_file) = filerow.current_file() else {
+            return;
+        };
+        glib::spawn_future_local(clone!(@weak appwindow, @strong current_file => async move {
+            dialogs::dialog_trash_file(&appwindow, &current_file).await;
+        }));
+        filerow.set_current_file(None);
```
flxzt, line 16: Could it be that this is the cause that the dialog only appears once? I think `dialog_trash_file` would need to give back if the file was actually deleted and only based on that set the file to `None`.

RayJW: Actually, if I just leave this out it seems to work fine. Do you see any problems with that? Worst case I'd just overwrite the value on the next action, no? The reason I moved all the logic into the dialogue in the first place was that it seemed kinda cumbersome to communicate from the dialogue back what choice was made.

Also, going to close this PR in favour of a new one since the merge conflicts are too time-consuming to properly fix for me.

flxzt: You're right, it probably is fine to leave it out entirely

Issue #1199, opened 2024-09-01
exporting selection doesn't work


ashish111333:
**Describe the bug**  
made a sketch and tried exporting it to png,svg using selection tool but it fails and doesn't export ...no error is show in the UI like "exporting failed" it just doesn't export.

**To Reproduce**  
simply select your sketch using all strokes and try to export it in png,svg it doesn'k work,fails without warning.

**Expected behavior**  
I should be able to export the selected area of my sketch using selection tool to the formats png,svg.

**Console Output**  
(rnote:2): Gtk-WARNING **: 00:31:33.440: No IM module matching GTK_IM_MODULE=cedilla found

(rnote:2): Gtk-CRITICAL **: 00:33:14.868: The layout manager of type GtkBoxLayout 0x5593ecd63090 does not create GtkLayoutChild instances

**Screenshots**  
![ksnip_20240902-003943](https://github.com/user-attachments/assets/18968dcc-1556-4aaf-8d5f-75a9ee43e689)

selected this whole sketch and tried to export it to png,svg it fails.

![ksnip_20240902-003948](https://github.com/user-attachments/assets/26dda7b8-0f80-47ee-9851-76736835e949)

selected this whole sketch and tried to export it to png,svg it fails.

**Desktop (please complete the following information):**  
 - OS: [suse tumbleweed]
 - App Version: [ Rnote v0.11.0]
 - Installation Source: [Flatpak]
 - Desktop Environment: [kde 6.1.4]
 - Display Server: [ X11]
 - Input Source: [ External Graphics Tablet,Huion]

Doublonmousse:
Is it only that mode or all export modes (export document or export document page) that don't work with any file ?
Is this on all or some file types (for selections there is svg, png and jpeg export type) ?
Did you choose a file to save to (as of now you always have to select a file first in the `export file` option) ? (If you don't the export button is greyed out so it won't trigger the export)
Normally there should always be either a confirmation or error message on all export options.

Could that be a flatpak file permission issue ?

ashish111333:
> Is it only that mode or all export modes (export document or export document page) that don't work with any file ? Is this on all or some file types (for selections there is svg, png and jpeg export type) ? Did you choose a file to save to (as of now you always have to select a file first in the `export file` option) ? (If you don't the export button is greyed out so it won't trigger the export) Normally there should always be either a confirmation or error message on all export options.
> 
> Could that be a flatpak file permission issue ?

[...]

PR #1201, opened 2024-09-02
Export selection potential fix

Doublonmousse:
make export selection match the other output modes

There is a minute difference.
On the two other modes we have
`if let Some(file)`
then we call `span_future_local`

This was the reverse for the selection export.
Trying to see if this fixes https://github.com/flxzt/rnote/issues/1199.
I've added a few debug trace to help with debugging for this issue

(it will only be a style PR otherwise)

### DIFF
--- a/crates/rnote-ui/src/dialogs/export.rs
+++ b/crates/rnote-ui/src/dialogs/export.rs
@@ -853,47 +853,51 @@ pub(crate) async fn dialog_export_selection_w_prefs(appwindow: &RnAppWindow, can
     export_selection_button_confirm.connect_clicked(clone!(@weak selected_file, @weak dialog, @weak canvas, @weak appwindow => move |_| {
         dialog.close();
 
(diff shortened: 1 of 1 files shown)


Doublonmousse:
This needs to be rewritten to the new `glib::clone` syntax. Maybe things _would_ work with the new glib version but in the current state, this was rewritten to the new syntax with potentially the same bug.

The fix was confirmed to work.

flxzt:
isn't the actual fix to change the reference taken by the closure either to a strong reference [here](https://github.com/flxzt/rnote/blob/f51a90f5ff1eb884553a4010efcdb18dda07cbc3/crates/rnote-ui/src/dialogs/export.rs#L1036) or completely remove it and move ownership into the selected file into the confirm button callback? I assume saving fails silently because upgrading the reference fails

flxzt:
I'll follow with a fix and some code cleanup and improved consistency

PR #1209, opened 2024-09-08
Build Rnote and Installer for Windows Arm64

OB0NE:
I made some adjustments to build Rnote for Windows ARM64 and put together a guide to outline the steps I followed. The **Rnote.exe** and **windows_arm64_installer.exe** files are located in the **_mesonbuild** folder.

Since I have very limited experience in building applications, my adjustments were made by following **rnote-windows-build.md** as best as I could. This is also my first attempt at contributing to a project, so if you find any errors or see room for improvement, please feel free to make any necessary changes.

I’ve only been able to test the build on a Microsoft Surface Pro 11, so I'm unsure how it will perform on other ARM devices.

**!!! I’ve made some significant modifications to a few key files, so it might be best to create a new branch for these changes. !!!**


Doublonmousse:
It shouldn't break the x86 build process. Maybe you can add a `win-arm64-build-environment-path` option in the meson option file ?
Then have both arm and x86 in the inno file and a check in the build.py file the architecture (I wonder if the rust flag can be set in meson instead).

Also of note : your upstream issue about black borders should be fixed on 4.16.0, maybe you can double check whether this is the case on your side  (not sure the package is up to date on mingw yet though)

Issue #375, opened 2022-11-25
[Feature Request] Shapes with dotted lines


astro-ray:
**Please submit only one feature request in one issue!**

**Describe the solution you'd like**
I would like it if there were shapes with dotted/dashed lines along with solid lines. Like dashed straight lines/rectangles/ellipses

aidanhadley:
i think this would be very beneficial! +1

PR #1210, opened 2024-09-08
Styled/Dashed Lines for Shapes

### LINKED ISSUE #852: Different line styles
Currently, you can only draw dotted, and dashed lines by manually drawing them (lifting and placing the pen many times). A side effect of this is that to erase the line, it is not treated as one stroke, and thus has to be entirely erased manually instead of all at once.

It would be nice to have differente line styles like dotted, or dashed that one can draw using the pen. Each of these line styles would be treated as a single stroke, so when erased they would erase the entire line.

Xournal++ has this feature for example.

**Additional context**  
N/A

anesthetice:
Dashed lines are pretty easy to create using piet's StrokeStyle. I think the main issue is how all of this should be integrated into Rnote's UI.

would fix #852 and #375

### DIFF
--- a/crates/rnote-compose/src/style/smooth/mod.rs
+++ b/crates/rnote-compose/src/style/smooth/mod.rs
@@ -1,8 +1,10 @@
 // Modules
 mod smoothoptions;
 
+use piet::StrokeStyle;
 // Re-exports
 pub use smoothoptions::SmoothOptions;
+use tracing::info;
 
 // Imports
 use super::Composer;
@@ -26,7 +28,13 @@ impl Composer<SmoothOptions> for Line {
 
         if let Some(stroke_color) = options.stroke_color {
             let stroke_brush = cx.solid_brush(stroke_color.into());
-            cx.stroke(line, &stroke_brush, options.stroke_width);
+            let stroke_style = StrokeStyle::new()
+                .dash_pattern(&[2.0, 3.0])
+                .line_cap(piet::LineCap::Round)
+                .dash_offset(0.0);
+
+            info!("{:#?}", stroke_style);
+            cx.stroke_styled(line, &stroke_brush, options.stroke_width, &stroke_style);
         }
         cx.restore().unwrap();
     }


anesthetice:
Good suggestions, things should be kept relatively simple, concerning more complicated choices, they are somewhat possible, for example we could use a StrokeDash ~ [0, 3, 1, 3] and a round LineCap to create a dot-dash line

### Implementation

Trickier than I initially thought, but still nonetheless possible, here's what I'm currently working on

* adding a shape_style entry to SmoothOptions `shape_style: Option<ShapeStyle>`
* ShapeStyle being a complex wrapper around piet::StrokeStyle
* ui: making a "Shaper Configuration"



Also considering adding a simple dashed line to the possible shapes

### Tricky parts

**1. piet::StrokeStyle contains piet::StrokeDash**
```
/// A type that represents an alternating pattern of drawn and undrawn segments.
///
/// We use our own type as a way of making this work in `const` contexts.
///
/// This type `Deref`s to `&[f64]`.
#[derive(Debug, Default, Clone, PartialEq)]
pub struct StrokeDash {
    slice: &'static [f64],
    alloc: Option<Rc<[f64]>>,
}
```
Rc<T> is neither Send nor Sync, which is required for items inside SmoothOptions

Currently trying to address this by using the Fragile crate, https://crates.io/crates/fragile, I thinks it's lightweight enough to add to rnote-compose (0 additional dependencies, read through the Fragile implementation, seems sound)

**2. consistency**

Hard to clearly explain this part, so I'll just be using an example:
> _Imagine you want a simple, pretty dashed line with rounded edges_
unlike piet::LineCap::Butt, piet::LineCap::Round extends the length of a line in proportion to it's width, this will not be taken into consideration by piet::StrokeDash, it will have to be updated, neither will the width of the line itself, etc.

MarcA711:
It would be useful if one can also select the brush style for the shape as well. This way one can select the brush `marker` and draw the shape `line` to highlight text. This would look much cleaner than highlighting the text freehand.

Moreover, why only implement styled lines for shapes? It would be useful to draw dashed/dotted lines freehand as well. Maybe this could be implemented similar to how it's done in Xournal++: You can select a pen independently (brush style, size, color, line style) and in the next step you choose if you want to use this selected pen freehand or with a shape. Why should freehand and shapes have different pens? This would also integrate nicely with https://github.com/flxzt/rnote/pull/1175, because you can select the pen you want, draw something and use the hold feature to convert it to a shape.

anesthetice:
> It would be useful if one can also select the brush style for the shape as well. This way one can select the brush `marker` and draw the shape `line` to highlight text. This would look much cleaner than highlighting the text freehand.
> 
> Moreover, why only implement styled lines for shapes? It would be useful to draw dashed/dotted lines freehand as well. Maybe this could be implemented similar to how it's done in Xournal++: You can select a pen independently (brush style, size, color, line style) and in the next step you choose if you want to use this selected pen freehand or with a shape. Why should freehand and shapes have different pens? This would also integrate nicely with #1175, because you can select the pen you want, draw something and use the hold feature to convert it to a shape.

Good point, I'll see what I can do, under the hood pen strokes and shapes uses the same styles, so this might not be too complicated. I've also been thinking about having presets for pens and shapes, as having to switch back and forth within the settings is slightly inconvenient

anesthetice:
Quick update: my Piet PR has been merged, and I’ve implemented the original idea - adding piet::StrokeStyle directly to SmoothOptions. As for the marker proposal, I tested it, but I don’t think it’s the best solution to the issues mentioned earlier. The preview doesn’t work (Worse, I'm not sure it can work at all), and (as expected) it still draws over text in an imported PDF. Instead, I’m considering adding a 'solid' and 'highlighter' option, where the highlighter would simply adjust the alpha of the current color to make it more transparent.

anesthetice:
> the `ShaperBrushType` and `ShaperStyle` feel redundant. What I think would make more sense is to integrate the Highlighter style into `ShaperStyle` and simply use the selected color opacity for this style.

Yes that would probably be nicer, just to be clear you mean: make `Highlighter` a `ShaperStyle` (uses `smooth_options` + `highlighter_opacity` under the hood for the logic) and then just integrate the current `Highlighter` `AdwSpinRow` into the `ShaperStyle` `GtkListbox`?

Also could I have your thoughts on the following:

1. Moving everything inside the `Shape Style` popover into the `Shape Configuration` one
2. Using [0.00; 1.00] instead of [0; 100] for the highlighter opacity spin button
3. If it looks nicer, to only show the options for the active `Shape Style` and hide the rest?
4. Do you know of a way to make the content of an `AdwSpinRow` not be highlighted for manual editing when clicked on?

anesthetice:
<div style="display: flex; justify-content: space-between;">
  
  
  
</div>

@flxzt In the end I don't think creating and using a highlighter shaper style is worth it as it doesn't fit the style or logic very well, however using a highlight mode toggle is pretty clean, also hiding the "unused" smooth or rough options seems nice (to me at least) as they are the same size

flxzt:
I noticed the following unusual behaviour when activating the highlighter mode:
The fill of shapes seems to be affected when they are previously transparent. Maybe a special case for this needs to be added?

image

crates/rnote-engine/src/pens/pensconfig/shaperconfig.rs
```diff
@@ -80,13 +89,27 @@ impl ShaperConfig {
     pub(crate) fn gen_style_for_current_options(&self) -> Style {
         match &self.style {
             ShaperStyle::Smooth => {
-                let options = self.smooth_options.clone();
-
+                let mut options = self.smooth_options.clone();
```
anesthetice, line 92: @flxzt Caused by this, I'll see what I can do

crates/rnote-engine/src/pens/pensconfig/shaperconfig.rs
```diff
@@ -80,13 +89,27 @@ impl ShaperConfig {
...
+                if self.highlight_mode {
+                    if let Some(ref mut color) = options.stroke_color {
+                        color.a = self.highlight_opacity;
+                    }
+                    if let Some(ref mut color) = options.fill_color {
+                        color.a = self.highlight_opacity;
```
anesthetice, line 100: @flxzt caused by this, I'll add a check so that if the selected fill color is fully transparent then it is not affected by the highlight mode being toggled on

anesthetice:
> I noticed the following unusual behaviour when activating the highlighter mode: The fill of shapes seems to be affected when they are previously transparent. Maybe a special case for this needs to be added?
> image

Good catch, never noticed as my background is white and my "transparent" fill color was always `#FFFFFF` as well, should be fixed now!

crates/rnote-compose/src/style/smooth/smoothoptions.rs
```diff
@@ -1,11 +1,17 @@
 // Imports
 use crate::Color;
 use crate::style::PressureCurve;
+use anyhow::Context;
+use num_derive::{FromPrimitive, ToPrimitive};
 use serde::{Deserialize, Serialize};
+use std::{
+    f64,
+    ops::{AddAssign, MulAssign},
+};
```
flxzt, line 11:
```suggestion

#[derive(Debug, Deserialize)]
#[serde(default, rename = "smooth_options")]
struct SmoothOptionsPrecursor {
    #[serde(rename = "stroke_width", with = "crate::serialize::f64_dp3")]
    pub stroke_width: f64,
    #[serde(rename = "stroke_color")]
    pub stroke_color: Option<Color>,
    #[serde(rename = "fill_color")]
    pub fill_color: Option<Color>,
    #[serde(rename = "pressure_curve")]
    pub pressure_curve: PressureCurve,
    #[serde(rename = "line_style")]
    pub line_style: LineStyle,
    #[serde(rename = "line_cap")]
    pub line_cap: LineCap,
}

impl From<SmoothOptionsPrecursor> for SmoothOptions {
    fn from(value: SmoothOptionsPrecursor) -> Self {
        SmoothOptions {
            stroke_width: value.stroke_width,
            stroke_color: value.stroke_color,
            fill_color: value.fill_color,
            pressure_curve: value.pressure_curve,
            line_style: value.line_style,
            line_cap: value.line_cap,
            piet_stroke_style: Self::compute_piet_stroke_style(
                value.stroke_width,
                value.line_style,
                value.line_cap,
            ),
        }
    }
}

impl From<SmoothOptions> for SmoothOptionsPrecursor {
    fn from(value: SmoothOptions) -> Self {
        Self {
            stroke_width: value.stroke_width,
            stroke_color: value.stroke_color,
            fill_color: value.fill_color,
            pressure_curve: value.pressure_curve,
            line_style: value.line_style,
            line_cap: value.line_cap,
        }
    }
}

impl Default for SmoothOptionsPrecursor {
    fn default() -> Self {
        SmoothOptions::default().into()
    }
}

```

crates/rnote-compose/src/style/smooth/smoothoptions.rs
```diff
@@ -1,11 +1,17 @@
...
 
-/// Options for shapes that can be drawn in a smooth style.
-#[derive(Debug, Clone, Serialize, Deserialize)]
-#[serde(default, rename = "smooth_options")]
+/// Options for shapes that can be drawn in a smooth style. Ensure the precursor struct used in deserialization matches this one.
+#[derive(Debug, Clone, Serialize)]
+#[serde(rename = "smooth_options")]
```
flxzt, line 14:
```suggestion
#[serde(rename = "smooth_options", from = "SmoothOptionsPrecursor")]
```

crates/rnote-compose/src/style/smooth/smoothoptions.rs
```diff
@@ -19,15 +25,232 @@ pub struct SmoothOptions {
...
+            line_cap: precursor.line_cap,
+            piet_stroke_style: Self::compute_piet_stroke_style(
+                precursor.stroke_width,
+                precursor.line_style,
+                precursor.line_cap,
+            ),
+        })
+    }
+}
+
```
flxzt, line 171:
```suggestion

```

crates/rnote-compose/src/style/smooth/smoothoptions.rs
flxzt: what about this instead of manually implementing `Deserialize` ?

flxzt (commented):
what about this instead of manually implementing `Deserialize` ?

crates/rnote-compose/src/style/smooth/smoothoptions.rs
anesthetice: Hey, sorry for the delay, I'd be fine with implementing your suggested changes; it's actually how things originally were, but I decided to stuff the Precursor struct declaration in the "manual" `deserialize` implementation as it seemed "cleaner" (less clutter)

flxzt: okay, so let's leave it as-is then!

flxzt:
I only pushed some minor adjustments (serialization with explicit names to avoid accidental breakage in case rust types are renamed), and some cosmetic code changes (builder pattern for piet's StrokeStyle). Looks good though, I'll merge when the CI finished successfully

PR #1213, opened 2024-09-09
Capitalize file format names because they are acronyms.

yuckdevchan:
Capitalizing file format acronyms is standard practice in software.

This pull request replaces every occurence of Pdf, Svg and Xopp with PDF, SVG and XOPP respectively.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,7 +17,7 @@
 > Sketch and take handwritten notes.  
 
 Rnote is an open-source vector-based drawing app for sketching, handwritten notes and to annotate documents and pictures.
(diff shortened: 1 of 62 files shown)


flxzt:
I don't think the code should be changed.

The Gnome HIG unfortunately does not include this, but Microsoft has some documentation about the case of abbreviations and acronyms: 

https://learn.microsoft.com/en-us/previous-versions/dotnet/netframework-1.1/141e06ef(v=vs.71)?redirectedfrom=MSDN

The project should use camel case for all acronyms longer than two characters.

PR #1218, opened 2024-09-15
refactor: update piet, cairo, gtk based deps, switch to new glib::clone macro syntax

flxzt:
This bumps glib, cairo, glib, .. gtk based dependencies to the latest version.

In order to do that, we have to patch piet with the unfortunately not yet merged [pango-cairo](https://github.com/flxzt/piet/tree/bump-cairo-pango) branch.

After doing so, there were some minor adjustments to the code needed to compile and adapt to the new API. The most prominent change is that all uses of the `glib::clone` macro are ported to the new syntax which enables code formatting and integrates better with rust-analyzer (reference renaming, warnings when references are unused!)

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,12 +20,12 @@ version = "0.11.0"
 rnote-compose = { version = "0.11.0", path = "crates/rnote-compose" }
 rnote-engine = { version = "0.11.0", path = "crates/rnote-engine" }
 
-adw = { version = "0.6.0", package = "libadwaita", features = ["v1_5"] }
+adw = { version = "0.7.0", package = "libadwaita", features = ["v1_5"] }
 anyhow = "1.0"
 approx = "0.5.1"
 async-fs = "2.1"
 base64 = "0.22.1"
-cairo-rs = { version = "0.19.4", features = ["v1_18", "png", "svg", "pdf"] }
+cairo-rs = { version = "0.20.1", features = ["v1_18", "png", "svg", "pdf"] }
 chrono = "0.4.38"
 clap = { version = "4.5", features = ["derive"] }
 dialoguer = "0.11.0"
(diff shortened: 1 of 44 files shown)


RayJW:
Although it might be worth opening a new issue or discussion for this, at what point might it be worth looking into slowly trying to branch into a [vello](https://github.com/linebender/vello) (or similar for that matter) port? It looks like piet might not see a new (feature) release ever, so things like this might just become more tedious as time goes on.

Doublonmousse:
@RayJW Not before we manage to make the gpu part work. For this to work we need to find a way to make wgpu work inside of a gtk window.

And for now this is not solved. There may be a way on 4.14 (and 4.16) using dma buf and the graphics offload but that's only supported on linux wayland (but maybe other in the future).

See https://github.com/gtk-rs/gtk4-rs/issues/1278
Also https://github.com/Doublonmousse/wgpu-dma-buf (maybe it works, still some work to do, and on open gl es, not vulkan). In this case, any help is welcome !

RayJW:
> @RayJW Not before we manage to make the gpu part work. For this to work we need to find a way to make wgpu work inside of a gtk window.
> 
> And for now this is not solved. There may be a way on 4.14 (and 4.16) using dma buf and the graphics offload but that's only supported on linux wayland (but maybe other in the future).
> 
> See [gtk-rs/gtk4-rs#1278](https://github.com/gtk-rs/gtk4-rs/issues/1278) Also [Doublonmousse/wgpu-dma-buf](https://github.com/Doublonmousse/wgpu-dma-buf) (maybe it works, still some work to do, and on open gl es, not vulkan). In this case, any help is welcome !

Thanks for the insight! That's why I asked, I just stumbled upon the same issues with piet, hence discovering its deprecation. But I see there is still some work left to get rid of piet. It might probably still be some time before requiring 4.16 is viable.

Cargo.toml
```diff
@@ -84,12 +84,15 @@ usvg = "0.43.0"
...
 xmlwriter = "0.1.0"
 # Enabling feature > v20_9 causes linker errors on mingw
-poppler-rs = { version = "0.23.0", features = ["v20_9"] }
+poppler-rs = { version = "0.24.1", features = ["v20_9"] }
 
 [patch.crates-io]
 # once a new piet (current v0.6.2) is released with updated cairo and kurbo deps, this can be removed.
-piet = { git = "https://github.com/linebender/piet", rev = "02eb5f0152e893626c43980bf37eeb0ccb1acb46" }
-piet-cairo = { git = "https://github.com/linebender/piet", rev = "02eb5f0152e893626c43980bf37eeb0ccb1acb46" }
+piet = { git = "https://github.com/flxzt/piet", rev = "a480bc20bd0d235928d46b009dae54e52e32767a" }
```
anesthetice, line 91: kurbo's version seems to be updated in a later commit, https://github.com/flxzt/piet/commits/bump-cairo-pango/, you could specify `branch = "bump-cairo-pango"`

on a side note, could you add this commit https://github.com/linebender/piet/pull/572/commits/091ddfd1fa92c3454cff0294be3ce2d835d95556 to the branch :eyes:

flxzt: oops, copied the wrong commit. Fixed it now.

I don't want to directly add your changes to the existing version-bump branch.
Maybe we could wait for some feedback and if it is clear that it will be merged,
we can include it.
But if we would add it now, and implement the dash feature based on your feature branch, we would possibly have to reverse the changes before doing a release. Because I want to remove the patch before releasing a new version.

flxzt:
I can't reasonably ask anyone to review all code changes in this PR, so I'll just be optimistic and merge

Issue #819, opened 2023-10-02
Erase faster to increase eraser size


Mitia99:
**Feature request**  
Erase faster to increase eraser size

https://github.com/flxzt/rnote/assets/56766778/687e4490-36c8-4254-9ed9-db253fafcbf7

**Additional context**  
It's available on [Microsoft Whiteboard](https://apps.microsoft.com/store/detail/microsoft-whiteboard/9MSPC6MP8FM4?hl=en-mg&gl=mg&rtc=1)

I find it pretty convenient and the gesture feels natural.

Thanks for considering this,

PerfectlyInternal:
I'd be willing to take on this feature, it looks like nobody else has started on it.

Doublonmousse:
@PerfectlyInternal See also https://github.com/flxzt/rnote/discussions/957

Not sure things have moved since then (I think it's free for you to take), but the help there should still be pretty much valid.

PR #1235, opened 2024-09-25
Erase faster to increase eraser size

PerfectlyInternal:
Eraser size now scales linearly with pen speed. Implements #819.

### DIFF
--- a/crates/rnote-engine/src/pens/eraser.rs
+++ b/crates/rnote-engine/src/pens/eraser.rs
@@ -27,7 +27,8 @@ pub struct EraserMotion {
 }
 
 impl EraserMotion {
-    pub const SMOOTHING_FACTOR: f64 = 3.0;
+    pub const SMOOTHING_FACTOR_ACCEL: f64 = 3.0;
+    pub const SMOOTHING_FACTOR_DECEL: f64 = 10.0;
     pub const SPEED_LIMIT: f64 = 10000.0;
 
     fn update(&mut self, element: Element, time: Instant) {
@@ -35,9 +36,17 @@ impl EraserMotion {
             let delta = element.pos - last_element.pos;
             let delta_time = time - last_element_time;
             let new_speed = delta.norm() / delta_time.as_secs_f64();
-            self.speed = Self::SPEED_LIMIT.min(
-                (self.speed * Self::SMOOTHING_FACTOR + new_speed) / (Self::SMOOTHING_FACTOR + 1.0),
-            );
+            if new_speed > self.speed {
+                self.speed = Self::SPEED_LIMIT.min(
+                    (self.speed * Self::SMOOTHING_FACTOR_ACCEL + new_speed)
+                        / (Self::SMOOTHING_FACTOR_ACCEL + 1.0),
+                );
+            } else {
+                self.speed = Self::SPEED_LIMIT.min(
+                    (self.speed * Self::SMOOTHING_FACTOR_DECEL + new_speed)
+                        / (Self::SMOOTHING_FACTOR_DECEL + 1.0),
+                );
+            }
         }
         self.last_element = Some((element, time));
     }
--- a/crates/rnote-engine/src/pens/eraser.rs
+++ b/crates/rnote-engine/src/pens/eraser.rs
@@ -22,8 +22,7 @@ pub enum EraserState {
 
 #[derive(Debug, Default, Clone, Copy)]
 pub struct EraserMotion {
-    last_element: Option<Element>,
-    last_element_time: Option<Instant>,
+    last_element: Option<(Element, Instant)>,
     pub speed: f64,
 }
 
@@ -31,22 +30,18 @@ impl EraserMotion {
     pub const SMOOTHING_FACTOR: f64 = 3.0;
 
     fn update(&mut self, element: Element) {
-        if let Some(last_element) = self.last_element {
-            if let Some(last_element_time) = self.last_element_time {
-                let delta = element.pos - last_element.pos;
-                let delta_time = Instant::now() - last_element_time;
-                let new_speed = delta.norm() / delta_time.as_secs_f64();
-                self.speed = (self.speed * Self::SMOOTHING_FACTOR + new_speed)
-                    / (Self::SMOOTHING_FACTOR + 1.0);
-            }
+        if let Some((last_element, last_element_time)) = self.last_element {
+            let delta = element.pos - last_element.pos;
+            let delta_time = Instant::now() - last_element_time;
+            let new_speed = delta.norm() / delta_time.as_secs_f64();
+            self.speed =
+                (self.speed * Self::SMOOTHING_FACTOR + new_speed) / (Self::SMOOTHING_FACTOR + 1.0);
         }
-        self.last_element = Some(element);
-        self.last_element_time = Some(Instant::now());
+        self.last_element = Some((element, Instant::now()));
     }
 
     fn reset(&mut self) {
         self.last_element = None;
-        self.last_element_time = None;
         self.speed = 0.0;
     }
 }
--- a/crates/rnote-engine/src/pens/eraser.rs
+++ b/crates/rnote-engine/src/pens/eraser.rs
@@ -28,14 +28,16 @@ pub struct EraserMotion {
 
 impl EraserMotion {
     pub const SMOOTHING_FACTOR: f64 = 3.0;
+    pub const SPEED_LIMIT: f64 = 10000.0;
 
     fn update(&mut self, element: Element, time: Instant) {
         if let Some((last_element, last_element_time)) = self.last_element {
             let delta = element.pos - last_element.pos;
             let delta_time = time - last_element_time;
             let new_speed = delta.norm() / delta_time.as_secs_f64();
-            self.speed =
-                (self.speed * Self::SMOOTHING_FACTOR + new_speed) / (Self::SMOOTHING_FACTOR + 1.0);
+            self.speed = Self::SPEED_LIMIT.min(
+                (self.speed * Self::SMOOTHING_FACTOR + new_speed) / (Self::SMOOTHING_FACTOR + 1.0),
+            );
         }
         self.last_element = Some((element, time));
     }
--- a/crates/rnote-engine/src/pens/eraser.rs
+++ b/crates/rnote-engine/src/pens/eraser.rs
@@ -29,15 +29,15 @@ pub struct EraserMotion {
 impl EraserMotion {
     pub const SMOOTHING_FACTOR: f64 = 3.0;
 
-    fn update(&mut self, element: Element) {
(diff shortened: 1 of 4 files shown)


crates/rnote-engine/src/pens/eraser.rs
```diff
@@ -20,15 +20,48 @@ pub enum EraserState {
     Down(Element),
 }
 
+#[derive(Debug, Default, Clone, Copy)]
+pub struct EraserMotion {
+    last_element: Option<Element>,
+    last_element_time: Option<Instant>,
```
Doublonmousse, line 26: I wonder if it makes sense to have only one tuple of Element, Instant instead of two separate elements given both will be either None or Some.

PerfectlyInternal: That would make sense, yeah

Doublonmousse:
Trying this, I feel like there is some additional work to be done.

There is no upper limit to the eraser size. When switching from the pen to the eraser, the initial size can sometimes be too large, even exceeding the visible canvas size (and delete a whole document instead of staying as a small eraser). It might be worse on some devices (an initial small displacement over a very small timestep will give a velocity estimation with very large error bounds, and that timestep might be 1/60 s on some pens down to 1/480 s. Hence the potential very large overestimate even with the initial smoothing). This should be fixed.

An upper bound on speed or eraser size should also be here.

Maybe we can try to re use the ink stroke modeler for this (but output the velocity estimation instead) or do some kalman based filtering (where we can model the error on the speed estimation and act on it if necessary).

The other thing I'm wondering is that on other software that scales the eraser size depending on speed, the actual behavior is not doing a speed estimation and scaling the eraser size with respect to that. There is some inertia to it, or some way to lock or decrease smoothly or in steps the eraser size when the speed decreases. 
It might be more akin to a "erase faster means getting the next eraser size and locking it to that size until you stop or slow down" situation.
Realistically, we're not obliged to follow suit on that, but we should come up with a behavior that makes sense and will satisfy the original feature request.

crates/rnote-engine/src/pens/eraser.rs
```diff
@@ -20,15 +20,43 @@ pub enum EraserState {
...
+    pub speed: f64,
+}
+
+impl EraserMotion {
+    pub const SMOOTHING_FACTOR: f64 = 3.0;
+
+    fn update(&mut self, element: Element) {
+        if let Some((last_element, last_element_time)) = self.last_element {
+            let delta = element.pos - last_element.pos;
+            let delta_time = Instant::now() - last_element_time;
```
Doublonmousse, line 35: Actually this is not the correct way to do this. You shouldn't use `Instant::now` here. 
The reason is that on some devices, the pen refreshes faster than the screen rate and gtk4 sends all the events that happened between two time frames once and regroups them. Hence you can get more than one update with basically the same `Instant::now` but that happened at different time steps.

You have to propagate to `update` the Instant the pen event actually happens.

See the `_now` argument from here : 
https://github.com/flxzt/rnote/blob/0964cb5d34077a5d36346f6a4b78b90b5cdd5533/crates/rnote-engine/src/pens/eraser.rs#L84
Remove the `_` (it's going to be used) and propage the `now` variable to the `update` function. This is the one you're going to use instead of the `Instant::now`.

Doublonmousse: This might be why I had issues. Fix this part and I will retest whether or not the issue persist on my side

PerfectlyInternal: Should be fixed now, thanks for catching that!

PerfectlyInternal:
In the video shown in the issue, the eraser seems to step in size as the user moves quickly for a while, then gradually step down once the user has stopped moving. This could be implemented either as smoothing over a very long period of time, so the eraser needs to be moved quickly for a while in order to get to its full size, or by checking if the eraser has been moving fast for at least X ms, then stepping up the size. Which one do you think would be better?

Doublonmousse:
I'm still getting the same issue with the start size being sometimes too large then going down. I will try to spend some times logging things to check where this comes from.

Doing things when the user stops moving is also a little harder to do. With pens, you always have some jitter hence events continue happening (but events stop happening if you stop with a mouse). This is the same issue I had with #1175 (needing a separate thread to call changes in the future and being cautious not to have more than one event fire at the same time because of it). But maybe in that case that's okay not to do that (as you trigger the change of size with movement, not stopping the pen)

Maybe we can use a better designed speed to eraser size function (somewhat of a stepped function so you get the same size for some range of speed) ? 
The other idea I had was to add a `size` to the `EraserMotion` structure and have it be updated by some well chosen function like `size = f(size, speed)` so that we get some hysteresis-like behavior (lock/converge to some size levels and change value significantly only for large/slow speed).

Honestly I don't have a perfect answer, though I want to check whether if we can get what we want with the current strategy (smoothing). I don't know what's the better approach here.

PerfectlyInternal:
The eraser now needs to exceed some minimum speed in order to start scaling up, and scaling is now a stepped function with hysteresis. The exact values to trigger stepping up/down can be tweaked, so we'll want to play around with those to get something that feels good to use. It might even be a good idea to expose them as config options for the user, since different users will probably have different speeds at which they'll want to trigger eraser scaling.

We still need some way of detecting a lack of motion, since events stop firing when you hold still, even on a drawing tablet. This will need to be faster than the long press detector you already have, but the underlying logic should be similar.

Doublonmousse:
Look at https://github.com/flxzt/rnote/commit/f20775cab7bddacbc2086cbda0b4424b4634c140#diff-4dd2da25347b4ca65437059e5072856b3085e9c029be3a91a1a6b140fabaf47aR493 to make it compile.

PerfectlyInternal:
FIxed it, thanks!

Doublonmousse:
Sorry, it's been a while since I've looked at the PR.
I still get the bug of the eraser being very large when the pen touches the screen (do we need or can we estimate correctly the speed when the pen is hovering ? maybe there is something happening here ?)

I think I need to take the time to debug this properly (and maybe extract some representative raw input sequences that we can use to fine tune the program. I wonder if it's of interest to add something like that to the dev menu for future uses as well : record raw/smoothed inputs for pen and eraser and save to a file/json for analysis). To be clear, I'm not asking you for this, I'll probably do a separate PR for this.

PerfectlyInternal:
I think what might help is calculating pen speed as part of input handling, so we can filter the input more, and possibly use it for other features later down the line.

crates/rnote-engine/src/pens/eraser.rs
```diff
@@ -53,14 +111,14 @@ impl PenBehaviour for Eraser {
...
+        now: Instant,
         engine_view: &mut EngineViewMut,
     ) -> (EventResult<PenProgress>, WidgetFlags) {
         let mut widget_flags = WidgetFlags::default();
-
         let event_result = match (&mut self.state, event) {
             (EraserState::Up | EraserState::Proximity { .. }, PenEvent::Down { element, .. }) => {
-                widget_flags |= erase(element, engine_view);
+                self.motion.update(element, now);
+                widget_flags |= erase(element, self.motion.speed, engine_view);
```
Doublonmousse, line 121: Speed instead of `added_width` here

crates/rnote-engine/src/pens/eraser.rs
```diff
@@ -94,8 +154,9 @@ impl PenBehaviour for Eraser {
                 }
             }
             (EraserState::Down { .. }, PenEvent::Up { element, .. }) => {
-                widget_flags |=
-                    erase(element, engine_view) | engine_view.store.record(Instant::now());
+                self.motion.reset();
+                widget_flags |= erase(element, self.motion.speed, engine_view)
```
Doublonmousse, line 158: Same thing, speed instead of `added_width`

crates/rnote-engine/src/pens/eraser.rs
```diff
@@ -167,7 +230,7 @@ impl DrawableOnDoc for Eraser {
                 engine_view
                     .pens_config
                     .eraser_config
-                    .eraser_bounds(*current_element),
+                    .eraser_bounds(*current_element, self.motion.speed),
```
Doublonmousse, line 233: Same thing

Doublonmousse:
The bug was caused by the eraser bounds still using the estimated speed instead of the added_width.
It now works correctly. Maybe we can fine tune parameters a little bit?

As far as calculating speed higher in the input handling code, it may not be worth estimating speed for nothing. I think it's the right place for it. We may mutualize some code between here and the ink stroke modeler in the future if we are to implement a finer velocity estimation method (though I think this already works pretty well)

Doublonmousse:
https://github.com/flxzt/rnote/blob/f7adb2acc0277f549db5f0de8764d1dff9ea6989/crates/rnote-engine/src/pens/eraser.rs#L190-L191
Missing a `self.motion.reset()`.

I'm not convinced we need to estimate speed while the eraser is in proximity mode. If we don't, what we can do is change the `EraserState` to have 
```rs
   Down(Element, EraserMotion)
```
Doing that means we don't have to do a `reset` as the `EraserMotion` will only be there if we are in the Down mode

crates/rnote-engine/src/pens/eraser.rs
```diff
@@ -20,15 +20,73 @@ pub enum EraserState {
...
+    pub const SMOOTHING_FACTOR_DECEL: f64 = 10.0;
+    pub const SPEED_LIMIT: f64 = 10000.0;
+    pub const MIN_SPEED: f64 = 100.0;
+    pub const SPEED_SCALING: f64 = 0.01;
+    pub const SPEED_SCALING_STEP_SIZE: f64 = 25.0;
+
+    fn update(&mut self, element: Element, time: Instant) {
+        if let Some((last_element, last_element_time)) = self.last_element {
+            let delta = element.pos - last_element.pos;
+            let delta_time = time - last_element_time;
```
Doublonmousse, line 41: Actually, after looking at my issue (with large speed jumps), it's not impossible to get a delta time that's either too small or negative (and in that case the - clamps the value to zero).

We should only update the speed if the delta_time > 0.0 or superior to an expected minimum time delta.
From https://github.com/flxzt/rnote/blob/7f3ff8aafcaf3137bec9993329667e5e9fbf4c4e/crates/rnote-engine/src/pens/penholder.rs#L497 (for each screen frame, gtk send all the events in one go and we only take events so that they are spread apart by at least the value here), this allows only one update per screen frame, so something like 8ms at 120 Hz.

Maybe a threshold at 4 ms would be good.

This will fix this issue for good

flxzt:
do we really need this? It makes the impression of feature creep to me.
For my personal use-case, an eraser that does not change it's size was perfectly fine.

flxzt:
Having in mind to keep the app simple, I am hesitant to accept this change.
Also the author did not respond to the review anymore

PR #1239, opened 2024-09-29
Document inserted at viewport when adjust_document is true #1088

### LINKED ISSUE #1088: A Way to append PDFs to a Document
Currently, when Importing a PDF into a document, it either loads as an Image or replaces / covers existing pages.
It would make sense if there was an option to append the PDF at the end of the document / inserted as new pages instead of covering the first few pages.

strang3nt:
Small edit that fixes #1088.
The commit makes it so that when `adjust_document` is on the PDF is positioned at current viewport instead of position zero.

### DIFF
--- a/crates/rnote-engine/src/engine/import.rs
+++ b/crates/rnote-engine/src/engine/import.rs
@@ -310,8 +310,6 @@ impl Engine {
     /// Generate image strokes for each page for the bytes.
     ///
     /// The bytes are expected to be from a valid Pdf.
-    ///
-    /// Note: `insert_pos` does not have an effect when the `adjust_document` import pref is set true.
     #[allow(clippy::type_complexity)]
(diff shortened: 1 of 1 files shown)


flxzt:
this kind of defeats the purpose of the "adjust document" option.

This would need to be implemented differently to reflect the desire to append to the existing document, either in a new option or by improving the positioning logic in some way

strang3nt:
Ok I get it. I came up with the following two options:

1. Improve the positioning logic in such a way that if the document is not blank the pdf is appended, otherwise it has the current behavior (positioned at the top of the document)
2. add a new option ("Append to document" maybe) which, if toggled, appends the pdf at the very end of the document. This new option is below the adjust-document option, and it is toggle-able only if adjust document itself is toggled. 

I personally prefer the second option, which seems clearer from a user perspective

PR #1242, opened 2024-10-01
Bump GNOME runtime and LLVM version

RayJW:
This is just pure maintenance PR. This bumps the GNOME runtime to 47 and LLVM to 18. I tested both changes and everything works fine. I've used the changes for a bit doing my regular work with all kinds of different tools and tasks, everything seemed fine and maybe even a bit smoother (although that might easily be placebo). It probably still makes sense to test this for a bit before applying to the Flatpak, which is why I opened the PR here instead of on Flathub. Although I've never seen a runtime change break something if building worked fine.

### DIFF
--- a/.github/workflows/nightly.yml
+++ b/.github/workflows/nightly.yml
@@ -10,7 +10,7 @@ jobs:
   flatpak:
     runs-on: ubuntu-latest
     container:
-      image: bilelmoussaoui/flatpak-github-actions:gnome-46
+      image: bilelmoussaoui/flatpak-github-actions:gnome-47
       options: --privileged
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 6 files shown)


RayJW:
I just realized I misread something somewhere. It seems like LLVM 18 is compatible with Rust 1.78 not 1.74, not sure where the mistake was on my side, but I'll revert that then to avoid any hidden consequences.

RayJW:
Seems like this might be less of a hassle than I initially thought. It seems like [LLVM 17 won't get a 24.08 branch](https://github.com/flathub/org.freedesktop.Sdk.Extension.llvm17/pull/7) due to being EOL, so it won't be possible to bump GNOME to 47 either since that doesn't have a 23.08 branch. I'm not sure if upping the minimum Rust version is viable.

flxzt:
Thank you!. Bumping Rust's MSCV should be fine in my opinion.

RayJW:
I used the [cargo-msrv](https://crates.io/crates/cargo-msrv) crate to check the compatibility and updated to the successful 1.79 version. Everything still compiles fine and seems to be working great so far :)

RayJW:
I bumped the MSRV to the latest Rust version (1.81) now since I don't see a reason to use a EOL version. I think that mostly makes sense for crates that are supposed to be used as dependencies.

PR #1257, opened 2024-10-18
Bump MSRV and LLVM to latest stable

RayJW:
As I already said in #1242 I think for binaries tracking the latest stable is best. Some discussion by distro maintainers around this can also be found [here](https://github.com/BurntSushi/ripgrep/issues/1019). Since Rust 1.82 just released, the latest LLVM 19 is now also stabilized, so I bumped them both up together :)


flxzt:
thank you! :)

PR #1268, opened 2024-10-29
feat: multi-click selections

### LINKED ISSUE #872: Double click to select word and sentences in the text?
If it is UX / UI related, don't post how a single aspect of the UI should be changed, rather in which use-case you felt the UI is lacking or should be improved.

**Is your feature request related to a problem?**
When selecting the texts typed in the note, it is not possible to use double click/triple click to select the entire word or all texts in a sentence, like other programs in Windows. 

Double clicking/triple clicking would select the word or sentence where the cursor is located. 

Holding down the left mouse button and move the mouse, but this is not as convenient.

**Additional context**  
Thanks for making this note-taking tool! It would be much more easy to type with quick selection implemented.

Kneemund:
Adds multi-click selection within typewriter text fields, mimicking the behavior of most text editors/browsers. 

- Fixes #872.
- Double click to select the current word. Drag after the second click to extend the selection onto other words.
- Triple click to select the current line. Drag after the third click to extend the selection onto other lines.
- Cycles through selection modes (caret, word, line) in case of more than 3 clicks.

### TODO
- [X] Performing a double click selection while hovering over whitespace will select the next word, but immediately expand the selection backwards onto the previous word (i.e. cover two words). This is nearly identical to the behavior of Firefox; the only difference being that Firefox only expands the selection after the mouse is moved in any way. Chromium and GNOME Text Editor simply select the whitespace because they generally snap the selection to any word boundary (start *and* end). I'd personally prefer the way Chromium/GNOME Text Editor handles this. EDIT: Switched to the Chromium variant.
- [X] The caret blink is always reset if the mouse moves, even if the selection doesn't change. This has been the case before but is more noticeable with multi-click selections, since the selections change less often. Not sure what the intended behavior is. EDIT: Cursor blink is now only reset if the position changes.

### Things to Discuss

- The gesture used for detecting multiple button presses is currently always enabled and has a generic name. It could either be tailored to this feature specifically and disabled when the typewriter tool is not selected, or an early return could be added (keeping it generic enough for potential future features that make use of the gesture).

### DIFF
--- a/crates/rnote-engine/src/engine/mod.rs
+++ b/crates/rnote-engine/src/engine/mod.rs
@@ -845,6 +845,32 @@ impl Engine {
         widget_flags
     }
 
+    pub fn text_select_closest_word(&mut self) {
+        if let Pen::Typewriter(typewriter) = self.penholder.current_pen_mut() {
+            typewriter.select_closest_word(&mut EngineViewMut {
+                tasks_tx: self.tasks_tx.clone(),
+                pens_config: &mut self.pens_config,
+                document: &mut self.document,
+                store: &mut self.store,
+                camera: &mut self.camera,
+                audioplayer: &mut self.audioplayer,
+            })
+        }
+    }
+
+    pub fn text_select_closest_line(&mut self) {
+        if let Pen::Typewriter(typewriter) = self.penholder.current_pen_mut() {
+            typewriter.select_closest_line(&mut EngineViewMut {
+                tasks_tx: self.tasks_tx.clone(),
+                pens_config: &mut self.pens_config,
+                document: &mut self.document,
+                store: &mut self.store,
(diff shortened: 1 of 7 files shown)


flxzt:
> The gesture used for detecting multiple button presses is currently always enabled and has a generic name. It could either be tailored to this feature specifically and disabled when the typewriter tool is not selected, or an early return could be added (keeping it generic enough for potential future features that make use of the gesture).


Because in practice the `text_select_*` methods return early when the typewriter is not active, I don't think it is a problem. In my opinion it can stay like this for now.

looks good to me, if you don't have anything to add it can be merged. :)

Kneemund:
Alright, I think we can merge it then. :)

flxzt:
thanks a lot!

PR #1269, opened 2024-10-31
feat: spellcheck for typewriter strokes

### LINKED ISSUE #271: Spellcheck
When typing input I sometimes misspell things. As I often use rnote for work, misspellings are not optimal. 

Add spell check to rnote text boxes.

The only one I can think of is to copy-paste text input into an external program with spell checking.

Kneemund:
Implements spellcheck for typewriter strokes using the [enchant](https://rrthomas.github.io/enchant/) library that wraps around many spellchecking libraries (Hunspell, Aspell, AppleSpell, ...). It is also used by `libspelling`, the official GTK4 spellchecking library.

The `enchant` crate for Rust is unfortunately not as good as libspelling's (official) bindings.

- Closes #271.
- Uses binary trees for caching results and only checks modified words.

### TODO

- [ ] Add per-document settings for enabling/disabling spellcheck and setting the language.
- [ ] Add interface for selecting and applying suggestions/corrections, e.g. a popover menu in the sidebar. Needs discussion.
- [ ] Allow locking spellcheck behind a password (https://github.com/flxzt/rnote/issues/271#issuecomment-1514241105). Might be out of scope, needs discussion.
- [ ] Evaluate new dependencies for Windows & MacOS. We might need to bundle additional libraries and dictionaries on Windows, that could be tricky. Switching to a different library (like [zspell](https://github.com/pluots/zspell)) should be simple. Needs discussion.
- [ ] Adapt Windows build workflow.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2123,6 +2123,35 @@ dependencies = [
 ]
 
 [[package]]
+name = "libspelling"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "1cbd36b794de5725e0b2be4cc90c57c5e3c7a5a3e5c317436e9e667305274c34"
+dependencies = [
+ "gio",
+ "glib",
+ "gtk4",
+ "libc",
+ "libspelling-sys",
+ "sourceview5",
+]
+
+[[package]]
+name = "libspelling-sys"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "2d2ec120461981daf9d0c5a8b0bc55ebf350292280e93fd6d063895593754484"
+dependencies = [
+ "gio-sys",
+ "glib-sys",
+ "gobject-sys",
+ "gtk4-sys",
+ "libc",
+ "sourceview5-sys",
+ "system-deps 7.0.3",
+]
+
+[[package]]
 name = "linux-raw-sys"
 version = "0.4.14"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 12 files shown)


Kneemund:
If the option to lock spellchecking is not out of scope, the UI could look something like this.

grafik

Questions about *how* to implement it, and maybe more importantly, how to avoid this being a footgun for users, remain. The digital dictionaries I've used in the past just played a fairly loud sound when unlocking the options again. That's not really an option here.

@LeSnake04 do you have any ideas on how to implement this? Admin passwords (which you suggested in your comment) seem like a big footgun to me - they're either ineffective, or can lock users out of this feature forever until they e.g. delete the app settings.

Kneemund:
This is what wavy error lines could look like. Not sure if I prefer these or the dashed lines.
### Wavy
grafik
### Dashed
grafik

Kneemund:
I think that there are mainly two options for displaying spellcheck corrections.

1. Popup with suggestions for word at cursor.
2. Panel with suggestions for all spellcheck errors in the current text stroke.

---

Proof of concept for the first option:

https://github.com/user-attachments/assets/a54f2f28-300f-4f4e-bf47-b2fe24ba628b

The "blinking button" is quite distracting IMO, maybe the sensitivity should always be enabled. Instead of making the button insensitive if there are no suggestions, we could add a hint telling users to move the cursor over a erroneous word.

Kneemund:
More fleshed out version without the button sensitivity changing; I personally like this a lot more.

https://github.com/user-attachments/assets/84d227fa-94cb-47a1-b9fb-68f53fc7adb7

Doublonmousse:
I wonder if the requirement of not having spellcheck could be done with a compile feature flag

Kneemund:
That would definitely solve it, but this requirement seems to be quite common, so I think we should try our best to make it accessible. The feature flag might still be a good idea for the CLI though.

Issue #956, opened 2024-01-01
Typewriter: Enter key doesn't create a new line


samualon:
**Describe the bug**  
When inside a Typewriter textbox, hitting the enter key does not create a new line. Instead nothing happens.

This does work on my Windows PC.

**To Reproduce**  
1. Select the Typewriter tool
2. Click somewhere to create a new box
3. Type some letters
4. Press enter

**Expected behavior**  
When hitting enter, the cursor should move to a new line.

**Console Output**  
No console output.

**Desktop (please complete the following information):**  
 - OS: Fedora 39
 - App Version: Rnote v0.9.4 (and Rnote v0.9.0)
 - Installation Source: Flatpak
 - Desktop Environment: Gnome 45.2
 - Display Server: X11
 - Input Source: Mouse, Keyboard and On-Screen-Tablet

Kneemund:
Hi, I think there's a good chance that this will get fixed by #1274. Are you able to build and test it (or test the nightly build after the PR is merged)?

Doublonmousse:
#1274 is included in the 0.12 version.
Can you test whether this works on this new version ?

PR #1274, opened 2024-11-03
improv: support keypad and uncommon keys

Kneemund:
Adds support for numerical keypad (numpad) keys and some more uncommon keys based on the [key mappings of Firefox](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_key_values).

- Addresses #956.


flxzt:
I am also optimistic that this will fix #956 . In any case, this is a nice improvement and should be included!

PR #1277, opened 2024-11-06
feat: laser tool

### LINKED ISSUE #476: Laser tool? (feature request)
A lot of handwriting notes app provide a laser too, so If you are presenting your screen, you can use that laser tool to draw over, but that draw is temporary and is timed, and disppears after some seconds. It would be good to have this feature.
A sample what I am talking about: https://youtu.be/G6lAWTgY0mo?t=527, this is actually a video about a notetaking app on IOS, but I just wanted to show how this feature works.

Kneemund:
Adds a laser pointer to the tool pen that produces stylized and temporary strokes with a constant width. The strokes fade out after lifting the pen (using animation frames), and fully disappear after a fixed amount of time (currently 1s).

- Closes #476 / #1062.
- Depends on #1309.

### Things to Discuss
- Changing stylistic choices (colors, widths, durations, ...).
  - Using a non-linear fade animation instead of the current linear one.
  - Waiting a few milliseconds before starting the fade animation.
- The tool currently uses an icon from icon library. It's supposed to depict CDs and lasers, but I'm not a huge fan. Can't find/think of a better icon though.


flxzt:
Regarding stylistic choices, the fade duration and whether to add a small time *before* fading: 

No objections to the current implementation, in my opinion the laser tool looks and behaves as expected.

> The tool currently uses an icon from icon library. It's supposed to depict CDs and lasers, but I'm not a huge fan. Can't find/think of a better icon though.

I don't love it, but I also don't have a better alternative. The only suggestion is to make it slightly smaller like this:

pen-tools-laser-symbolic

Kneemund:
I've changed the icon to the smaller one. Most other apps seem to use a literal laser pointer, but it's difficult to design one that is recognizable/detailed enough and still follows the HIG.

flxzt:
looks good, thank you!

maymage:
Over at [Papers](https://gitlab.gnome.org/GNOME/Incubator/papers/-/issues/295) we're also thinking about a laser tool - in case @Kneemund  you would like to cime in there.

Issue #1260, opened 2024-10-21
Allow import of password protected pdf


kq98:
Importing a password protected PDF currently fails. It would be great if the user could be prompted for a password instead.

If there is interest in including this feature, I offer to implement it myself and do a PR.

hamdan7m:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

astro-ray:
Is it possible to implement it? Cause that would solve a lot of my problem

kq98:
I'll look into it this weekend, but no promises.

Doublonmousse:
Closed with #1279

PR #1279, opened 2024-11-08
Allow import of encrypted PDF files

kq98:
This PR implements my suggestion from the Issue #1260 allowing for the import of encrypted PDF files by prompting the user for a password and passing it along to all appropriate poppler::Document calls.

I also provide the German translation of the new dialog used to prompt for the PDF password.

### DIFF
--- a/crates/rnote-ui/data/ui/dialogs/import.ui
+++ b/crates/rnote-ui/data/ui/dialogs/import.ui
@@ -280,8 +280,11 @@
       <response id="unlock" translatable="yes" appearance="suggested">_Unlock</response>
     </responses>
     <property name="extra-child">
-    <object class="GtkBox">
-      <property name="homogeneous">true</property>
+    <object class="GtkListBox" id="pdf_password_entry_box">
+      <property name="selection-mode">none</property>
+      <style>
+        <class name="boxed-list"/>
+      </style>
       <child>
         <object class="AdwPasswordEntryRow" id="pdf_password_entry">
           <property name="activates-default">True</property>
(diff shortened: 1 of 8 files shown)


crates/rnote-ui/data/ui/dialogs/import.ui
```diff
@@ -269,4 +269,26 @@
...
+    <property name="default-response">unlock</property>
+    <property name="close-response">cancel</property>
+    <property name="heading" translatable="yes">Encrypted PDF</property>
+    <property name="follows-content-size">False</property>
+    <responses>
+      <response id="cancel" translatable="yes">_Cancel</response>
+      <response id="unlock" translatable="yes" appearance="suggested">_Unlock</response>
+    </responses>
+    <property name="extra-child">
+    <object class="GtkBox">
```
flxzt, line 283: Change this to the following to fix the styling of the password entry. (see: https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.6/boxed-lists.html)

```suggestion
    <object class="GtkListBox">
      <property name="selection-mode">none</property>
      <style>
        <class name="boxed-list"/>
      </style>
```

kq98: Resolved in [e87107d](https://github.com/flxzt/rnote/pull/1279/commits/e87107d78fea840eb97b6e1aad11fce62c512b61) 
I opted for animating the ListBox, instead of the entry itself because doing so causes some really funky behaviour. It seems ListBox does not expect its children to be animated.

flxzt:
looks very good, thank you for your contribution!

PR #1281, opened 2024-11-09
fix: improve the clipboard pasting of files for better compatibility

Doublonmousse:
use `read_future(&["text/uri-list") ...` instead of `read_text_future` when pasting files from the clipboard.

This will make it work on the next gtk stable release (4.18) on mac os.
Testing on https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/7850 which fixes clipboard support on mac os for gtk.

As this shouldn't break things for the current gtk version we're currently using on the different os, I'm pushing this now.
- [x] Verify on current gtk, linux. Seems to improve things on kde, some paste operations on files list would fail before, I can't test on gnome though
- [x] Verify on current gtk, mac : Actually it makes it work on current gtk ! We'll have to wait on the next gtk release to have paste support of rnote stroke content between instances though
- [x] Verify on current gtk, windows. Doesn't work before, doesn't work after the change.

Overall, a net positive

### DIFF
--- a/crates/rnote-ui/src/appwindow/actions.rs
+++ b/crates/rnote-ui/src/appwindow/actions.rs
@@ -1160,38 +1160,73 @@ impl RnAppWindow {
                 async move {
                     debug!("Recognized clipboard content format: files list");
 
-                    match appwindow.clipboard().read_text_future().await {
-                        Ok(Some(text)) => {
-                            let file_paths = text
-                                .lines()
-                                .filter_map(|line| {
-                                    let file_path = if let Ok(path_uri) = url::Url::parse(line) {
-                                        path_uri.to_file_path().ok()?
-                                    } else {
-                                        PathBuf::from(&line)
-                                    };
-
-                                    if file_path.exists() {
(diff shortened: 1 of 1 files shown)


crates/rnote-ui/src/appwindow/actions.rs
```diff
@@ -1160,38 +1160,73 @@ impl RnAppWindow {
...
-                                })
-                                .collect::<Vec<PathBuf>>();
-
-                            for file_path in file_paths {
-                                appwindow
-                                    .open_file_w_dialogs(
-                                        gio::File::for_path(&file_path),
-                                        target_pos,
-                                        true,
+                    match appwindow
```
flxzt, line 1163: This is a candidate for DRY - almost the same code is used in three places in this source file

Doublonmousse: Fixed

flxzt:
I can't verify the functionality on macOS but the code looks fine to me! Thanks!

Issue #1228, opened 2024-09-21
Cannot build on windows


Doublonmousse:
**Describe the bug**  
Various errors when compiling on Windows, maybe from yet-to-be updated packages on mingw, for now I'm getting errors from `poppler-rs`

Tracking issue for this: https://gitlab.gnome.org/World/Rust/poppler-rs/-/issues/1

Cairo may also be affected.

Comes from the version bumps in #1218

Doublonmousse:
It's a `c_long` is not an i64 on windows issue.
Using a patched version of the crates.io package https://github.com/Doublonmousse/poppler-patch will work here (beware of using the gitlab version, doesn't play nice with dependencies as it uses git links to other repos instead)

hamdan7m:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PR #1282, opened 2024-11-09
Workaround for poppler_rs

Doublonmousse:
Whilst waiting on https://gitlab.gnome.org/World/Rust/poppler-rs/-/issues/1 to be fixed, fixes https://github.com/flxzt/rnote/issues/1228.

I've changed the github repo to highlight the changes with the current poppler_rs: https://github.com/Doublonmousse/poppler-patch/commit/3acaa4e28d45a05939a71116b0d5ebd4b1c99b98

I'll leave this as a draft for now, and will change to a dependency bump once poppler is updated with the fix.


flxzt:
I am uncertain when this will be fixed upstream. Do you think we should merge this dependency update to fix the windows build?

Doublonmousse:
I should add the revision to the cargo.toml file (and clean up the lock file) in that case before a potential merge.

Maybe it's better to merge now (makes a new release possible on windows, makes building easier for contributors) as we don't have a timeframe.

I was hopeful we'd see some activity tied with the usual annual linux releases (new distribution versions, gtk etc...) but I've not seen much for now.

Doublonmousse:
I locked the git revision (and the `Cargo.lock` diff ends up being only this change of source for `poppler` so all is good)

flxzt:
alright, looks good. Thanks!

PR #1283, opened 2024-11-09
feat: duplicate offset taking into account snap

### LINKED ISSUE #1245: "Duplicate Selection" moves objects not by a multiple snapping-distance
**Describe the bug**  
If you duplicate an object, its duplicate is offset in a way that you can't realign it horizontally or vertically with the original without turning of snapping.

**To Reproduce**  

1. Turn on snapping
2. Go in Selection mode
3. Select an object
4. Click on 'Duplicate Selection'
4. Try to align the duplicate with the original

**Expected behavior**  
Duplicating an object should offset the duplication a multiple of the snapping-distance

 - App Version: [v0.11.0]

Doublonmousse:
If snap is activated, the offset is one exactly the size of one grid element, and if not, uses the previous value.

We could do fancier things there. Always take into account the grid size (but switch based on whether the grid is visible or not). Or do an offset based on content (find a place that does not overlap with what's already there if possible)

### DIFF
--- a/crates/rnote-engine/src/store/selection_comp.rs
+++ b/crates/rnote-engine/src/store/selection_comp.rs
@@ -95,7 +95,11 @@ impl StrokeStore {
     /// Duplicate the selected keys.
     ///
     /// The returned, duplicated strokes then need to update their geometry and rendering.
-    pub(crate) fn duplicate_selection(&mut self, pattern_size:na::Vector2<f64>, snap_mode:bool) -> Vec<StrokeKey> {
+    pub(crate) fn duplicate_selection(
+        &mut self,
+        pattern_size: na::Vector2<f64>,
+        snap_mode: bool,
+    ) -> Vec<StrokeKey> {
         let old_selected = self.selection_keys_as_rendered();
         self.set_selected_keys(&old_selected, false);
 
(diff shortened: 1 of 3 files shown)


crates/rnote-engine/src/store/selection_comp.rs
```diff
@@ -126,8 +130,14 @@ impl StrokeStore {
             .collect::<Vec<StrokeKey>>();
 
         // Offsetting the new selected stroke to make the duplication apparent
-        self.translate_strokes(&new_selected, Stroke::IMPORT_OFFSET_DEFAULT);
-        self.translate_strokes_images(&new_selected, Stroke::IMPORT_OFFSET_DEFAULT);
+        // check if snap position is activated or not here
+        let offset = if snap_mode {
+            pattern_size
```
flxzt, line 135: It often is the case that the selection is not snapped to the grid even when `snap_positions` is active.
It would be better if the logic does not simply use the grid's size as offset but actively snap the duplicated selection
to the grid when it is active (maybe "default offset" + nearest grid edge?).

Doublonmousse:
Closing this as I'm not sure I'm going to finish this soon

PR #1285, opened 2024-11-09
make zooming consistent and better support for keypad keys

### LINKED ISSUE #1180: Implement Zoom Level Snap-to-100%
**Description**
Currently, the app allows users to zoom in and out freely either with zoom tool or using two fingers. The problem is going back to 100% zoom level. For, this I have to click on an icon in the top-left corner and then press a button showing the current zoom level. Though this might not sound like a big problem, it is problematic when I use the app for long time.

**Probable solution**
To improve user experience and consistency, I propose implementing a feature where the zoom level automatically goes to 100% when the user is within a certain proximity (e.g., 95%-105%) of the normal zoom level (while zooming in or out).

### LINKED ISSUE #1280: Inconsistency in zooming with Zoom in and Zoom out button or keyboard shortcut and no keyboard shortcut to reset zoom
**Description** 

Press the Zoom out and Zoom in button a few times or press `ctrl+-` and `ctrl++`. For example here is what happens when I press `ctrl+-` five times and then  `ctrl++` five times:

Zoom level: 90, 81, 73, 66, 59 | 65, 71, 79, 86, 95

Do it multiple times, and you will reach even more different zoom levels.

**Suggested behavior** 

Any of the two behaviors would be okay:

1. Zoom level: 90, 80, 70, 60, 50 | 60, 70, 80, 90, 100
2. Zoom level: 90, 81, 73, 66, 59 | 66, 73, 81, 90, 100

**Additional context**  

It would be better to set zoom in shortcut to `ctrl+=` as `ctrl++` means `ctrl+shift+=`.
Also, many apps use `ctrl+0` to reset zoom. But there is no keyboard shortcut to reset zoom in rnote.

Doublonmousse:
make zoom consistent

Take 1/zoom_in for zoom_out so that doing one zoom in then one zoom out ends up back to the same place

add more accelerators for zoom and keypad

- accelerator to reset zoom : ctrl + 0 (be it on the keyboard or numpad)
- added accelerators for keypad 1 to 6 for the selection of tools
- add `ctrl + =` for zoom in
- add `ctrl + +` and `ctrl + -` for + and -  the numpad keys

Fixes #1280 (and related to #1180)

### DIFF
--- a/crates/rnote-ui/src/appwindow/actions.rs
+++ b/crates/rnote-ui/src/appwindow/actions.rs
@@ -719,8 +719,8 @@ impl RnAppWindow {
                     return;
                 };
                 let viewport_center = canvas.engine_ref().camera.viewport_center();
-                let new_zoom =
-                    canvas.engine_ref().camera.total_zoom() * (1.0/(1.0 + RnCanvas::ZOOM_SCROLL_STEP));
+                let new_zoom = canvas.engine_ref().camera.total_zoom()
+                    * (1.0 / (1.0 + RnCanvas::ZOOM_SCROLL_STEP));
                 let mut widget_flags = canvas.engine_mut().zoom_w_timeout(new_zoom);
                 widget_flags |= canvas
                     .engine_mut()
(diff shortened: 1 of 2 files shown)


Kneemund:
Totally forgot about accelerators in #1274, nice catch.

flxzt:
both the improved zoom calculations and the added shortcuts are nice improvements, thanks!

PR #1286, opened 2024-11-09
change logic to keep selection on hover/pen up

### LINKED ISSUE #1222: Unable to select text using stylus
**Describe the bug**  
Using the typewriter, trying to select text with stylus immediately undoes the selection when you lift the stylus away from the screen. It works completely fine with mouse.

.

**Screenshots**  

https://github.com/user-attachments/assets/96fac41a-0fbc-4db9-a403-f7f471fae21e

 I think you already know those lol but here it goes:
 - OS: Fedora 40
 - App Version: 0.11.0
 - Desktop Environment: Gnome 46
 - Display Server: Wayland
 - Input Source: Onscreen tablet

Doublonmousse:
Fixes  #1222

This prevents the state from switching to up/hover state if the pen is lifted when a selection is active (aka the `ModifyState` is `Selecting`).

I'm not sure how useful the hover state is right now.

### DIFF
--- a/crates/rnote-engine/src/pens/typewriter/penevents.rs
+++ b/crates/rnote-engine/src/pens/typewriter/penevents.rs
@@ -454,15 +454,22 @@ impl Typewriter {
                 pen_down,
                 ..
             } => {
-                // detect hover state
-                *modify_state = if typewriter_bounds
-                    .map(|b| b.contains_local_point(&element.pos.into()))
-                    .unwrap_or(false)
-                {
-                    ModifyState::Hover(element.pos)
-                } else {
-                    ModifyState::Up
-                };
+                match modify_state {
+                    ModifyState::Selecting { .. } => {
+                        // do nothing in this case : this keeps the text selected when using a pen
+                    }
+                    _ => {
+                        // detect hover state
+                        *modify_state = if typewriter_bounds
+                            .map(|b| b.contains_local_point(&element.pos.into()))
+                            .unwrap_or(false)
+                        {
+                            ModifyState::Hover(element.pos)
+                        } else {
+                            ModifyState::Up
+                        };
+                    }
+                }
                 *pen_down = false;
 
                 EventResult {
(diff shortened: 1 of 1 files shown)


crates/rnote-engine/src/pens/typewriter/penevents.rs
```diff
@@ -454,15 +454,22 @@ impl Typewriter {
...
-                // detect hover state
-                *modify_state = if typewriter_bounds
-                    .map(|b| b.contains_local_point(&element.pos.into()))
-                    .unwrap_or(false)
-                {
-                    ModifyState::Hover(element.pos)
-                } else {
-                    ModifyState::Up
-                };
+                match modify_state {
```
flxzt, line 457: Wouldn't it be nicer to write something like `if !matches!(ModifyState::Selecting {..})` ?

flxzt:
> I'm not sure how useful the hover state is right now.

It is used to change the appearance of the resize/rotate nodes when hovering over them.
I haven't tested it yet but wouldn't this change cause the nodes to appear getting "stuck" in the down state?

Doublonmousse:
> > I'm not sure how useful the hover state is right now.
> 
> It is used to change the appearance of the resize/rotate nodes when hovering over them. I haven't tested it yet but wouldn't this change cause the nodes to appear getting "stuck" in the down state?

Oh right, now I get it. And yes, it will force the state to `Down` as long as text is selected but that's also what happens currently when text is selected with the mouse

https://github.com/user-attachments/assets/0204cd08-114f-4cfe-9ff1-f60c88966f31

And with the the enum definition, there's no way around it https://github.com/Doublonmousse/rnote/blob/891b8e4cb68369424d2cae6665bca5abff79b974/crates/rnote-engine/src/pens/typewriter/mod.rs#L28C17-L28C28

It's either we have the text selected (`ModifyState::Selecting`) or we are hovering (`ModifyState::Hover`). We cannot have text selected and see the hover effect on the nodes.

crates/rnote-engine/src/pens/typewriter/penevents.rs
Doublonmousse: Fixed

flxzt:
You're right, it' s a code architecture flaw. `Hover` really shouldn't be a enum variant of ModifyState..

PR #1290, opened 2024-11-12
mapping between gtk timestamp and instant

### LINKED ISSUE #1289: Incorrect timings on event from using `Instant::now`
In the input file, there is this line
https://github.com/flxzt/rnote/blob/7f3ff8aafcaf3137bec9993329667e5e9fbf4c4e/crates/rnote-ui/src/canvas/input.rs#L21
That uses the instant at which the controller is called for the pointer controller to calculate the time of the event (or more precisely, gives the timestamp for the latest event of that `gdk::Event`).

But this creates erroneous times. Debugging what's given by #1235, we get the following
```
┌─────────────┬──────────────────────────┬────────────────┬────────────┬────────────┬────────────────┐
│ speed       ┆ positions                ┆ delta position ┆ delta_time ┆ status_pen ┆ one_time_speed │
│ ---         ┆ ---                      ┆ ---            ┆ ---        ┆ ---        ┆ ---            │
│ f64         ┆ array[f64, 2]            ┆ f64            ┆ f64        ┆ str        ┆ f64            │
╞═════════════╪══════════════════════════╪════════════════╪════════════╪════════════╪════════════════╡
│ 972.168716  ┆ [227.732923, 101.225992] ┆ 6.571739       ┆ 0.009849   ┆ Proximity  ┆ 667.249416     │
│ 883.789741  ┆ [227.732923, 101.225992] ┆ 0.0            ┆ 0.0014774  ┆ Down       ┆ 0.0            │ 
│ 10000.0     ┆ [223.591976, 101.919672] ┆ 4.198647       ┆ 0.0        ┆ Down       ┆ null           | 
│ 9143.296429 ┆ [221.321795, 102.31905]  ┆ 2.305043       ┆ 0.004      ┆ Down       ┆ 576.260717     │
│ 8376.738644 ┆ [218.526114, 102.844562] ┆ 2.844643       ┆ 0.004      ┆ Down       ┆ 711.160797     │
│ 10000.0     ┆ [214.805568, 103.664364] ┆ 3.809796       ┆ 0.0001306  ┆ Down       ┆ 29171.483352   │
│ 9175.957186 ┆ [211.169086, 104.547222] ┆ 3.742116       ┆ 0.004      ┆ Down       ┆ 935.529043     │
│ 8406.43496  ┆ [208.415475, 105.261924] ┆ 2.844851       ┆ 0.004      ┆ Down       ┆ 711.212699     │
│ 7686.83764  ┆ [206.523649, 105.787437] ┆ 1.963458       ┆ 0.004      ┆ Down       ┆ 490.864439     │
│ 10000.0     ┆ [204.631849, 106.291927] ┆ 1.957912       ┆ 0.0        ┆ Down       ┆ null           │
└─────────────┴──────────────────────────┴────────────────┴────────────┴────────────┴────────────────┘
```
What we're concerned about is the delta position and delta time between two events (and the ratio of the two, in the last column). This is with the `BacklogPolicy::NoLimit` option. And we see we get events with a non zero position difference and a zero time delta (not just close to zero, EXACTLY 0).

Looking at the raw event data, the first three events are here :
```
RawEvent(
    x=[227.7329226372115], 
    y=[101.22599202560606], 
    event_type='other', 
    time_str=['Instant { t: 1383023.4689412s }'], 
    gtk_time=[1382966381]
    )
RawEvent(
    x=[227.7329226372115], 
    y=[101.22599202560606], 
    event_type='ButtonPress',  // the button Press Event
    time_str=['Instant { t: 1383023.4704186s }'], 
    gtk_time=[1382966381] // for gtk, this is sent at exactly the same position as the previous location with the SAME time
    )
RawEvent(
    x=[223.59197575860676, 221.3217951414422, 218.5261142707021], 
    y=[101.91967188204403, 102.31904963454664, 102.84456208919609], 
    event_type='other', 
    time_str=['Instant { t: 1383023.4703054s }', 'Instant { t: 1383023.4743054s }', 'Instant { t: 1383023.4783054s }'], 
    // for rnote, the first event is at 1383023.4703054s which is EARLIER than the previous event
    // at  1383023.4704186s
    gtk_time=[1382966388, 1382966392, 1382966396] // the first event is 7 ms after the last one
    )
```
So we get later events placed before previous events, when there is a non-empty history. This is why we get exactly 0.0 as the difference of instants (`(time - last_element_time).as_secs_f64()` saturates to 0 if the difference is negative).

So we should add somewhere (in the canvas/ui code ?) a mapping between `Instant` and gtk's u32 time stamping (maybe set this on the first event with a time stamp. so we have a `Instant::now` corresponding to a `u32` value and do other calculations relative to this to still keep `std::Instant` lower in the engine code).
I don't know whether there is other precaution to take regarding gtk4's time (does it overflow ? I mean it'll take a while for that to happen, 50 days from my napkin math, but if it increases with the system clock and not screen time, it's possible to have this issue on tablets that are never shut down with rnote staying open)

Doublonmousse:
Fixes #1289

I'm not sure if I need to take into account the overflow, this complicates things quite a bit.
It's a little bit ugly, but it's probably enough to test things out.

I'm getting a lot of 0 for event times when scrolling, I need to see what event triggers this so there isn't any residual weirdness

### DIFF
--- a/crates/rnote-ui/src/canvas/input.rs
+++ b/crates/rnote-ui/src/canvas/input.rs
@@ -18,7 +18,7 @@ pub(crate) fn handle_pointer_controller_event(
     event: &gdk::Event,
     mut pen_state: PenState,
 ) -> (glib::Propagation, PenState) {
-    let now = Instant::now();
+    let now = canvas.get_time(event.time());
     let mut widget_flags = WidgetFlags::default();
     let touch_drawing = canvas.touch_drawing();
     let gdk_event_type = event.event_type();
@@ -370,9 +370,7 @@ fn retrieve_pointer_elements(
             }
 
             let entry_delta = Duration::from_millis(event_time.saturating_sub(entry.time()) as u64);
-            let Some(entry_time) = now.checked_sub(entry_delta) else {
-                continue;
-            };
+            let entry_time = canvas.get_time(event_time);
 
             if let BacklogPolicy::Limit(delta_limit) = backlog_policy {
                 // We go back in time, so `entry_delta` will increase
--- a/crates/rnote-ui/src/canvas/mod.rs
+++ b/crates/rnote-ui/src/canvas/mod.rs
@@ -28,9 +28,9 @@ use rnote_engine::ext::GraphenePointExt;
 use rnote_engine::ext::GrapheneRectExt;
 use rnote_engine::Camera;
 use rnote_engine::{Engine, WidgetFlags};
-use std::cell::{Cell, Ref, RefCell, RefMut};
+use std::cell::{Cell, OnceCell, Ref, RefCell, RefMut};
 use std::path::Path;
-use std::time::Duration;
+use std::time::{Duration, Instant};
 use tracing::{debug, error, warn};
 
 #[derive(Debug, Default)]
@@ -53,6 +53,8 @@ struct Connections {
 }
 
 mod imp {
+    use std::time::Instant;
+
     use super::*;
 
     #[derive(Debug)]
@@ -87,6 +89,11 @@ mod imp {
         pub(crate) show_drawing_cursor: Cell<bool>,
 
         pub(crate) last_export_dir: RefCell<Option<gio::File>>,
+
+        // mapping from gtk timestamp to instant
+        pub mapping_time: OnceCell<Instant>,
+        pub last_gtk_time: RefCell<u32>,
+        pub overflow_duration: RefCell<Duration>,
     }
 
     impl Default for RnCanvas {
@@ -181,6 +188,10 @@ mod imp {
                 show_drawing_cursor: Cell::new(false),
 
                 last_export_dir: RefCell::new(None),
+
+                mapping_time: OnceCell::new(),
+                last_gtk_time: RefCell::new(0),
+                overflow_duration: RefCell::new(Duration::default()),
             }
         }
     }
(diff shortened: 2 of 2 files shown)


Doublonmousse:
More weirdness :
- two events can be sent with the same gtk time (hence why the backlog policy is useful, even if it's to put it at 1 ms)
- scroll events don't have timestamps (I think that would only affect things when scrolling with the left click pressed)

Doublonmousse:
After thinking a bit, I'm not sure if I'm not making a bit deal of a (relatively) small issue. We have a timestamp for all events, albeit sometimes less precise than gtk when it gives it, with larger deviations in the case of the `ButtonPress` for pen on windows. But for all time sensitive algorithm, as long as we do the following
- reject negative time deltas
- robustness to time deltas that are too small (compared to their real duration)
Things will be fine.

Maybe I can check on my hardware by how much doing the `Instant::now` vs relying on gtk time changes things

flxzt:
what should we conclude? Maybe it's not worth the effort to change on Gtk timestamps, when it already works reasonably well. And at least Instant gives a monotonically increasing time. I have the feeling when we rely on Gtk here, we end up having to work around certain corner cases.
With regards to this fixing #1289, it was raised because of the increased size eraser pull request, but I am questioning if we actually add this in the first place.

Doublonmousse:
It only matters if we calculate speed (for the eraser size pr or maybe for the ink stroke modeler -I need to recheck, I don't remember if the stroke modeler takes some `dt` or works on fixed size).

Also, _technically_ the instant will give us the correct time only if there isn't lag (between the actual last event and when we get to `Instant::now`). 
The gtk timestamp is also a little cumbersome to work with because of the rollover nature (1 to MAX_INT and back to 1 to ...)

I think we can close this, and work around edge cases (documented in #1289) when they occur.

PR #1294, opened 2024-11-16
fix: selection resize/rotate handle sensitivity in 'single' selection mode

### LINKED ISSUE #1292: Selecting objects with "Select One by One" doesn't allow to resize/rotate them
**Describe the bug**  
Selecting objects with "Select One by One" doesn't allow to resize them, **if** there's another object behind the resize square or the rotate circle.

This happens mainly when selecting images, since they usually occupy a big area.

**To Reproduce**  
_Here I go to the more error prone case_
1. Go to 'Selector'
2. Click on 'Select One by One'
3. Select a stroke over an image
4. Try to resize/rotate the selected stroke (you will select the image instead)

**Expected behavior**  
Resize/rotate the object instead of selecting whatever is behind it

**Console Output**  
Not necessary.

**Screenshots**  
https://github.com/user-attachments/assets/21a14624-fe6d-490f-a92f-866908ac4209

 - OS: Fedora 41
 - App Version: 0.11.0
 - Desktop Environment: Gnome 47
 - Display Server: Wayland
 - Input Source:Mouse, External Graphics Tablet

**Additional context**

Doublonmousse:
This allows one element to be resized/rotated in this mode even if a stroke/image is behind.

The old behavior can be forced in this mode when pressing `Shift`

Fixes #1292 

As we are on the subject, if you click inside the selection in this mode and a stroke is there under, it will be selected as well. This means resizing/rotating without adding strokes is facilitated by this PR but not translation. There's not really a translation node (in addition to the rotate and scaling ones)

### DIFF
--- a/crates/rnote-engine/src/pens/selector/penevents.rs
+++ b/crates/rnote-engine/src/pens/selector/penevents.rs
@@ -88,8 +88,7 @@ impl Selector {
                             )
                             .pop();
 
-                        if (engine_view.pens_config.selector_config.style == SelectorStyle::Single
-                            || modifier_keys.contains(&ModifierKey::KeyboardShift))
+                        if (modifier_keys.contains(&ModifierKey::KeyboardShift))
                             && key_to_add
                                 .and_then(|key| engine_view.store.selected(key).map(|s| !s))
                                 .unwrap_or(false)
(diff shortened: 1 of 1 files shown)


flxzt:
looks good to me, thanks!

PR #1308, opened 2024-12-01
refactor: tools event handling

flxzt:
this refactors that the individual tools handle pen events completely independent from each other.
It shouldn't change anything in their behaviour, if that is the case it's a bug.


Kneemund:
Looks good, maybe we should also split up the individual tools into their own modules/files?

flxzt:
good idea, done.

Kneemund:
I've fixed two small issues in b9b180820e2e5fce5ab69fb8564a57ede7cdad81

PR #1309, opened 2024-12-03
feat: universal animation frames

Kneemund:
Adds the ability to (continuously) request animation frames to render animations that are not strictly tied to events.

- Needed for #476 and its PR.
- Supersedes #1276.
- An additional struct `Animation` is added to the engine and used to request animation frames.
- Uses [`Widget#add_tick_callback`](https://docs.gtk.org/gtk4/method.Widget.add_tick_callback.html) to hook into the render loop and call `handle_animation_frame` & redraw the canvas if frames are requested.

### DIFF
--- a/crates/rnote-engine/src/pens/penholder.rs
+++ b/crates/rnote-engine/src/pens/penholder.rs
@@ -253,10 +253,19 @@ impl PenHolder {
             widget_flags |= wf;
         }
 
-        // Always redraw after handling a pen event
+        // Always redraw after handling a pen event.
+        //
+        // This is also needed because pens might have claimed/requested an animation frame.
         widget_flags.redraw = true;
 
         (event_result.propagate, widget_flags)
+    }
+
+    /// Handle a requested animation frame.
+    ///
+    /// Returns true if another frame is requested.
+    pub fn handle_animation_frame(&mut self, engine_view: &mut EngineViewMut) -> bool {
+        self.current_pen.handle_animation_frame(engine_view)
     }
 
     /// Handle a pressed shortcut key.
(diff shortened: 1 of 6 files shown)


flxzt:
This looks good to me, it's a significant improvement to the other approach. Feel free to merge when you think it's ready!

flxzt:
One improvement that comes to mind - since we already have the "improve E-PD" option, it would be nice if animations/the tick callback is disabled with it.

Kneemund:
Thanks for your help.

> One improvement that comes to mind - since we already have the "improve E-PD" option, it would be nice if animations/the tick callback is disabled with it and animation implementations are able to somehow deal with it accordingly

I guess we could either disable the entire tick callback and make sure that animations are always optional, or continue to tick the animation updates (`handle_animation_frame`) and just not call `queue_draw` at the end. The latter would still show certain frames of the animation whenever the canvas is redrawn, so it'd probably look weird (besides being a hacky solution).

crates/rnote-engine/src/pens/penholder.rs
```diff
@@ -253,12 +253,22 @@ impl PenHolder {
...
+        //
+        // This is also needed because pens might have claimed/requested an animation frame.
         widget_flags.redraw = true;
 
         (event_result.propagate, widget_flags)
     }
 
+    /// Handle a requested animation frame.
+    ///
+    /// Can request another frame using `ÈngineViewMut#animation.claim_frame()`.
```
flxzt, line 266: Pedantic but the `E` shouldn't have a grave accent

Kneemund: fixed, I somehow didn't see that

PR #1314, opened 2024-12-05
Make `insert_stroke` public

Doublonmousse:
This makes it easier to use rnote as a rust library if the intent is to add strokes programmatically to a rnote file.

### DIFF
--- a/crates/rnote-engine/src/store/mod.rs
+++ b/crates/rnote-engine/src/store/mod.rs
@@ -314,7 +314,7 @@ impl StrokeStore {
     /// Optionally a desired layer can be specified, or the default stroke layer is used.
     ///
     /// The stroke then needs to update its rendering.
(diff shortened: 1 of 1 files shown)


Doublonmousse:
Maybe this isn't necessary (and calling `import_generated_content` would be preferred ? This other method is public)

flxzt:
yeah, would `import_generated_content` satisfy what you need?

Doublonmousse:
Widget flags are only for the UI, is that correct ? 
I don't need to call `canvas.emit_handle_widget_flags(widget_flags)` after `import_generated_content`, right ? 

If that's the case I can use `import_generated_content` without issue

flxzt:
Yeah only when something changed in the engine and has to be updated/reflected in the UI.

Doublonmousse:
Yes that works. Closing this

PR #1318, opened 2024-12-07
refactor: hover states in selector and typewriter

flxzt:
refactors the hover states to be independent of the modify states. see discussion in #1286 

prerequisite for progress on #1286 

@Doublonmousse


Doublonmousse:
looks good to me !

Do you want me to use #1286 to implement the original change requested + expand this PR to make the nodes on text appear as selected when hovering over it even when text is selected ?

flxzt:
Thanks for taking a look!

This would be a prerequisite for continuing with the original change in #1286 since it resolves the issue with the hover state you were having while implementing it

Issue #744, opened 2023-07-11
OneNote import


flxzt:
**Is your feature request related to a problem? Please describe.**  

The only import option that keeps notes in a "rich" and editable way is the xournal++ file import. Because many potential users come from using OneNote, we should consider adding a direct import & conversion from their `.one` files. It's a bit hidden in OneNote, but users are able to save them from onenote or download them from OneDrive when saved in the cloud.

**Describe the solution you'd like**  

Another import option, similar to the Xournal++ import.

**Additional context**  

also discussed in #733 .

I found some rust crates for parsing/processing onenote files: [onenote.rs](https://github.com/msiemens/onenote.rs) is a parser, but currently only supports the format that is saved in OneNote's cloud. [one2html](https://github.com/msiemens/one2html) and [one_note_to_xopp](https://github.com/jonas-kell/one_note_to_xopp) are crates that utilize this parser and save to HTML / .xopp format.

einsJannis:
is this already beeing worked on?

flxzt:
I have a WIP branch locally, but it is stuck because of a [bug](https://github.com/msiemens/onenote.rs/issues/11) in the parsing library that I wanted to use.

justinrubek:
@flxzt do you still have your branch somewhere? The bug mentioned looks like it could be a serious issue, but also never mentions stroke data directly which is what I am interested in because I have a number of OneNote notebooks that I primarily want the handwritten notes and drawings from. I'd be willing to make an attempt to look into this given that importing at least those things may be possible and I may be able to work on that bug as part of it.

Doublonmousse:
One thing to note : All microsoft-based applications (onenote, microsoft journal but also office) copy the InkML information of strokes when selecting ink strokes and copying them to the clipboard as well as Ink Serialized Format.
![image](https://github.com/flxzt/rnote/assets/115779707/3643380e-1c7f-4653-97ab-9a5387839093)

When image are selected, base64 encoding of image seems to be available under the HTML format.

InkML is an open W3C format that's readable (contrary to ISF which is a serialized format, although converter exist and the standard is open as well).
Unfortunately, this info is only available when selecting strokes and copying it to the clipboard, and not in any export menu.

Anyway, I think supporting InkML as an import format (when pasting from the clipboard) could be done and with a little more work 1-1 import of *selected* parts of onenote copied to the clipboard could be achieved in rnote.

[...]

PR #1320, opened 2024-12-08
feat: add inkml parser and writer

Doublonmousse:
Related to #744 (import of selected onenote strokes, but exclusive to that. No text nor images are read).

As I'm on windows, I'm using #1282 for now.
The parser/writer crate is on a separate repository here https://github.com/Doublonmousse/writer_reader_inkml

Export and import are supported (from the clipboard).
- I think it would make sense to move some of the import code inside the engine (conversion from inkml to stroke ?)
- The export part is only for brush strokes (I could extend it to shapes with lines being the easiest to do, and other shapes would have to be discretized)
- I'm using the windows equivalent of the mime type for the inkml clipboard export part
- The crate still needs some polishing (better error message and removing some potentially panic inducing unwrap with error propagation instead).
 
I've encountered a weird behavior with clipboards in gtk and windows (that ofc I can't reproduce today ...). Copying ink content then an image from a source application wouldn't work as somehow gtk would think the inkml format was still present before failing with a gio quark error.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4238,6 +4238,7 @@ dependencies = [
  "tracing",
  "unicode-segmentation",
  "usvg",
+ "writer_inkml",
  "xmlwriter",
 ]
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5929,7 +5929,7 @@ checksum = "1ffae5123b2d3fc086436f8834ae3ab053a283cfac8fe0a0b8eaae044768a4c4"
 [[package]]
 name = "writer_inkml"
 version = "0.1.0"
-source = "git+https://github.com/Doublonmousse/writer_reader_inkml?rev=6f5f903199a16996d5c2cb2a074cfd98cebd3f67#6f5f903199a16996d5c2cb2a074cfd98cebd3f67"
+source = "git+https://github.com/Doublonmousse/writer_reader_inkml?rev=9370cf6a65ce566d91c232ca47d2cf08008e605e#9370cf6a65ce566d91c232ca47d2cf08008e605e"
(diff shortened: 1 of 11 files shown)


Doublonmousse:
The issue I had was fixed upstream, so this should be working for gtk 4.20

Doublonmousse:
One passing though on conversions: I _feel_ like an interface/trait for general conversion could be done. 

Also, if it'd be possible to have a minimal rnote library that's just big enough to load rnote files but also create elements and push to the store then save to a rnote file that'd facilitate how to do conversions (and also embed the reader in other contexts)

Doublonmousse:
Implemented shapes export ("rasterized" to lines) using the `outline_path` method + a couple of fixes
I need to retest on windows to verify everything is fine with import/export from/to onenote/journal/office.

I'm also encountering some weird behavior for rendering. When doing a roundtrip (copy from rnote, then modify the clipboard to only keep the inkml content then pasting again), the arrow should not display correctly (as I'm currently doing it in one stroke vs taking into account the `MoveTo` as an indication to do a new stroke) but it does
image
If I remove the start,
image

A pasted polygon (so represented as a brushstroke here !) can "lose" a side when seen from further away
image
image

Doublonmousse:
Re tested on windows with journal/onenote/word and everything works as expected (including shapes).

Export to journal doesn't take the inkml format but prefers the png (edit: retested and it outputs inkml but the reverse is false, only ISF would work on import).

Now there may also be some annoyances here (if I prefer inkml over other clipboard format, I may miss on images ..). Though apart from a `copy as/paste as` function I'm not sure it can be fixed

PR #1321, opened 2024-12-08
refactor: reinstall pen only if necessary

Kneemund:
This PR refactors the pen progress handling to only reinstall the current pen if necessary, i.e. if there was an override. It also fixes a few bugs along the way:

1. Fixes temporarily switching to the laser tool and back.
2. The D key no longer deselects all strokes during selection.
3. (Proper event results when the A key is pressed during selection. Does not seem to have an effect.)

The state of the laser tool and its strokes is now retained after `PenProgress::Finished`, making the first bug fix possible (without breaking something else).

Previously, all events that ended with `PenProgress::Finished` always caused a cancel pen event, a full state reset of the pen and a subsequent call to `update_state()`. Some pens relied on this, which this PR attempts to fix/change **without** changing their previous behavior.

If anyone decides to test this and finds some unintentional changes to the previous behavior, please let me know.

### DIFF
--- a/crates/rnote-compose/src/builders/penpathcurvedbuilder.rs
+++ b/crates/rnote-compose/src/builders/penpathcurvedbuilder.rs
@@ -71,7 +71,10 @@ impl Buildable for PenPathCurvedBuilder {
             (_, PenEvent::Up { element, .. }) => {
                 self.buffer.push(element);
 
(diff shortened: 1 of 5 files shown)


flxzt:
looks good to me, I briefly tested it and could verify all fixes for the issues you described.

PR #1331, opened 2024-12-21
refactor: avoid using format! when String creation is unnecessary

Integral-Tech:
- When String creation is unnecessary, replace `format!()` with `format_args!()` to avoid overhead of heap allocation.


flxzt:
looks good, thanks!

PR #1360, opened 2025-02-17
fix: select strokes after duplicate

madonuko:
Fix selection disappearing after `ctrl+D`. This is *really really* annoying in my usecase.

### DIFF
--- a/crates/rnote-engine/src/store/selection_comp.rs
+++ b/crates/rnote-engine/src/store/selection_comp.rs
@@ -104,7 +104,6 @@ impl StrokeStore {
             .filter_map(|&old_key| {
                 let new_key =
                     self.insert_stroke((**self.stroke_components.get(old_key)?).clone(), None);
-                self.set_selected(new_key, true);
 
                 // duplicate and insert the render images of the old stroke to avoid flickering
                 if let Some(render_comp) = self.render_components.get(old_key) {
@@ -121,6 +120,9 @@ impl StrokeStore {
                         );
                     }
                 }
+
+                self.set_selected(new_key, true);
+
                 Some(new_key)
             })
             .collect::<Vec<StrokeKey>>();


Doublonmousse:
So after the duplication no stroke is selected (neither the new or the old selection) ?

madonuko:
After duplication, no stroke *was* selected, but this PR fixes the behaviour so that the new stroke is selected.

flxzt:
I can't reproduce the issue - the duplicated strokes are selected for me.
Can you give details how your change fixes it in your PR?

madonuko:
https://github.com/user-attachments/assets/dc7bdf18-db21-4362-818d-baa597b771bf


https://github.com/user-attachments/assets/e9580e90-95a2-4215-bfad-a9f22eb281fb

Doublonmousse:
Is there any way to debug why the `get_mut` call on Arc fails ? Maybe we can get some info with `rust-gdb` by adding prints/breakpoint inside the `set_selected` function when it fails ?

madonuko:
I believe it has to do with race conditions because it works some of the time (but 90% of the time it doesn't)

Doublonmousse:
I think I'm getting a rough idea of what could be happening. Though if that's the case, it probably can happen in other places and situation and result in very unwanted results.

Inside of the `set_selected` function, can you do something like
```rs
    pub(crate) fn set_selected(&mut self, key: StrokeKey, selected: bool) {
        let key_out = Arc::make_mut(&mut self.selection_components).get_mut(key);
        if let Some(arc_selection_comp) = key_out {
            // we check the weak/strong reference count before
            let weak_count = Arc::weak_count(arc_selection_comp);
            let strong_count = Arc::strong_count(arc_selection_comp);

            // then call `get_mut` and NOT `make_mut`
            if let Some(selection_comp) = Arc::get_mut(arc_selection_comp) {
                selection_comp.selected = selected;
                self.update_chrono_to_last(key);
            } else {
                println!(
                    "failed to get a mut reference to the stroke, would have fallbacked to a copy with `make_mut`. Weak count was {:?}, strong count was {:?}",
                    weak_count, strong_count
                );
            }
        }
    }
```
and keep both calls (on the removed line where it didn't work for you and on the new line where it worked) for 
```rs
self.set_selected(new_key, true);
```
Then run again and log the output.

I want to see the difference between both calls. Hopefully I don't trigger the observer effect and make the bug disappear trying to observe it in the wild.

I think we may be in the example case here https://doc.rust-lang.org/stable/alloc/sync/struct.Arc.html#method.make_mut 
with `data`/`other_data` pointing to different objects, but as the `stroke` here is also in `self.selection_components`, only one of the two is actually saved in the engine (and we'd be writing to a copy that's now living outside the engine)

flxzt:
Just chiming in - using `make_mut` is intentional, the way this works is that `make_mut()` call only clones (in-place!) when the current state is also in the history - this is why `get_mut()` fails.

Doublonmousse:
Right, so with the reference from the `self.selection_components` only `make_mut` can work.

I'm wondering if there can't be another additional reference to the same stroke at the time we call this function (in addition to the `self.selection_components`).

I'd like to really understand the bug here
- Do we not have the key yet in the `selection_components` ? is this possible ?
- Do we have weird issues with the `Arc` reference ? Should we do a `check_selected` debug function that immediately `make_mut` again the  `selection_comp` to check that it is actually properly selected after the call ? (in that case we'll print the `selection_comp` from the engine here so there's no question of wrong reference possible here)
- Is this a bug from somewhere else entirely (`rustc` ??)

Given what the fix is, I feel like the issue is in that part of the code (and not the handling of the selection change by the selector afterward).

It _feels_ like we correctly record the event but it doesn't register.

madonuko:
I can't reproduce anymore after upgrading to Fedora 42…

flxzt:
so this can be closed? I am also skeptical this change is a proper fix, I can't see how invoking `set_selected()` at a different point would change anything.

I think the another angle would have been to look at the input events, maybe the palm rejection didn't work previously resulting in a deselect or something.

madonuko:
I'd rather let somebody else that can reproduce this figure this out, let's close this for now.

PR #1361, opened 2025-02-17
feat: shortcuts for color setters

madonuko:
Add `<Alt>1`, `<Alt>2`, ..., `<Alt>9` as the shortcuts for the color pickers (at the top of
the window).


mahdi4261:
I really wanted that feature for a long time. Thanks for implementing that. But, this shortcut is already used for switching to specific tabs. For example <kbd>Alt</kbd> + <kbd>3</kbd> to go to 3rd tab.

madonuko:
Didn't know that, I feel like it's more frequent to switch colors than to switch tabs, so maybe it's not a bad idea to remap tab switching to Ctrl+Alt+n?

mahdi4261:
I agree with you that I rarely (probably never) switch tabs with that keyboard shortcut. I just reminded you about the conflict.

Doublonmousse:
Pretty sure the shortcut is not set by rnote but by the use of [Adw](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.6/index.html)>[TabBar](https://gnome.pages.gitlab.gnome.org/libadwaita/doc/1.6/class.TabBar.html) in rnote for having tabs, and I'm not sure it's settable.

I'd be somewhat weird to have rnote be the only gtk/adwaita application where the `Alt+<Number>` keyboard shortcut for switching tabs didn't work so I'm not really in favor of overriding this keyboard shortcut.

I didn't know the shortcut and I'm not sure it not showing in the keyboard shortcut window is intended behavior (though that's a separate issue altogether)

flxzt:
what about using `ctrl+shift+<n>` for either switching colors?

crates/rnote-ui/src/appwindow/actions.rs
```diff
@@ -1171,6 +1186,12 @@ impl RnAppWindow {
         app.set_accels_for_action("win.pen-style::eraser", &["<Ctrl>4", "<Ctrl>KP_4"]);
         app.set_accels_for_action("win.pen-style::selector", &["<Ctrl>5", "<Ctrl>KP_5"]);
         app.set_accels_for_action("win.pen-style::tools", &["<Ctrl>6", "<Ctrl>KP_6"]);
+        (1..=9).for_each(|i| {
+            app.set_accels_for_action(
+                &format!("win.set-color-{i}"),
+                &[&format!("<Ctrl><Shift>{i}"), &format!("<Alt>KP_{i}")],
```
Kneemund, line 1192: The keypad shortcuts still use <kbd>Alt</kbd> atm.
```suggestion
                &[&format!("<Ctrl><Shift>{i}"), &format!("<Ctrl><Shift>KP_{i}")],
```

crates/rnote-ui/src/appwindow/actions.rs
```diff
@@ -164,6 +164,21 @@ impl RnAppWindow {
...
         self.add_action(&action_active_tab_close);
 
+        let color_setters = {
+            let p = self.overlays().colorpicker();
+            [
+                [p.setter_1(), p.setter_2(), p.setter_3()],
+                [p.setter_4(), p.setter_5(), p.setter_6()],
+                [p.setter_7(), p.setter_8(), p.setter_9()],
+            ]
+            .concat()
```
Kneemund, line 174: Not sure if we really need to split the array for aesthetic purposes, but it does look better ;)

madonuko: I mean I could revert it but… it looks nice so :3

Atreyu-94:
I was thinking if there was still the possibility of adding a shortcut to make the fill or stroke transparent. 

This would help us save one of the colors. Boxes without background and with background are usually used and right now there is no easy way to do it.

crates/rnote-ui/src/appwindow/actions.rs
flxzt: let's go with what has the best vibe :)

flxzt:
hm, it does not seem to work on my machine, not sure why.

Kneemund:
That's probably due to `<Shift>` being part of the shortcut, which makes keyboards produce special characters instead of numbers. GTK *could* handle this, but it doesn't seem to. Changing the shortcut to e.g. `<Ctrl><Alt>{i}`, like suggested above, works fine.

flxzt:
> Changing the shortcut to e.g. <Ctrl><Alt>{i}, like suggested above, works fine.

sounds good to me.
These new shortcuts also need to be documented in the [shortcuts window](https://github.com/flxzt/rnote/blob/bae2a2fe8ccc4d968b532a583d95640f14ae0d62/crates/rnote-ui/data/ui/shortcuts.ui)

Atreyu-94:
> Wouldn't using a transparent color as color field satisfy this usecase?
No, because that way we lose a color. For example, if I have yellow selected, I would like to be able to select a key to give it 50% transparency and then return it. That would help me to have transparencies for each color and not a single color that I have to be constantly changing to achieve transparency

madonuko:
I'm back! (?)

So I've been thinking: I started from assigning `1~9`, then `<Alt>1~9`, then `<Ctrl><Alt>1~9`, and then I find myself barely using those shortcuts anymore simply because they are not really *short*-cuts.

I was like why not revisit the supposedly simpliest shortcut ver. and fix the issue? So here I am again back to `1~9`.

madonuko:
I'm back again! (?)

I've been using this for basically the last 2 months and I found no issues with the latest implementation!
Final to-do list for self:

- [ ] [`crates/rnote-ui/data/ui/shortcuts.ui`](https://github.com/flxzt/rnote/blob/bae2a2fe8ccc4d968b532a583d95640f14ae0d62/crates/rnote-ui/data/ui/shortcuts.ui)
- [ ] fix merge conflicts

madonuko:
Requires testing and approval

madonuko:
I think the latest ver made it brokey again
- Ctrl + a → typing numbers ⇒ add new texts without replace
- Ctrl + 1 / 2 / 3 / … ⇒ nothing happens

I'm also thinking of setting `0` as the shortcut to make it transparent

madonuko:
I added a `println!()` in the closure to `action.connect_activate`, and nothing gets printed when I press e.g. ctrl+1

I've no idea where the problem comes from…

Doublonmousse:
mh ... Do you build with `meson` ? I'm having similar issue but on main directly with different behavior between a `flatpak` build and a `meson` one (the shortcuts `ctrl+1` to 6 work on flatpak, but don't with meson).
If you run with `GTK_DEBUG=interactive` (and the flatpak equivalent), clicking on the `GtkShortcutController`
image
then shortcuts
image
they are listed.

But If I go inside Data and open the last one
image
the trigger is a list as expected on flatpak
image
with info 
image
and
image

On a meson compiled version, well I get the same thing though the shortcuts are not strictly in the same order (just checked ... )

image

That's a mystery. Well I'm not going to complain too much if at the very least the flatpak version works. Smells like a weird gtk bug (or maybe a combined gtk/packaging issue ?)

And it works on the 0.12 flatpak but not the latest flatpak nightly ...

madonuko:
yes I was building with meson (don't know any other methods)

I feel like there's something in gtk or adwaita not processing shortcuts correctly, especially since I think the major changes are well the code from master which I don't think is the cause, and my system upgrades which apparently updates gtk and stuff to newer versions

to confirm the conspiracy theory however we need a repro... and no idea how to test older versions of gtk on my system

Doublonmousse:
Traced the `ctrl+i` shortcuts not working to a commit, should be fixed with #1515. I probably was on a not up-to-date branch so I didn't bissect this last time

madonuko:
thank you so much!

madonuko:
should be ready

madonuko:
could somebody review?

Doublonmousse:
For me this looks good.
The control shortcuts being the full list from 1 to 9 makes the shortcut window a little bit larger, but it's not a problem I think.

armablign:
and update on this?

Also, would it be possible to maybe add a color wheel (using the defined colors in the color pallet) to be bound to one of the  'button shortcuts' (for drawing tablets/pens)

madonuko:
This PR is ready for review (@flxzt)

For your latter question I don't think I quite get what you mean but we don't have a color wheel in rnote yet afaik.

armablign:
I see.

Regarding the latter, yes I am aware there isn't a color wheel, I apologize if my previous comment didn't convey it more clearly. What I meant was, it would **additionally** be nice to have a color wheel feature, to be able to use the color picker, at the current location of your cursor/pointer (button press -> open color wheel menu, which surrounds current location of cursor).

I came across this feature on [infinipaint](https://github.com/ErrorAtLine0/infinipaint), (was testing other software that might have me move away from rnote) but infinipaint lacked alot of the features I loved from rnote, so, came back. rnote is (to me atleast) only lacking proper shortcuts for changing colors fast.

madonuko:
solved the merge conflict

> it would additionally be nice to have a color wheel feature, to be able to use the color picker, at the current location of your cursor/pointer (button press -> open color wheel menu, which surrounds current location of cursor).

I think I know what you mean now but I think it would be best if you could create a new github issue for that, seems to be out of the scope of this PR.

armablign:
I do apologize for my question burdening,
but, I am new (and unfamiliar) with github's contribution side of things.

I understand a 'pull request' is a request for a contributor to have their code merged to the main project (right?) and obviously requires 'review' (as I saw above).
However, I am unfamiliar with what happens from there moving forward.

Can users manually download the new 'patch'/'feature', is there a 'beta merged' feature? Or do we basically just wait for the main author to merge, and submit it under the releases tab/page?

travis-aaron:
Just want to add my voice here.

The changes are pretty localized. They work with all other current features. The code is easily understandable and I see no issues with it.

I'm a Uni student who uses this for many hours each day, and there are no bugs or regressions that I've been able to identify.

I'd love if we could see this merged.

flxzt:
I changed the shortcut back to `<Ctrl>KP_{1,..,9}` and removed the logic that writes the number as text.

Thanks you for this contribution!

PR #1381, opened 2025-03-05
move is_terminal check after on_conflict check

LeSnake04:
fixes program not using on_conflict when terminal non-interactive

[bug reported in matrix chat](https://matrix.to/#/!ZRnXgXwOYJsMAKmIRz:matrix.org/$z1nGW4HjSjM6Piq7e32CoyuOiVw-CCUUnXEur4JuC-Q?via=matrix.org&via=envs.net&via=kde.org)


LeSnake04:
can be tested with `rnote-cli export selection --output-file=test.svg --on-conflict=overwrite all test.rnote | cat`

flxzt:
thanks!

Issue #1246, opened 2024-10-03
Option to disable/stop scrolling while drawing


puyral:
**Is your feature request related to a problem? Please describe.**  
When one initiates a scroll (with touch or trackpad) it is possible to start drawing while continuing to scroll. This can be quite annoying when one mistakenly starts to scroll just before writing because of poor palm rejection. It seems that it also applies to inertial scrolling.

**Describe the solution you'd like**  
It would be nice to be able to stop any scrolling when the pen gets near / start drawing.

**Describe alternatives you've considered**  
I've looked into disabling the touchscreen entirely, but it doesn't seem possible either.

**Additional context**  
Palm rejection is a difficult problem (but I must say that I'm amazed how well it works already) this should help sidestep the issue.

puyral:
I've been fairly successfull at deactivating touch input on demand: https://github.com/puyral/rnote/tree/add-block-touch

This is my first time ever modifying a GTK app, so I'm assuming I broke many thing.

Doublonmousse:
Pretty similar to #1055.
The code looks pretty good.

A few notes though (maybe open a PR for it, i'll be easier to comment)

https://github.com/flxzt/rnote/compare/main...puyral:rnote:add-block-touch#diff-5a05f1fbb8c3c6082ec6bfadb2b457446578b7af7aa2f1367aafb3730e2b8f31R254
-> bad copy/paste here (wrong name duplicated from above, you're not loading the correct variable here)

https://github.com/flxzt/rnote/compare/main...puyral:rnote:add-block-touch#diff-5a05f1fbb8c3c6082ec6bfadb2b457446578b7af7aa2f1367aafb3730e2b8f31R421
Is changing the propagating phase of the drag gesture not enough ? Is it to stop a canvas drag gesture that's in progress when toggling the option ? (Because of inertial scrolling)

hamdan7m:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

PhilDevProg:
Hey, @puyral
Could you open a pull request for your code? I think it would be greatly appreciated.

puyral:
Done !

Sorry, I didn't have GitHub notifications set up in October, so this flew under my radar...

PR #1385, opened 2025-03-11
feat: Option to disable touch scrolling

puyral:
This is a merge request closing #1246 

This adds a small button to quickly activate/deactivate the touch scrolling on the canvas. For many situations, this is a very good workaround to inconsistent palm rejection.

Remaining to do:
- [ ]  the corresponding icon is black regardless of the theme. I couldn't figure out what takes care of the color of icons
- [ ] activating inertial scrolling make mostly useless

Outside those two points, it has been working flawlessly on my (admittedly outdated) fork for the last 6 months.


Doublonmousse:
For the first point, maybe check https://wiki.gnome.org/HowDoI(2f)CreateSymbolicIconsThatChangeColorAccordingToTheme.html (it's for gtk3 but I think it should suffice).

I think that's the absence of fill color that play a role in the theme working.
For example, there is a `fill="#2e3436"` in https://github.com/flxzt/rnote/blob/33fdab032faaf3bee41e9ed65e286f04988a00b0/crates/rnote-ui/data/icons/scalable/actions/add-page-symbolic.svg?short_path=1fc3354 

I'm not sure I understand the second point. Can you explain in more detail ?

puyral:
>  think that's the absence of fill color that play a role in the theme working.
For example, there is a fill="#2e3436" in https://github.com/flxzt/rnote/blob/33fdab032faaf3bee41e9ed65e286f04988a00b0/crates/rnote-ui/data/icons/scalable/actions/add-page-symbolic.svg?short_path=1fc3354

That was my understanding as well. But it didn't look good without the gray part. It seems to be possible to have some styling because some icons [do it anyway](https://github.com/flxzt/rnote/blob/main/crates/rnote-ui/data/icons/scalable/actions/keyboard-ctrl-space-shortcut-symbolic.svg?short_path=c46929f)... I'll look at the gnome documentation you linked because it seems to hold the answer I'm looking for.

> I'm not sure I understand the second point. Can you explain in more detail ?

From what I understand, around [here](https://github.com/flxzt/rnote/pull/1385/files#diff-5a05f1fbb8c3c6082ec6bfadb2b457446578b7af7aa2f1367aafb3730e2b8f31R430-R466) I'm "blocking" all events linked to drag motions. But apparently if it comes from the intertial/kinetic scrolling, it goes somewhere else (I'm not sure where). In practice this means that gtk still recognize the touch input and generate inertial scrolling for them which is then not ignored.
So with inertial scrolling activated, when the palm rejection fails it still move the canvas (even though you can't really move it "willingly").

puyral:
A month later, I finally had time to work on this again ^^'

As far as I can tell, all the problems I listed above are fixed. I ended up changing the icon to another one straight from the gtk repository because I couldn't make it follow the theme...

julianfairfax:
Could this maybe be looked at again? It would be practical to have this feature.

puyral:
Well, I far as I can tell, my branch works (I've been using it since my last comment, and it hasn't crashed on me).

I need to rebase anyway because this is the version of `rnote` that I use, and an update is long overdue. I'll try to do that this weekend if I get the time.

puyral:
just so you know everything has been rebased and is apparently working.

crates/rnote-ui/src/canvaswrapper.rs
```diff
@@ -305,9 +324,30 @@ mod imp {
...
+        fn canvas_touch_pan_update(&self) {
+            if !self.block_touch.get() && !self.canvas.touch_drawing() {
+                self.canvas_drag_gesture
+                    .set_propagation_phase(PropagationPhase::Bubble);
+                self.touch_two_finger_long_press_gesture
+                    .set_propagation_phase(PropagationPhase::Capture);
+                self.touch_long_press_gesture
+                    .set_propagation_phase(PropagationPhase::Capture);
+            } else {
+                // set everythinbg to `None`
```
Doublonmousse, line 336: typo

crates/rnote-ui/src/canvaswrapper.rs
```diff
@@ -405,6 +445,9 @@ mod imp {
                     #[weak(rename_to=canvaswrapper)]
                     obj,
                     move |_, _, _| {
+                        if canvaswrapper.block_touch() {
```
Doublonmousse, line 448: does the canvas drag gesture activate even with the propagation phase set to none ? or is it to stop a gesture in progress when clicking on the button ?

I _think_ the latter _could_ be done with https://gtk-rs.org/gtk4-rs/stable/latest/docs/gtk4/prelude/trait.GestureExt.html#method.set_state (EventSequenceState::Denied)

puyral: I am a bit unsure, I was following what `block_pinch_zoom` was doing, but it does not appear here. So maybe. 


I would have to test again make sure, but I won't get access to my touch screen device for 3 more weeks...

flxzt: is there any progress on this? @puyral

flxzt:
I'll set to draft until the remarks are resolved. Feel free to mark it ready again.

Issue #359, opened 2022-11-16
Enable one-click building in Builder


bertob:
To make contributions easier it'd be great if this app could be built in GNOME Builder. This requires a `json` flatpak manifest somewhere in the repo. There does seem to be a manifest in this case, but it's `yaml`, so it doesn't work. Simply converting the manifest should work.

flxzt:
You can always build natively with meson. But you are right, gnome builder currently [does not support yaml](https://gitlab.gnome.org/GNOME/gnome-builder/-/issues/520) . 

I want to keep the manifest in yaml format, because it is a lot more pleasant to write and supports comments, but it is easy to convert to json with `yq` by running:
```
yq -o=json com.github.flxzt.rnote.Devel.yaml > com.github.flxzt.rnote.Devel.json
```

I'll update the json whenever I modify the yaml file.

PR #1391, opened 2025-03-20
Bump GNOME Runtime to Version 48

RayJW:
New GNOME release, new runtime. Just the regular maintenance work to make sure, that the next release is up to speed, so we won't run into an EOL warning in a year's time again.

Did some testing for a few minutes with the usual stuff included in my workflow and I see no functional difference.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -59,7 +59,7 @@ flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.f
 The flatpak Gnome Runtime, SDK and some extensions are needed:
 
 ```bash
-flatpak install org.gnome.Platform//47 org.gnome.Sdk//47 org.freedesktop.Sdk.Extension.rust-stable//24.08 \
+flatpak install org.gnome.Platform//48 org.gnome.Sdk//48 org.freedesktop.Sdk.Extension.rust-stable//24.08 \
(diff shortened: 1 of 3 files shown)


flxzt:
Thanks! The documentation needs an update as well: https://github.com/flxzt/rnote/blob/main/BUILDING.md#building-with-flatpak

RayJW:
> Thanks! The documentation needs an update as well: [`main`/BUILDING.md#building-with-flatpak](https://github.com/flxzt/rnote/blob/main/BUILDING.md?rgh-link-date=2025-03-24T19%3A33%3A09Z#building-with-flatpak)

You're right, how did I miss that? :D Anyway, adjusted that now.

RayJW:
Btw. is there any reason that you maintain both the `.yaml` and the `.json` manifest? Realistically there is no need since either is perfectly fine, and I'd personally advise on just keeping the JSON since that's what Flathub uses for some other necessary files. Of course, there is no strong reason to get rid of one, but I thought I'd make sense to avoid some possible freak problem down the line. An overlooked diff between the two could lead to different builds etc.

flxzt:
Thanks!

> Btw. is there any reason that you maintain both the .yaml and the .json manifest?

yes, because Gnome Builder does not support YAML manifests. See: https://github.com/flxzt/rnote/issues/359#issuecomment-1318580727

PR #1396, opened 2025-03-24
chore: add justfile for common tasks, adapt CI

flxzt:
adds a justfile (https://just.systems/man/en/) for common tasks and adapts the CI to use it.
Additionally now that there is a `fmt` recipe that also formats meson files. It's formatting is applied in this PR.
YAML files are also checked with `yamllint`, and it's suggestions applied here.


flxzt:
ready for review. @Kneemund if you have some time, please take a look, especially at the CI stuff :)

justfile
```diff
@@ -0,0 +1,234 @@
...
+build_folder := "_mesonbuild"
+flatpak_app_folder := "_flatpak_app"
+flatpak_repo_folder := "_flatpak_repo"
+mingw64_prefix_path := "C:/msys64/mingw64"
+
+sudo_cmd := if ci == "true" {
+    ""
+} else {
+    "sudo"
+}
```
Kneemund, line 16: Isn't running sudo in GH Actions just a no-op, or is this necessary?

justfile
```diff
@@ -0,0 +1,234 @@
...
+        {{sudo_cmd}} apt-get update
+        {{sudo_cmd}} apt-get install -y \
+            yamllint yq opencc
+    else
+        echo "Unable to install system dependencies, unsupported distro."
+        exit 1
+    fi
+
+    if [[ "{{ci}}" != "true" ]]; then
+        cp -f build-aux/git-hooks/pre-commit.hook .git/hooks/pre-commit
```
Kneemund, line 92: We should probably use a symlink here, so that changes are applied without having to re-run this command.

```suggestion
        ln -sf build-aux/git-hooks/pre-commit.hook .git/hooks/pre-commit
```

.github/workflows/ci.yml
```diff
@@ -1,6 +1,8 @@
+---
+
 name: CI
 
-on:
+"on":
```
Kneemund, line 5: Did a formatter do this? It's the only key that's explicitly a string, which looks a bit weird. Some languages treat on/off as booleans (where this would make more sense), but I don't think YAML is one of them (?).

justfile
Kneemund, line 1: We could add support for arch-based distros here, then we'd cover most of the popular package managers.

The GitHub syntax highlighting complains about a few things here, but that's probably just a bug.

Kneemund (commented):
Looks good, I've added a few of my thoughts. Everything seems to work in the workflows, and I didn't catch anything that looks wrong :)

.github/workflows/ci.yml
flxzt: Apparently YAML by standard does treat it as boolean, at least yamllint complained which caused me to change this (see: https://github.com/adrienverge/yamllint/issues/430). Seems that Github implements a custom parser that does not conform to the spec

justfile
flxzt: good idea!

justfile
flxzt: Wouldn't it be dependent on the used job container image? This originally came from a gitlab workflow. I am not sure if it's needed. I'll keep the `sudo_cmd` variable, but assign it unconditionally to `sudo`.

justfile
flxzt: True, I'll check if I can follow-up with that.

> The GitHub syntax highlighting complains about a few things here, but that's probably just a bug.

Yeah, it seems it can't yet handle justfile syntax.

justfile
Kneemund: That's true, I was assuming that it's root. If it's not root, then we'd need sudo anyways. This should always work as long as sudo is pre-installed (or there's some kind of shim) in the image.

flxzt:
thanks for the review!

PR #1401, opened 2025-04-05
Missing tooltips

Doublonmousse:
A couple of small updates for tooltip text.
- Some of them were missing for the header bar and tool section
- Added generic tooltips for spin buttons


flxzt:
awesome, thanks!

Issue #1372, opened 2025-02-25
Buggy in program UI size (looks bigger than it actually is)


SeongGyu1216:
**Describe the bug**  
The UI of the program seems to be a bit larger than the actual size it appears to be.

**To Reproduce**  
When in windowed mode, it seems to recognize it as slightly larger than it actually is.
In fact, if I put another program behind rNote and click on the outer border of rNote, it should show the program behind it as active, but when I click on the
It doesn't show up, and I have to click beyond a certain distance to activate the program behind it.

**Desktop (please complete the following information):**  
- Windows 11, version 24H2
- Rnote v0.11.0
- mouse, tablet pen

**Additional context**  
I want it to look like the size it actually is.

Doublonmousse:
I think that's a windows-specific gtk4 bug 
See https://gitlab.gnome.org/GNOME/gtk/-/issues/6255#note_1952796 and 

Gaphor uses a workaround in the meantime (see https://github.com/gaphor/gaphor/commit/51671742dd65de3b194b8f40438080f5ccae47b0 and https://github.com/gaphor/gaphor/blob/597140ba014586e163bafcaf47b48f1ba6fba4bb/gaphor/ui/styling-windows.css#L1C1-L14C1 )
This is basically some custom css 

Until gtk4 uses native shadow (instead of drawing themselves the shadow and having a larger than the window clickable area) this is the only fix there is (The upstream issue for gtk4 is here : https://gitlab.gnome.org/GNOME/gtk/-/issues/7001).

Should we follow gaphor and implement the workaround ?

SeongGyu1216:
It's a bug with the Windows operating system,...

I see, there's no problem when using it in full screen, but it's only slightly inconvenient when in windowed mode, so it's okay. Still, I hope it will be fixed someday.

Doublonmousse:
It's not really a windows OS bug, more a bug in gtk4.

I'm trying the gaphor workaround on rnote and it's working pretty well overall, reducing by a lot this (pretty large) distance you'd have to be from the window's borders for clicks to work.

I think it's worth implementing the workaround (and removing it once https://gitlab.gnome.org/GNOME/gtk/-/issues/7001 is closed)

Doublonmousse:
Fixed by #1403

Issue #842, opened 2023-10-18
Black Borders when not in fullscreen on windows


winpower21:
**Describe the bug**  
Thick black borders around the window when not in fullscreen on windows

**To Reproduce**  
1. Install Rnote on windows 11
2. Run Rnote

**Expected behavior**  
No black borders around the window.

**Screenshots**  
![image](https://github.com/flxzt/rnote/assets/95289325/6c4a674a-9506-421f-ae0c-d8a1cf53f166)

Desktop:
 - OS: Windows 11 22H2 22621.2428
 - App Version: Rnote v0.9.0
 - Installation Source: .exe installer as well as winget

flxzt:
Is this still consistently happening for you? It looks like the compositing failed. I saw this personally on linux as well, but not specifical to Rnote and usually only when a crash in the desktop environment/the compositor happened.

Doublonmousse:
Maybe related ? https://gitlab.gnome.org/GNOME/gtk/-/issues/6401 
It _might_ be fixed on gtk 4.14 regardless of the renderer for nvidia graphics cards (gl or ngl, the gl one was reported to have black borders)

Doublonmousse:
From #1197
[...]

PR #1403, opened 2025-04-06
add custom css for windows

Doublonmousse:
Fixes/workaround for #1372 (this makes it possible to click outside the window without having the mouse click intercepted).

Taken from https://github.com/gaphor/gaphor/blob/1f53cfa8cc8eced7ce8b6776218d218dc121fae3/gaphor/ui/styling-windows.css#L4-L10

```
 * The issue: shadows defined by LibAdwaita are very wide: up to 60px.
 * The shadow is drawn inside the (clickable) window.
 *
 * These styles change the (window) background to be a lot more compact.
 * This results in a smaller shadow around the window, and as such a
 * smaller click surface for the whole window, making it feel less awkward
 * when you click just outside the window.
```

This potentially helps as well when the shadow isn't transparent #842 as the shadow won't be as large if it doesn't render (maybe we can remove the shadow entirely using the same technique if we wanted to)

### DIFF
--- a/crates/rnote-ui/src/env.rs
+++ b/crates/rnote-ui/src/env.rs
@@ -124,3 +124,45 @@ fn macos_is_in_app_bundle(canonicalized_exec_dir: impl AsRef<Path>) -> bool {
             }
         })
 }
+
+#[cfg(target_os = "windows")]
+/// Workaround for windows for shadow that intercept mouse events outside of the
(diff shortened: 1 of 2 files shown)


flxzt:
looks good, thanks!

> maybe we can remove the shadow entirely

not sure this is a good idea, if windows are beneath with similar colors without a shadow being drawn it can be problematic. But this workaround looks like a significant improvement.

PR #1404, opened 2025-04-06
improv: updated color fills for indicators to improve contrast

### LINKED ISSUE #919: The selection icon for text is hard to see on blue background
The selection icon for text is hard to see on blue background.
I first thought there were a bug and the icon was missing.
image

Version 0.9.2

Doublonmousse:
This updates the fill colors for indicators.

Edit : The new strategy is to take the background color and fill the indicator nodes with it (before applying the proximity/down colors)

This fixes #919

Light mode
image
image
Dark mode
image
image
image

What I haven't done is do anything to the background color (should I set an alpha value ?).


flxzt:
I think it is good as it is, alpha 100% for the node background colors is not necessary in my opinion!

PR #1405, opened 2025-04-06
Improv : logic changes for selection cancellation

### LINKED ISSUE #414: Define exit behavior of the selection tool

### LINKED ISSUE #888: Unable to make a new selection with pen button while having another active selection.
**Describe the bug**  
When using the buttons on the pen to select something, that button does not work again until the selection has been cancelled again. That means when e.g. moving things around after selecting one part and moving it, you first have to cancel the selection e.g. switching tools before the selection tool works again.

**To Reproduce**  
1. Use pen hotkey to select content
2. Do whatever with the selection, e.g. scale, move that keeps the selection, so not deleting
3. Try to do another selection with the pen button
4. Instead of the selection tool, it will just start to draw

**Expected behavior**
I think generally, when pressing a "hotkey" again after releasing it, the intended behaviour would be for the action to get triggered again. An example workflow would be selecting something while pressing the pen button with the lasso tool, moving it where I want the selection to be, and then selecting something else that I want to move to the previous selection's place without first writing something new. I think this wouldn't break any workflows, since you could still just start writing if that's what you want to do.

**Console Output**  
```bash
2023-11-16T15:15:10.571Z DEBUG rnote > ... env_logger initialized

(rnote:2): Gtk-CRITICAL **: 16:15:10.820: gtk_widget_class_set_accessible_role: assertion '!gtk_accessible_role_is_abstract (accessible_role)' failed
```

**Screenshots**  
[Video demo](https://github.com/flxzt/rnote/assets/29835902/34ad0b36-bdc7-466b-8c62-96ef7ea894fa)

 - OS: Fedora 39
 - App Version: Rnote 0.9.2
 - Desktop Environment: 45.1

Doublonmousse:
Fixes #414
Fixes #888 

- This fixes the small dots that appears on selection cancellation with the pen (whilst allowing for strokes to still be written if they're larger than small dots)

A couple remarks : 
- ~~shapes need an additional up/down cycle : we could change it to follow the same logic as the brush but this could also be done by not changing the logic of the shape part and only disallowing shapes to be too small (in that case we can create shapes that aren't even visible but are still part of the document)~~
- the eraser and typewriter both need an additional pen up/down to activate
- tools aren't affected (they won't change anything if they're activated on such a short pen down/up cycle)
- ~~This does not change behavior when a selection is deleted.~~

Rebase of the previous PR #1139 with the new clone! syntax

> I've added a condition to only create shapes if they're large enough. As things like arrows would still appear on a single pen event, I'll leave the current behavior for shapes after a selection cancellation (that is we need an extra up/down cycle)

> I've added the cancellation of the selection tool upon deleting a selection when the selection tool is temporary

> Putting back in draft as this needs a little work before it can be merged. I'm wondering if the condition can be slightly modified to include time delay as well. It will cancel a dot after a selection but sometimes this dot can be large enough to stay (if the pen slides a bit instead of a clean hit)

### DIFF
--- a/crates/rnote-engine/src/pens/selector/penevents.rs
+++ b/crates/rnote-engine/src/pens/selector/penevents.rs
@@ -23,6 +23,7 @@ impl Selector {
         modifier_keys: HashSet<ModifierKey>,
         _now: Instant,
         engine_view: &mut EngineViewMut,
+        temporary_tool: bool,
     ) -> (EventResult<PenProgress>, WidgetFlags) {
         let mut widget_flags = WidgetFlags::default();
         self.pos = Some(element.pos);
@@ -201,13 +202,22 @@ impl Selector {
                             // when clicking outside the selection bounds, reset
                             tracing::debug!("cancelling the selection");
                             engine_view.store.set_selected_keys(selection, false);
-                            self.state = SelectorState::Idle;
 
-                            // This event is in the same sequence than the one that
-                            // cancelled the selection. We thus set the variable to true here
-                            engine_view.store.set_cancelled_state(true);
+                            if temporary_tool {
+                                // we had a selection active then for the next sequence
+                                // we activated the temporary mode for the selector and
+                                // clicked outside the selection. We expect to be able
+                                // to do another selection
+                                self.state = SelectorState::Selecting { path: Vec::new() };
+                                progress = PenProgress::InProgress;
+                            } else {
+                                self.state = SelectorState::Idle;
+                                progress = PenProgress::Finished;
 
-                            progress = PenProgress::Finished;
+                                // This event is in the same sequence than the one that
+                                // cancelled the selection. We thus set the variable to true here
+                                engine_view.store.set_cancelled_state(true);
+                            }
                         }
                     }
                     ModifyState::Translate {
--- a/crates/rnote-engine/src/pens/selector/penevents.rs
+++ b/crates/rnote-engine/src/pens/selector/penevents.rs
@@ -203,10 +203,8 @@ impl Selector {
                             engine_view.store.set_selected_keys(selection, false);
                             self.state = SelectorState::Idle;
 
-                            // we create a new engine variable that indicates that the pen event is one from the same
-                            // event sequence that the one that cancelled the selection
-                            // this is done by setting this variable to true here and
-                            // removing it on a pen up event (and doing more things if this variable is true on a pen up)
+                            // This event is in the same sequence than the one that
+                            // cancelled the selection. We thus set the variable to true here
                             engine_view.store.set_cancelled_state(true);
 
                             progress = PenProgress::Finished;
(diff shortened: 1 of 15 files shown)


Doublonmousse:
I've added a commit to fix #888. This last commit needs some testing (and maybe some renaming) to be sure everything works as expected.

Doublonmousse:
Rebased to main, seems to compile/work as before.
I think I still need to retest on tablet though

crates/rnote-engine/src/pens/selector/penevents.rs
```diff
@@ -198,10 +199,24 @@ impl Selector {
...
                             engine_view.store.set_selected_keys(selection, false);
-                            self.state = SelectorState::Idle;
 
-                            progress = PenProgress::Finished;
+                            if temporary_tool {
+                                // we had a selection active then for the next sequence
+                                // we activated the temporary mode for the selector and
+                                // clicked outside the selection. We expect to be able
+                                // to do another selection
+                                self.state = SelectorState::Selecting { path: Vec::new() };
```
Doublonmousse, line 210: I was thinking maybe having a 'if shift' condition here would be good (so doing shift + new selection can add to what's already here). 
Now that would mean we'd have to change the SelectorState strcture (so that we can select and have the previous selection still saved as an optional)

Doublonmousse:
Retested and working almost as intended.

On smaller stroke width, the heuristic I'm using for cancelling small strokes doesn't work very well (volume comparison to a * stroke_width  ^ 2). It works if the pen is orthogonal to the surface but will still leaves traces behind if you slip a little.

Maybe calculating the perimeter of the outline and comparing it to a fixed threshold * zoom will work better (so that we essentially check whether the stroke path is small or not compared to the current view).
I'll try but if that still fails I'll fallback to an additional up/down cycle

Doublonmousse:
I've changed the computation for the cancellation. It seems to work better though not totally foolproof. I'm still on the fence about switching back to an additional up/down cycle for the selection cancellation to pen stroke situation. 
Other that than, does what I expect

flxzt:
> I've changed the computation for the cancellation. It seems to work better though not totally foolproof. I'm still on the fence about switching back to an additional up/down cycle for the selection cancellation to pen stroke situation.
Other that than, does what I expect

I am not up to speed with the latest changes, did you change anything in that regard or is that still the latest state?

Doublonmousse:
No, I only rebased so I left it as is, this is still the latest state.

PR #1413, opened 2025-04-13
adding refocus if new file opened from file manager

### LINKED ISSUE #1411: Always request focus upon opening a file from the file explorer
When opening a rnote file from the file explorer with rnote closed, the window appears and takes focus.

But opening a second file does not (it opens the file in a new tab on the same window but does not request focus upon opening the file)

Request focus when opening files in the second case.

Maybe we can also improve tab gestion (detach method to separate what is two tabs into two separate windows) and/or prefer opening new files in a new window vs a new tab

Niggi1105:
this closes #1411 .


flxzt:
thanks!

PR #1419, opened 2025-04-17
fix: unicode segmentation for inital selection when renaming file

Kneemund:
Changes the stem selection logic when renaming files to use unicode segmentation, since GTK expects character instead of byte indices.

|Previously|Now|
|---|---|
|grafik|grafik|


flxzt:
looks good, thanks!

PR #1420, opened 2025-04-17
Fix cancel not sent for import dialogs

Doublonmousse:
Pressing escape on dialogs did not send back the value so the progress bar would never stop

### DIFF
--- a/crates/rnote-ui/src/dialogs/import.rs
+++ b/crates/rnote-ui/src/dialogs/import.rs
@@ -476,22 +476,13 @@ pub(crate) async fn dialog_import_pdf_w_prefs(
     }));
 
     // Overwrite builtin close shortcut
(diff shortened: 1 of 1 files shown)


crates/rnote-ui/src/dialogs/import.rs
```diff
@@ -476,22 +476,13 @@ pub(crate) async fn dialog_import_pdf_w_prefs(
     }));
 
     // Overwrite builtin close shortcut
```
flxzt, line 478: Both comments above dialog.connect_closed()` are then out of date

Doublonmousse: fixed

flxzt:
thanks, lgtm!

Issue #1400, opened 2025-04-05
Moving selection always snaps to grid


madonuko:
**Describe the bug**  
All attempts to move selected objects cause the object to awkwardly snap to grid.

**To Reproduce**  
1. Draw a stroke
2. Select it
3. Move it around

**Expected behavior**  
The old behaviour… but the weirdest thing is I did NOT upgrade rnote and most of its dependencies. I tried downgrading them, issue persists. Tried using flatpak, issue persists. It just appeared out of nowhere I swear.

**Console Output**  
```
2025-04-05T11:33:22.589007Z DEBUG rnote: .. tracing subscriber initialized.

(rnote:2): Gtk-CRITICAL **: 19:33:33.823: The layout manager of type GtkBoxLayout 0x5636ce1cd4f0 does not create GtkLayoutChild instances
```

**Screenshots**  

https://github.com/user-attachments/assets/b9906502-712b-4e39-97eb-482fa4e0010d

**Desktop (please complete the following information):**  
 - OS: Ultramarine Linux 41
 - App Version: tried 972fa2ec and 0.11.0
 - Installation Source: Flatpak, `meson compile -C build`, Terra rpm, etc.
 - Desktop Environment: KDE Plasma 6.3.4
 - Display Server: Wayland
 - Input Source: Mouse and Keyboard

**Additional context**  
Also tried downgrading various system components that rnote relies on (e.g. cairo)
And figured this could be useful:
```
$ dconf dump / | sed -n '/rnote/,/^$/p'
[com/github/flxzt/rnote]
active-fill-color=(0.0, 0.0, 0.0, 0.0)
active-stroke-color=(0.87843137979507446, 0.10588235408067703, 0.14117647707462311, 1.0)
autosave-interval-secs=uint32 60
brush-width-1=2.0
brush-width-2=4.0
brush-width-3=12.0
colorpicker-color-1=(0.87843137979507446, 0.10588235408067703, 0.14117647707462311, 1.0)
colorpicker-color-2=(0.10196078568696976, 0.37254902720451355, 0.70588237047195435, 1.0)
colorpicker-color-3=(0.14800000190734863, 0.63200002908706665, 0.40999999642372131, 1.0)
colorpicker-color-5=(0.89800000190734863, 0.37799999117851257, 0.0, 1.0)
colorpicker-color-6=(0.86274510622024536, 0.54117649793624878, 0.86666667461395264, 1.0)
[...]

PR #1428, opened 2025-04-24
adding snap to grid button

Niggi1105:
same as the previous pr (just messed up my gh)


crates/rnote-ui/data/icons/scalable/actions/snap-to-grid-symbolic.svg
```diff
@@ -0,0 +1,10 @@
...
+
<g id="SVGRepo_iconCarrier">
+
<path d="M10 4H6V1H5v3H1v1h4v4H1v1h4v4H1v1h4v4H1v1h4v3h1v-3h4v3h1V1h-1zM6 5h4v4H6zm0 5h4v4H6zm0 9v-4h4v4zM17.8 6H12v4h5.8a2 2 0 0 1 0 4H12v4h5.8a6 6 0 0 0 0-12zM13 7h2v2h-2zm0 8h2v2h-2zm4.8 2H16v-2h1.8a3 3 0 0 0 0-6H16V7h1.8a5 5 0 0 1 0 10z"/>
+
<path fill="none" d="M0 0h24v24H0z"/>
+
</g>
+
</svg>
```
Doublonmousse, line 10: This svg doesn't work on windows after building (https://github.com/Doublonmousse/rnote/actions/runs/15083573129).

I'm proposing the following changes : see commit https://github.com/flxzt/rnote/commit/bf8ae936ffa0ad11418764f1f85649b44c9a1938 after what the icon works

(successful build with the icon showing correctly here : https://github.com/Doublonmousse/rnote/actions/runs/15083917699 )

Doublonmousse (changes requested):

crates/rnote-ui/data/icons/scalable/actions/snap-to-grid-symbolic.svg
Nugfle: updated it

Doublonmousse (approved):

Doublonmousse:
Also : fixes #1400

flxzt:
Okay, I am a sceptical here because of the following arguments:

- I don't want to clutter the UI more than necessary, the canvas menu was intended to place settings that are sometimes changed, where I think the snap to grid option fits.
- The snap to grid is not specific to the selector, so it wouldn't be the right place (and being described as "Snap to Grid when moving the selection" is incorrect because it also affects the text and (recently added) the shaper tool.

Nugfle:
I think the biggest issue with the current design is that users (including myself) tend to not find the canvas menu, also mentioned in #1400. Idk if we can make it so the menu is in a more intuitive location.

flxzt:
I am open suggestions for an improved location. Also, would a different icon help in that case? Something that would suggest more like "in here are actions that can be performed and options that probably are changed frequently"?

flxzt:
I will close, I think an additional button is not the way to go. I am very open to improve the location and improve the discoverability of the canvas-menu though, if someone has a suggestion.

PR #1429, opened 2025-04-24
fixing refocus if reopening already open file

Nugfle:
now we should also gain focus if reopening a file, that is already open in another tab.


flxzt:
looks good, thanks!

PR #1435, opened 2025-05-02
refactor: global engine configuration, save/restore document presets

### LINKED ISSUE #1414: Close all tab from Tabs manager reset the settings
I saw that in the new version of Rnote (0.12) the possibility of seeing all the active tabs in a grid and eventually closing them has been added. The problem I'm having is that when you close all the tabs and then reopen a new one, or close the app, all the saved configurations are reset (eraser size, brushes, right mouse button actions or pen buttons etc.). This can be a bit frustrating because you have to reconfigure everything manually.

PS: is there a way to export personal configurations to keep them as a backup? I use windows and therefore I can't use gsettings

flxzt:
These changes split the engine configuration into a part that is "global" and is shared across all engines (tabs) via shared pointer `Arc<RwLock<EngineConfig>>`. The UI application window object holds this configuration as well, and all changes must now be applied to it because the configuration is supposed to be supplied externally and once installed on the engine with `install_config()` it is not visible outside of the engine anymore.

In addition the `Document` struct now has a `DocumentConfig` sub-struct that holds per-engine document configuration. The UI settings now have the ability to save and restore document settings presets that get applied whenever a new engine (tab) is created.

Because the `Document` struct is saved in the file as well, this is a change in the file format. Luckily the upgrade path is relatively trivial.

I also moved the "show-format-borders" and "show-origin-indicator" options to the document settings, which I think makes more sense.

should fix #1414


flxzt:
I think I am done, if anyone wants to test this here's a nightly flatpak :) https://github.com/flxzt/rnote/actions/runs/14914020460

flxzt:
I think I am done, if anyone wants to test this here's a nightly flatpak :) https://github.com/flxzt/rnote/actions/runs/14914083927

flxzt:
thanks for testing! should be fixed with [169df8d](https://github.com/flxzt/rnote/pull/1435/commits/169df8da3b75ca2267814615d7b00b6a1442dd4a)

Nugfle:
All seems to work just fine now.

Issue #1436, opened 2025-05-03
Performance degrades on larger files


Intranox:
I love Rnote and that's why I decided to use it to take notes. The problem I'm having is that as the amount of elements in the canvas increases (I use the continuous vertical) the performance tends to degrade making it difficult to write fluidly and see the precise lines of what you've written on the screen, more than anything else only incomprehensible signs appear. I read a bit about the project before writing here and I noticed that there are already some ideas about changing the format to make it better. I understand the complexity behind it and that it will take time, but do you think that these changes will solve this problem? Are there currently any solutions to the problem other than separating the content into multiple files?
I'll leave a demonstration video, the problem in this case is even more amplified due to the screen recording (I use Windows, the file in question is 8 pages)

Video:

https://github.com/user-attachments/assets/9a7a87c3-8a07-459b-814b-0345f3a105d0

Niggi1105:
What is the size of the file that you made this video on? Behaviour like the one in the video is not supposed to happen no matter the size of the file. Can you activate developer mode and show visual debugging information (in the top right menu)? There should be information on the total amount of strokes etc.

Intranox:
> What is the size of the file that you made this video on? Behaviour like the one in the video is not supposed to happen no matter the size of the file. Can you activate developer mode and show visual debugging information (in the top right menu)? There should be information on the total amount of strokes etc.

The .rnote file size is 1.42MB, the file contains quite a bit of text (not handwritten but keyboard-typed), a couple of images and several hand-drawn strokes.

![Image](https://github.com/user-attachments/assets/beb8014c-05a4-42db-aee8-82d3242895a7)

Niggi1105:
This should really not cause an performance issue than. What are your hardware specs? Have you tried monitoring resource usage of rnote and your machine in general?

Intranox:
> This should really not cause an performance issue than. What are your hardware specs? Have you tried monitoring resource usage of rnote and your machine in general?

As long as there are not many strokes everything works fine, the problem begins when the documents start to get bigger (not in the sense of memory occupied but in the number of elements of the json file that describes the canvas). I give you the performance of the hw in which I am using it which is not monstrous but should be acceptable:
CPU: AMD Athlon 300U
GPU: AMD Radeon Vega 3 (Picasso) 2GB
RAM: 8GB DDR4 SDRAM
OS: Windows 10 Home (x64)

88neil:
[...]

Issue #1418, opened 2025-04-17
Black window on Windows 11


thedevleon:
**Describe the bug**  
Rnote shows as a black / blank window on Windows 11.
The cursor changes to a dot when hovering the window, and clicking in the rough area of the maximise button also maximizes the window (still black).

**To Reproduce**  
1. Launch RNote
2. Get a black window

**Expected behavior**  
RNote should launch normally.

**Console Output**  
Unable to do in windows. I tried launching it via CMD and Powershell, but get no console output.
```
PS C:\Program Files\Rnote\bin> .\rnote.exe
PS C:\Program Files\Rnote\bin>
```

**Screenshots**  
![Image](https://github.com/user-attachments/assets/bc3e584a-e6ed-48b4-9043-4b3c28042737)

**Desktop (please complete the following information):**  
 - OS: Windows 11 Pro, 24H2, 26100.1301
 - CPU: AMD Ryzen AI 9 HX 370 w/ Radeon 890M
 - GPU: NVidia RTX 4060 Laptop
 - App Version: Rnote v0.12.0
 - Installation Source:  GitHub (https://github.com/flxzt/rnote/releases/tag/v0.12.0)

thedevleon:
I also tried 0.11.0 and 0.10.0, same issue. Graphics drivers are up to date.

thedevleon:
After installing the dependencies for building from source (https://github.com/flxzt/rnote/blob/main/misc/building/rnote-windows-build.md#dependencies), the window is no longer black, but there's still a black outline around the window:

![Image](https://github.com/user-attachments/assets/949ba138-4192-40fb-8438-f41ee803ba61)

Doublonmousse:
For the last one, that's #842 (so changes have to happen with gtk4 upstream).

For building the application, are you running it from the msys terminal or did you create the .exe and then installed it ?

dax-leo:
Same problem on Windows 10. Black screen. Didn't try to compile it, just reverted back to 0.11.

Doublonmousse:
Does this happen with the build from https://github.com/Doublonmousse/rnote/actions/runs/15075694294 ?
This is from #1443, using windows server 2025 (instead of 2022) for building rnote on windows.

As compiling from source seems to work, seems like this is more related to the build process (or gtk4 version ?)

PR #1443, opened 2025-05-08
update windows ci to 2025 version

Doublonmousse:
- update windows to 2025 version
- install inno (done manually as this isn't part of the image otherwise)
- set `cli=true` for window ci (fails otherwise, because refers to a `cargo_build` target only defined if the cli is set to true)

This makes it possible to use the 2025 version for compiling on the CI.

TODO
- [ ] fix the inno version (set to a specific version)
- [ ] checksum and validation (checksum on the exe + verify signature before launching)
- [ ] changing how the cli parameter is setup ?

### DIFF
--- a/.github/workflows/release-windows.yml
+++ b/.github/workflows/release-windows.yml
@@ -9,7 +9,7 @@ name: Release Windows
 
 jobs:
   build:
-    runs-on: windows-2022
+    runs-on: windows-2025
     permissions:
       # needed for uploading release artifact
       contents: write
@@ -17,6 +17,12 @@ jobs:
       run:
         shell: msys2 {0}
     steps:
+      - name: Install Inno 6
+        id: install_inno
+        shell: pwsh
+        run: |
+           Invoke-WebRequest -Uri https://jrsoftware.org/download.php/is.exe -OutFile D:\\is.exe
+           Start-Process -FilePath D:\\is.exe -ArgumentList "/VERYSILENT", "/SUPPRESSMSGBOXES", "/NORESTART", "/SP-" -Wait
 
       - name: Set installer name
         id: set_installer_name
--- a/justfile
+++ b/justfile
@@ -123,7 +123,7 @@ setup-win-installer installer_name="rnote-win-installer":
     meson setup \
         --prefix={{ mingw64_prefix_path }} \
         -Dprofile=default \
-        -Dcli=false \
+        -Dcli=true \
         -Dwin-installer-name={{ installer_name }} \
         -Dci={{ ci }} \
         {{ build_folder }}


Doublonmousse:
This may be also related to issue #1418 (so maybe using a more recent windows image fixes it ?)

.github/workflows/release-windows.yml
```diff
@@ -9,14 +9,27 @@ name: Release Windows
...
+        run: |
+           $exe_out = Join-Path (Get-Location) is.exe
+           Invoke-WebRequest -Uri https://files.jrsoftware.org/is/6/innosetup-6.4.3.exe -OutFile $exe_out
+           $is_signed = Get-AuthenticodeSignature $exe_out
+           $hash = Get-FileHash $exe_out -Algorithm SHA256
+           If(($is_signed.Status -eq 'Valid') -and ($hash.Hash -eq 'f3c42116542c4cc57263c5ba6c4feabfc49fe771f2f98a79d2f7628b8762723b')) {
+             Start-Process -FilePath $exe_out -ArgumentList "/VERYSILENT", "/SUPPRESSMSGBOXES", "/NORESTART", "/SP-" -Wait 
+             } else {
+             throw "failed to verify signature"
+             }
```
Kneemund, line 32: Is it possible to use winget here? From what I've read, it is pre-installed in windows-2025. Should be something along the lines of `winget install JRSoftware.InnoSetup --accept-package-agreements --accept-source-agreements --disable-interactivity`.

Doublonmousse: Trying this here : https://github.com/Doublonmousse/rnote/actions/runs/15084280914 
Just have to wait until it finishes

Kneemund: Maybe we have to add `--scope machine`? I think it defaults to user space.
https://learn.microsoft.com/en-us/windows/package-manager/winget/install

Doublonmousse: Apparently winget doesn't tell you where the package is installed, it could be in the `Program Files`, could be in the `Program Files (x86)`, could be in `%LOCALAPPDATA%\Programs` ...

In that case it was in the last one (so the add to path failed)

Doublonmousse: oh it did work https://github.com/Doublonmousse/rnote/actions/runs/15084776262/job/42405943404
I'll remove the dir but keep the cd in that case (maybe a fancier try/catch is possible to check the location exists ?).

No need to go further if we can't find inno

Doublonmousse: Ok, running it to the end here : https://github.com/Doublonmousse/rnote/actions/runs/15084819604/job/42406028327

I'll probably have to force push once here to cleanup the history a bit

Kneemund: Alright, nice! One last (purely visual) thing: `-h`/`--silent` disables the progress visualization, maybe we should add that? I have no idea if that makes `--disable-interactivity` redundant, but it can't hurt to keep it.

Doublonmousse: Well, that didn't do it https://github.com/Doublonmousse/rnote/actions/runs/15085030326/job/42406506773

Doublonmousse: Given https://github.com/microsoft/winget-cli/issues/3494 I'm not sure it's possible through the cli.

And changing the settings seems to be only done through a json file https://github.com/microsoft/winget-cli/blob/master/doc/Settings.md (`winget settings` open a json editor, there's no setter from the cli afaik ?).

So I'm not sure I want to spend too much time on this part.

I've forced pushed to cleanup the history as well

Kneemund: Yeah, that's not really worth the effort then. Especially given that the logs are still readable.

Doublonmousse:
From https://github.com/flxzt/rnote/issues/1436#issuecomment-2896985103
> Note: the provided installer was missing the libraries libcrypto-3-x64.dll and libssl-3-x64.dll which I simply copied and pasted from the latest release to make it work

so it may not be working for windows 10

Doublonmousse:
Running a new CI run here : https://github.com/Doublonmousse/rnote/actions/runs/15354717368
Locally this works (no missing dlls) but I might have to add a
```
pacman -S mingw-w64-x86_64-openssl
```
if this doesn't pass on CI.

flxzt:
looks good, thank you!

Issue #1359, opened 2025-02-16
Default save location set to unwriteable `/` and no error is given when trying to save


elouin:
**Describe the bug**  
The default save location is set to `/` and since the user doesn't have rights to write there, the file isn't saved and there is no note about it. This happens for example, when you haven't saved the document yet and are closing rnote. Lost a few documents this way already. I know that for some people it sets the home directory as the default save location.

**To Reproduce**  
1. Open rnote
2. make changes
3. close the window without saving
4. click on save on the dialog

**Expected behavior**  
I expect the Dialog to either give a writeable path or to at least error and give the option to change the path, when trying to save to an unwriteable location.

**Console Output**  
```
~> flatpak --env=RUST_LOG=DEBUG --env=RUST_BACKTRACE=1 run com.github.flxzt.rnote
2025-02-16T11:51:31.159936Z DEBUG rnote: .. tracing subscriber initialized.
2025-02-16T11:51:31.714604Z DEBUG selectors::matching: Matching complex selector :not(:root)svg for svg id=None    
2025-02-16T11:51:31.714626Z DEBUG selectors::matching: Matching complex selector :root for svg id=None    
2025-02-16T11:51:31.714632Z DEBUG selectors::matching: Matching complex selector image for svg id=None    
2025-02-16T11:51:31.714636Z DEBUG selectors::matching: Matching complex selector marker for svg id=None    
2025-02-16T11:51:31.714640Z DEBUG selectors::matching: Matching complex selector pattern for svg id=None    
2025-02-16T11:51:31.714644Z DEBUG selectors::matching: Matching complex selector symbol for svg id=None    
2025-02-16T11:51:31.714647Z DEBUG selectors::matching: Matching complex selector defs for svg id=None    
2025-02-16T11:51:31.714651Z DEBUG selectors::matching: Matching complex selector clipPath for svg id=None    
2025-02-16T11:51:31.714655Z DEBUG selectors::matching: Matching complex selector mask for svg id=None    
2025-02-16T11:51:31.714664Z DEBUG selectors::matching: Matching complex selector marker for svg id=None    
2025-02-16T11:51:31.714669Z DEBUG selectors::matching: Matching complex selector desc for svg id=None    
2025-02-16T11:51:31.714677Z DEBUG selectors::matching: Matching complex selector title for svg id=None    
2025-02-16T11:51:31.714687Z DEBUG selectors::matching: Matching complex selector metadata for svg id=None    
[...]

PR #1444, opened 2025-05-08
Fixes: prevent walk_up_dir from going to root and set save in progress to false

Doublonmousse:
A couple of small fixes while looking at issue #1359


flxzt:
thanks!

PR #1448, opened 2025-05-11
updating numeric-sort to fix filename sorting

### LINKED ISSUE #1430: Add sorting of files based on name (like before)
If it is UI/UX related, don't post how a single aspect of the UI should be changed, rather in which use-case you felt
the UI is lacking or should be improved.

In the past I was able to sort my files based on the filename starting with yy-mm-dd but since the latest update the files are not any more sorted in the right way.

Settings to change the sorting

**Additional context**  

Thanks for the great software

Nugfle:
closes #1430


flxzt:
Thanks! The `Cargo.lock` file is checked in and needs to be updated as well

Nugfle:
done

Issue #1438, opened 2025-05-06
Lagging with OBS Studio recording


Weierstton:
**Describe the bug**  
Hello, I'm very happy to use Rnote, but there's a problem with OBS recording.  At the beginning everything's fine, but after 6 or 7 mins it starts to lag, not heavily, but couldn't write fast normally.

**Screen Recording*  
Here's what I've recorded, note the lag while writing the word "Canonical".
![Image](https://github.com/user-attachments/assets/632afba1-21b9-412d-a6ff-61ed1326652a)

**Desktop (please complete the following information):**  
 - OS: Windows 10 22H2
 - App Version: Rnote v0.12.0
 - Installation Source: Install using Winget
 - Processor: 11th Gen Intel(R) Core(TM) i7-11800H @ 2.30GHz   2.30 GHz
 - GPU: NVIDIA GeForce RTX 3050 Ti Laptop GPU
 - Installed RAM:  16.0 GB (15.8 GB usable)
 - Input Source: External Graphics Tablet - Huion
 
**OBS recording settings**  
  - Source Type: Window Capture 
  - Quality: High, bitrate of 2300
  - Format: MPEG-4(.mp4)
  - Video Encoder: Hardware(QSV, H.264)
  - Audio Encoder: AAC

And I've noticed there's no lag without recording, so I'm wondering if it's the problem with my OBS settings. 

Thanks a lot.

Intranox:
> **Describe the bug** Hello, I'm very happy to use Rnote, but there's a problem with OBS recording. At the beginning everything's fine, but after 6 or 7 mins it starts to lag, not heavily, but couldn't write fast normally.

I think it could be related to the problem in #1436. You can also try to see if as the writings to the file increase (possibly more pages) you notice rendering problems. The problem seems to occur if the document starts to be large and in the case of CPUs that are not too high performing or that are more stressed (e.g. with screen recording)

Weierstton:
> > **Describe the bug** Hello, I'm very happy to use Rnote, but there's a problem with OBS recording. At the beginning everything's fine, but after 6 or 7 mins it starts to lag, not heavily, but couldn't write fast normally.
> 
> I think it could be related to the problem in [#1436](https://github.com/flxzt/rnote/issues/1436). You can also try to see if as the writings to the file increase (possibly more pages) you notice rendering problems. The problem seems to occur if the document starts to be large and in the case of CPUs that are not too high performing or that are more stressed (e.g. with screen recording)

[...]

PR #1462, opened 2025-06-08
Performance fixes for larger files

### LINKED ISSUE #1436: Performance degrades on larger files

Doublonmousse:
Fixes #1436 
(and maybe #1438, waiting on a response)

- use `push_repeat` for background nodes : This seems to reduce RAM usage and buffer uploads to the gpu so there's more headroom for the rest of the nodes
- use the rtree to speed up `regenerate_rendering_in_viewport_threaded` : We use the rtree to get the strokes in the viewport without recalculating any stroke bounds, and get the ones not in the view by iterating on keys and using a hashmap to filter out the previously selected ones

On that subject, it seems we could use a https://docs.rs/slotmap/1.0.7/slotmap/sparse_secondary/struct.SparseSecondaryMap.html for the render components and/or be smarter on clearing the `render_comp` components of strokes (we know there's no work to do if we have x strokes in view that we just initialized the rendering components and x strokes in the rendering component slot)

I've updated the rtree to only keep bounds for visible strokes (so trashed strokes are removed from the tree)

- [use rtree for update_content_rendering_current_viewport calculations](https://github.com/flxzt/rnote/commit/e75519864ef25c03b0560c06f6044736ccae817e) Once again, using the rtree to calculate document bounds is much faster

I've also added to the visual debug mode a way to see the envelope of the whole rtree (marked by orange bounds)

### DIFF
--- a/crates/rnote-compose/src/ext.rs
+++ b/crates/rnote-compose/src/ext.rs
@@ -151,8 +151,6 @@ where
     /// splits a aabb into multiple which have a maximum of the given size. Their union is the given aabb.
     /// the split bounds are exactly fitted to not overlap, or extend the given bounds
     fn split(self, split_size: Vector2) -> Vec<Self>;
-    /// get the top patch for a given split size
-    fn get_origin(self, split_size: Vector2, split_order: SplitOrder) -> Self;
     /// splits a aabb into multiple of the given size. Their union contains the given aabb.
     /// The boxes on the edges most likely extend beyond the given aabb.
     fn split_extended(self, split_size: Vector2) -> Vec<Self>;
@@ -164,6 +162,9 @@ where
         split_size: Vector2,
         split_order: SplitOrder,
     ) -> Vec<Self>;
+    /// Gives an aabb of size split_size, snapped to a multiple of split_size
+    /// This returns the first element of a call to `split_extended_origin_aligned`
+    fn split_first_origin_aligned(self, split_size: Vector2) -> Self;
     /// Converts a Aabb to a kurbo Rectangle
     fn to_kurbo_rect(&self) -> kurbo::Rect;
     /// Converts a kurbo Rectangle to Aabb
@@ -362,22 +363,10 @@ impl AabbExt for Aabb {
         split_aabbs
     }
 
-    fn get_origin(self, split_size: Vector2, split_order: SplitOrder) -> Self {
-        let (outer_idx, inner_idx) = match split_order {
-            SplitOrder::RowMajor => (1, 0),
-            SplitOrder::ColumnMajor => (0, 1),
-        };
-
-        let offset_outer =
-            (self.mins[outer_idx] / split_size[outer_idx]).floor() * split_size[outer_idx];
-
-        let offset_inner =
-            (self.mins[inner_idx] / split_size[inner_idx]).floor() * split_size[inner_idx];
-
-        let mins = match split_order {
-            SplitOrder::RowMajor => Vector2::new(offset_inner, offset_outer),
-            SplitOrder::ColumnMajor => Vector2::new(offset_outer, offset_inner),
-        };
+    fn split_first_origin_aligned(self, split_size: Vector2) -> Self {
+        let offset_outer = (self.mins[1] / split_size[1]).floor() * split_size[1];
+        let offset_inner = (self.mins[0] / split_size[0]).floor() * split_size[0];
+        let mins = Vector2::new(offset_inner, offset_outer);
         Aabb::new(mins, mins + split_size)
     }
 
--- a/crates/rnote-compose/src/ext.rs
+++ b/crates/rnote-compose/src/ext.rs
@@ -151,6 +151,8 @@ where
     /// splits a aabb into multiple which have a maximum of the given size. Their union is the given aabb.
     /// the split bounds are exactly fitted to not overlap, or extend the given bounds
     fn split(self, split_size: Vector2) -> Vec<Self>;
+    /// get the top patch for a given split size
+    fn get_origin(self, split_size: Vector2, split_order: SplitOrder) -> Self;
     /// splits a aabb into multiple of the given size. Their union contains the given aabb.
     /// The boxes on the edges most likely extend beyond the given aabb.
     fn split_extended(self, split_size: Vector2) -> Vec<Self>;
@@ -358,6 +360,25 @@ impl AabbExt for Aabb {
         }
 
         split_aabbs
+    }
+
+    fn get_origin(self, split_size: Vector2, split_order: SplitOrder) -> Self {
+        let (outer_idx, inner_idx) = match split_order {
+            SplitOrder::RowMajor => (1, 0),
+            SplitOrder::ColumnMajor => (0, 1),
+        };
+
+        let offset_outer =
+            (self.mins[outer_idx] / split_size[outer_idx]).floor() * split_size[outer_idx];
+
+        let offset_inner =
+            (self.mins[inner_idx] / split_size[inner_idx]).floor() * split_size[inner_idx];
+
+        let mins = match split_order {
+            SplitOrder::RowMajor => Vector2::new(offset_inner, offset_outer),
+            SplitOrder::ColumnMajor => Vector2::new(offset_outer, offset_inner),
+        };
+        Aabb::new(mins, mins + split_size)
     }
 
     fn split_extended_origin_aligned(
--- a/crates/rnote-engine/src/engine/mod.rs
+++ b/crates/rnote-engine/src/engine/mod.rs
@@ -207,7 +207,8 @@ pub struct Engine {
     background_tile_image: Option<Image>,
     #[cfg(feature = "ui")]
     #[serde(skip)]
-    background_rendernodes: Vec<gtk4::gsk::RenderNode>,
+    background_rendernode: Option<gtk4::gsk::RenderNode>,
+    origin_background_rendernode: Option<Aabb>,
     // Origin indicator rendering
     #[serde(skip)]
     origin_indicator_image: Option<Image>,
@@ -234,7 +235,8 @@ impl Default for Engine {
             tasks_rx: Some(EngineTaskReceiver(tasks_rx)),
             background_tile_image: None,
             #[cfg(feature = "ui")]
-            background_rendernodes: Vec::default(),
+            background_rendernode: None,
+            origin_background_rendernode: None,
             origin_indicator_image: None,
             #[cfg(feature = "ui")]
             origin_indicator_rendernode: None,
--- a/crates/rnote-engine/src/engine/mod.rs
+++ b/crates/rnote-engine/src/engine/mod.rs
@@ -208,7 +208,6 @@ pub struct Engine {
     #[cfg(feature = "ui")]
     #[serde(skip)]
     background_rendernode: Option<gtk4::gsk::RenderNode>,
-    origin_background_rendernode: Option<Aabb>,
     // Origin indicator rendering
     #[serde(skip)]
     origin_indicator_image: Option<Image>,
@@ -236,7 +235,6 @@ impl Default for Engine {
             background_tile_image: None,
             #[cfg(feature = "ui")]
             background_rendernode: None,
-            origin_background_rendernode: None,
             origin_indicator_image: None,
             #[cfg(feature = "ui")]
             origin_indicator_rendernode: None,
--- a/crates/rnote-engine/src/engine/rendering.rs
+++ b/crates/rnote-engine/src/engine/rendering.rs
@@ -47,7 +47,6 @@ impl Engine {
                     )
                     .upcast(),
                 );
-                self.origin_background_rendernode = Some(origin_aabb);
             }
         }
 
@@ -116,7 +115,6 @@ impl Engine {
         #[cfg(feature = "ui")]
         {
             self.background_rendernode = None;
-            self.origin_background_rendernode = None;
             self.origin_indicator_rendernode.take();
         }
         widget_flags.redraw = true;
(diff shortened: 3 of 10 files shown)


crates/rnote-compose/src/ext.rs
```diff
@@ -161,6 +161,8 @@ where
     /// splits a aabb into multiple which have a maximum of the given size. Their union is the given aabb.
     /// the split bounds are exactly fitted to not overlap, or extend the given bounds
     fn split(self, split_size: na::Vector2<f64>) -> Vec<Self>;
+    /// get the top patch for a given split size
+    fn get_origin(self, split_size: na::Vector2<f64>, split_order: SplitOrder) -> Self;
```
flxzt: Improved docs and possibly renaming this method would be nice to clarify what it actually does.

crates/rnote-engine/src/store/keytree.rs
```diff
@@ -69,6 +70,18 @@ impl KeyTree {
     pub(crate) fn clear(&mut self) {
         *self = Self::default()
     }
+
+    pub(crate) fn get_tree(&self) -> &rstar::RTree<KeyTreeObject, rstar::DefaultParams> {
```
flxzt: The idea was to encapsulate all needed functionality from the R-Tree. Better would be to add the methods operating on the R-Tree itself here instead of exposing it to the rest of the crate

crates/rnote-engine/src/store/keytree.rs
```diff
@@ -69,6 +70,18 @@ impl KeyTree {
...
+        &self.0
+    }
+
+    pub fn get_bounds(&self) -> AABB<[f64; 2]> {
```
flxzt: RTree's AABB type should not be exposed outside of the module and instead be converted to parry's Aaab

crates/rnote-engine/src/store/render_comp.rs
```diff
@@ -6,9 +6,12 @@ use crate::strokes::content::GeneratedContentImages;
 use crate::{Drawable, render};
 use p2d::bounding_volume::{Aabb, BoundingVolume};
 use rnote_compose::ext::AabbExt;
-use rnote_compose::shapes::Shapeable;
+use std::collections::HashMap;
 use tracing::error;
 
+#[cfg(feature = "ui")]
+use rnote_compose::shapes::Shapeable;
```
flxzt: Import dependencies in the guards themselves like it is done in other parts of the code (e.g. L606)

crates/rnote-engine/src/store/render_comp.rs
```diff
@@ -242,28 +245,39 @@ impl StrokeStore {
...
+        // we want to iterate on the keys that are in the viewport using the
+        // rtree but also get from this the keys that are not in here
+        // for that also create a slotmap of keys that are in the viewport
+        // so that we can iterate a second time on keys and filter on elements not in the slotmap
+        let mut hashmap_in_viewport: HashMap<StrokeKey, ()> = HashMap::new();
+
+        let keys_in_viewport = self
+            .key_tree
+            .get_tree()
+            .locate_in_envelope_intersecting(&rstar::AABB::from_corners(
```
flxzt: use method `keys_intersecting_bounds` defined in keytree:L38 instead of operating on the raw R-Tree here.

crates/rnote-engine/src/store/render_comp.rs
```diff
@@ -242,28 +245,39 @@ impl StrokeStore {
...
+        // for now we are using directly the tree because we want to iter without actually
+        // collecting elements
+        let viewport_extended =
+            viewport.extend_by(viewport.extents() * render::VIEWPORT_EXTENTS_MARGIN_FACTOR);
+
+        // we want to iterate on the keys that are in the viewport using the
+        // rtree but also get from this the keys that are not in here
+        // for that also create a slotmap of keys that are in the viewport
+        // so that we can iterate a second time on keys and filter on elements not in the slotmap
+        let mut hashmap_in_viewport: HashMap<StrokeKey, ()> = HashMap::new();
```
flxzt: instead of collecting the keys as a side-effect of the `.map()` adapter, consider collecting into a `HashSet`. This way we can still only iterate once over all keys of `render_components` and check whether the key is in this hashset before calculating the bounds.

crates/rnote-engine/src/store/render_comp.rs
```diff
@@ -656,6 +686,18 @@ impl StrokeStore {
...
         }
 
+        // draw the rtree root
+        let tree_bounds = self.key_tree.get_bounds();
+        visual_debug::draw_bounds_to_gtk_snapshot(
+            Aabb::new(
+                na::point![tree_bounds.lower()[0], tree_bounds.lower()[1]],
+                na::point![tree_bounds.upper()[0], tree_bounds.upper()[1]],
+            ),
+            rnote_compose::Color::new(1.0, 0.5, 0., 1.0),
```
flxzt: don't hardcode the color, instead define a color on module level.

crates/rnote-engine/src/store/mod.rs
```diff
@@ -10,6 +10,7 @@ pub mod trash_comp;
 pub use chrono_comp::ChronoComponent;
 use keytree::KeyTree;
 pub use render_comp::RenderComponent;
+use rstar::AABB;
```
flxzt: this shouldn't be needed if it's use is encapsulated in `rtree`, but imports should be under `// Imports`

crates/rnote-engine/src/engine/rendering.rs
```diff
@@ -35,21 +34,20 @@ impl Engine {
                     }
                 };
 
-                for split_bounds in viewport.split_extended_origin_aligned(
+                let origin_aabb = viewport.get_origin(
                     self.document.config.background.tile_size(),
                     SplitOrder::default(),
```
flxzt: What's the split order used for here? I don't understand why it is an argument in the first place.

crates/rnote-engine/src/engine/rendering.rs
```diff
@@ -251,17 +250,23 @@ impl Engine {
...
 
-        for r in self.background_rendernodes.iter() {
-            snapshot.append_node(r);
+        if let (Some(bounds), Some(render_node)) = (
+            self.origin_background_rendernode,
+            self.background_rendernode.clone(),
+        ) {
+            snapshot.push_repeat(
+                &graphene::Rect::from_p2d_aabb(doc_bounds),
+                Some(&graphene::Rect::from_p2d_aabb(bounds)),
```
flxzt: it seems to me that we can get by by passing `None` for the child bounds argument here, because then the entire background tile rendernode size is used

crates/rnote-engine/src/engine/mod.rs
```diff
@@ -199,7 +199,8 @@ pub struct Engine {
     background_tile_image: Option<render::Image>,
     #[cfg(feature = "ui")]
     #[serde(skip)]
-    background_rendernodes: Vec<gtk4::gsk::RenderNode>,
+    background_rendernode: Option<gtk4::gsk::RenderNode>,
+    origin_background_rendernode: Option<Aabb>,
```
flxzt: *If* this is still needed, this should be renamed and guarded by the ui feature.

flxzt (commented):
when this lands these are nice improvements to the rendering part of the engine. I added some comments

flxzt:
Only storing non-trashed strokes in the R-Tree is problematic and will require additional changes in other parts - essentially everywhere where one of the methods of `KeyTree` are used (like `keys_in_bounds()`, `keys_intersecting_bounds()` etc. Right now they are all used under the assumption that **all** strokes are included in the tree. Did you consider this?

crates/rnote-engine/src/store/stroke_comp.rs
```diff
@@ -127,17 +127,12 @@ impl StrokeStore {
...
-            .clone()
-            .fold(0.0, |acc, stroke| stroke.bounds().mins[1].min(acc));
-        let strokes_max_y = strokes_iter.fold(0.0, |acc, stroke| stroke.bounds().maxs[1].max(acc));
-
-        strokes_max_y - strokes_min_y
+        if self.keytree_is_empty() {
+            return 0.0;
+        } else {
+            let bounds = self.key_tree.get_bounds();
+            bounds.upper()[1] - bounds.lower()[1]
```
Doublonmousse: Having the keytree only hold the non trashed keys makes it possible to get the bounds in constant time here.

I'm currently not doing that  (only holding non trashed keys in the rtree) so some behavior is modified but I'm unclear of the consequences as of now. The only part I can find is that the vertical spacer tool will not move thrashed keys anymore.

So I'm not sure what's the best way forward.

crates/rnote-engine/src/store/render_comp.rs
Doublonmousse: I can't do that directly as the method has already collected into a vec (and doing an into_iter didn't really work because of type signatures).
The next best thing is a `keys_intersecting_bounds_hash` variant that collects into a HashMap

crates/rnote-engine/src/store/render_comp.rs
```diff
@@ -242,28 +241,46 @@ impl StrokeStore {
...
+        // we want to iterate on the keys that are in the viewport using the
+        // rtree but also get from this the keys that are not in here
+        // for that also create a slotmap of keys that are in the viewport
+        // so that we can iterate a second time on keys and filter on elements not in the slotmap
+        let mut keys_in_viewport_hash = self
+            .key_tree
+            .keys_intersecting_bounds_hashset(viewport_extended);
+
+        // if we have a current key in the store, push it as well
+        if let Some(current_stroke_key) = self.current_key() {
```
Doublonmousse: Ah, this is a mistake from me trying to force adding a current key (so that a stroke in progress is always rendered even if the start position is not on the screen anymore), this fails as I've not pushed the rest of the modifications that goes with this

Doublonmousse: removed for now

crates/rnote-engine/src/store/stroke_comp.rs
Doublonmousse: I think I'll try to do two keytrees : for trashed and non trashed keys.
The cost of filtering after the fact is too great for this function (as this kills the optimization)

Doublonmousse: The two keytrees are implemented in https://github.com/flxzt/rnote/pull/1462/commits/5aee84d644f114053d01b8ad27e9ed99967990ff

Doublonmousse:
I'm trying to rely more on the keytree to get stroke bounds whenever possible (as calculating a bunch of stroke bounds on each event is potentially pretty expensive).

Testing the two keytree solution, I'm encountering quite weird results with the splitting eraser

Intranox:
I'm not entirely clear on the issues arising from this pull request. I've been using a version containing these fixes for some time now, and in my daily use, I haven't encountered any problems. In fact, I can say I've rediscovered a software that finally meets my needs. I take notes that can span several pages, and as the content increases, it becomes completely impossible to continue on the same paper. However, I think it's very important for an app like this to be able to take as long notes as I want. With these fixes, this has been possible, and I've also rediscovered a general fluidity that wasn't there before (even in scrolling or the selection tool, for example).

flxzt:
good to know, but not really relevant for the technical discussion this PR should focus on @Intranox

Doublonmousse:
Rebased on main.

New thing: opening the visual debug panel triggers the following message: 
```bash
*** BUG ***
In pixman_region32_init_rect: Invalid rectangle passed
Set a breakpoint on '_pixman_log_error' to debug
```
Seems to be caused by the rectangle holding the statistics. Not sure if this is from this PR or not (maybe since a specific gtk version? Found https://github.com/fpaquet/gtksheet/pull/21/files and https://gitlab.gnome.org/GNOME/gtk/-/issues/6338)
There's probably more performance stuff to do but I'll leave it for other potential PRs.

crates/rnote-engine/src/store/render_comp.rs
```diff
@@ -570,9 +581,28 @@ impl StrokeStore {
...
+
+        // draw the trashed rtree root
+        let trashed_tree_bounds = self.trashed_key_tree.get_bounds();
+        visual_debug::draw_bounds_to_gtk_snapshot(
+            trashed_tree_bounds,
+            visual_debug::COLOR_TRASH_RTREE_BOUNDS,
+            snapshot,
+            border_widths,
+        );
+
```
Doublonmousse: Turns out that drawing the two rtree borders creates absolute pandemonium in the gtk4 renderer for _whatever_ reason.

- It creates semi-permanent flickering zones with opacity changes from previous snapshot accumulating instead of resetting if I draw them _after_ the for loop on the keys, but not If I moved them before (which I did just here).

- It does not work for fixed pages AT ALL. It flickers constantly whatever the os. This is the reason for the `pixman_region32_init_rect` warning messages on windows (but only on windows for some reason, maybe some scaling?). On linux you get this abomination: 


Screencast_20260128_182427

I don't want to deal with gtk4 madness (for now) here so I'm just gonna remove the draws and maybe at some point try to report upstream (though getting a minimum example working won't be super simple so until then ...)

Doublonmousse:
Rebased onto 0.14.2

Doublonmousse:
Rebased after #1776

PR #1472, opened 2025-07-01
PT and PT_BR translations update

SuperAmoeba:
I am Portuguese and I fixed some mistakes in the PT translation aswell as completely finished it. In PT_BR there were a few typos, which I corrected. I also spoke to a Brazilian friend to add a few things that were missing. I didn't look through all of PT_BR, though, because I don't come from Brazil, but I do recognize if there a typos, because it is the largely the same language as Portuguese from Portugal.

TLDR:
Fixed and finished PT
Fixed a couple of typos and missing translations in PT_BR


flxzt:
thanks!

PR #1480, opened 2025-07-19
fix: prevent data race in xopp import code

### LINKED ISSUE #1478: importing a xopp file silently fails
**Describe the bug**  
importing a xopp file silently fails

**Expected behavior**  
Should work

 - OS: Fedora 42
 - App Version:Rnote v0.13.0 nightly (this works on 0.12)
 - Installation Source: Flatpak nightly + compiled from meson
 - Desktop Environment: kde
 - Display Server:  Wayland
 - Input Source: [e.g. Mouse and Keyboard, On-Screen-Tablet, External Graphics Tablet, ...]

**Additional context**  

This feels very similar to #1227 but I've yet to find a way to fix this by doing a change of weak/strong reference here.

The only reliable way I found is to comment this line
https://github.com/flxzt/rnote/blob/81aea502ae55e01b6f4a1a60d466adb0fe9cc71c/crates/rnote-ui/src/dialogs/import.rs#L575

For context
https://github.com/flxzt/rnote/blob/81aea502ae55e01b6f4a1a60d466adb0fe9cc71c/crates/rnote-ui/src/dialogs/import.rs#L574-L579

Now doing this also means that you _then_ cancel the dialog and it comes with the following error
```
   7.659640728s ERROR rnote::dialogs::import: XOPP import dialog closed, but failed to send signal through channel. Err: TrySendError { kind: Disconnected }
```

So it IS connected to the dialog closing, but the file is (seemingly) not the culprit (or maybe not the only one ?). Maybe the `inner_tx_confirm` ?

Doublonmousse:
Fixes data race bugs preventing imports.

Fixes #1478.

The way I've done it is the following
- The two buttons + dialog close only take a `tx` and send back a `false/true` meaning a cancel/confirm of the import, and a second bool to say if the dialog is closed or not.
  - This way a single message is sent for each action with no weird loop (dialog.close -> emit_click -> dialog.close ...)
  - The responsability to close the dialog is external to the callbacks (it is done externally, checking first that we don't get a message from the close event itself not to close the dialog twice)

We can thus wait for the next message from the `rx` and deal with it (close the dialog and launch or not the import).

The import is thus done later if needed in another (tx,rx) communication line, so that we can wait on the import to complete to get out of the async function.

No more data race on whether or not the dialog -> cancel_button emit -> tx emits first a message vs the import itself. 
The bug is whether or not this is the first message + we'd get through the `rx.next().await` before the thread is launched or not ...

### DIFF
--- a/crates/rnote-ui/src/dialogs/import.rs
+++ b/crates/rnote-ui/src/dialogs/import.rs
@@ -553,66 +553,95 @@ pub(crate) async fn dialog_import_xopp_w_prefs(
 
     // Listen to responses
 
-    let (tx, mut rx) = futures::channel::mpsc::unbounded::<anyhow::Result<bool>>();
+    let (tx, mut rx_confirm) = futures::channel::mpsc::unbounded::<(bool, bool)>();
     let tx_cancel = tx.clone();
     let tx_confirm = tx.clone();
+    let tx_close = tx.clone();
 
-    import_xopp_button_cancel.connect_clicked(clone!(
-        #[weak]
-        dialog,
-        move |_| {
-            dialog.close();
-
-            if let Err(e) = tx_cancel.unbounded_send(Ok(false)) {
-                error!(
-                "XOPP import dialog closed, but failed to send signal through channel. Err: {e:?}"
+    import_xopp_button_cancel.connect_clicked(clone!(move |_| {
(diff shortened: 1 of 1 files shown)


flxzt:
thank you!

PR #1481, opened 2025-07-19
Xopp text import

Doublonmousse:
Text import from xournal files.

Now the actual text is a little bit larger in rnote. Should I increase the font weight to compensate ? Or maybe it is spaces that are larger ?

image

Also added some notes for the export.

From #1320 there's a way to retain shapes by converting them to brushstrokes, but under some conditions. Maybe I can add a `can_convert_to_brush` method to test these conditions and create a common code path for both conversions ?

### DIFF
--- a/crates/rnote-engine/src/engine/snapshot.rs
+++ b/crates/rnote-engine/src/engine/snapshot.rs
@@ -183,6 +183,20 @@ impl EngineSnapshot {
                                 }
                             }
                         }
+
+                        for new_xopptext in layers.texts.into_iter() {
+                            match Stroke::from_xopptext(new_xopptext, offset, xopp_import_prefs.dpi)
+                            {
+                                Ok(new_text) => {
+                                    engine.store.insert_stroke(new_text, None);
+                                }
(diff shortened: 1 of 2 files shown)


flxzt:
thanks, looks good!

Maybe the weight does not correspond 100% to the weight in Xournal++? It seems to be a bit larger/heavier, but this could also be just differences how fonts are rendered between Rnote and Xournal++.

Is this happening for a variable or fixed weight font?

Doublonmousse:
For now testing on the most default settings. In xournal++ Sans regular 12.

Testing with source code pro (regular on xournal++ side), the difference is less so (in the green rectangle)

But there's quite a big difference if spaces are involved.
image

flxzt:
let's just go with it, right?
If it's ready from your side, feel free to merge @Doublonmousse (with a commit message following [conventional commits](https://www.conventionalcommits.org/en/v1.0.0/#summary)) :)

Doublonmousse:
> let's just go with it, right?

Yeah, will open an issue to keep tabs on the remaining issue later

Issue #1155, opened 2024-07-25
Implement the ability to flip the current selection


parmjotsinghrobot:
**Please submit only one feature request in one issue!**  
If it is UI/UX related, don't post how a single aspect of the UI should be changed, rather in which use-case you felt
the UI is lacking or should be improved.

**Is your feature request related to a problem? Please describe.**  
In many photo editors, such as GIMP and the like, they have the ability to flip the current selection, albeit in a submenu. I find this ability in Rnote is missing, which I didn't expect and was kinda hurt by it as well, since I was working on a detailed (for me) joint. I was hoping that I could just flip it, so I could save some time, but I can't. I've attached a picture below if you're curious.

**Describe the solution you'd like**  
Add 2 buttons next to where the copy and paste button is to flip the current selection vertically and horizontally.

**Describe alternatives you've considered**  
If this doesn't fit, maybe it could be put into the tools area, but I don't really think it belongs there.

**Additional context**  
_
![20240725-12_45_43_grim](https://github.com/user-attachments/assets/1238485e-24b5-4138-a7c2-c8c3c9deb24a)
_

Doublonmousse:
I think that's a duplicate of #207. I think the negative resizing part is a good idea. Though maybe a dedicated button would be there (as this might not be super "discoverable")

parmjotsinghrobot:
Yea I agree it's a duplicate, but it's an enhancement for sure. Should we close as duplicate and bump that with the negative resizing bit?

10/11/2024 06:15:51 Doublonmousse ***@***.***>:

PR #1487, opened 2025-08-03
Feature: Mirror Selection Tool

### LINKED ISSUE #207: Mirror selection

podzolelements:
Fix #207 by adding a horizontal and vertical mirroring tool that operates on the actively selected elements

### Basic required functionality
- [x] BrushStrokes
- [x] ShapeStrokes
- [x] Implement vertical mirroring
- [x] Make proper icons
- [x] Compliance with code style and pass tests

### Planned: Mirroring images
- [x] BitmapImages
- [x] VectorImages

### Demo

https://github.com/user-attachments/assets/5ca87ece-0094-4809-97b5-9b31906bb793

https://github.com/user-attachments/assets/2e28df5e-6e41-4b04-9cb5-2fb87dde0cf4

### Not currently planned
- TextStrokes. Mirroring text seems complicated and not all that useful. If mirrored text is required, it can be pasted in and be reverse justified through the typewriter already. Perhaps this could be a "macro" that does those operations on the selected text? I might think about this more once I have more familiarity with the codebase, but only after the rest of the functionality is implemented
- ~~Altering the `Transformable` trait to include a mirror function as a basic geometric transformation. I think this would be the best way to scale the mirroring functionality to new features, however it would be a tremendous undertaking and as of writing I'm not confident I could pull it off. Again, I want to focus on getting the feature implemented and working~~
- "Negative resizing" as described in #1155. Mirroring will solve the root cause of not being able to flip the selection as mentioned in the issue, but I think negative resizing is a different enough feature to warrant a separate PR. This is sort of what I first considered doing, using only `Transformable` functions, but `scale()` would instantly crash if the x and y scalings had different signs, so I moved on from that idea

### Feedback
Some things I'm not sure about yet and want to discuss:
- Location of mirror buttons. I've currently placed it in the selection sidebar, however, #614 has a suggestion to place the mirror tool onto the selection area itself. I personally think it would make the selection area more cluttered, especially with small selections
- Are both a horizontal and vertical mirror tool necessary? I think so, even though geometrically a vertical mirror can be done via rotate->mirror->rotate, as that isn't super intuitive if you're not already familiar with it
- Other ideas and thoughts are welcome for discussion


Doublonmousse:
## Not currently planned

> TextStrokes. Mirroring text seems complicated and not all that useful. If mirrored text is required, it can be pasted in and be reverse justified through the typewriter already. Perhaps this could be a "macro" that does those operations on the selected text? I might think about this more once I have more familiarity with the codebase, but only after the rest of the functionality is implemented

There might be some jankiness because of this. I think it makes sense not to have it for textstrokes but maybe we need to adapt some things here.
For example if your selection contains text and other stroke elements, what should the mirror operation do ? Should we _visibly_ deactivate/activate the buttons as soon as a selection contains/does not contain text ? 

> Altering the Transformable trait to include a mirror function as a basic geometric transformation. I think this would be the best way to scale the mirroring functionality to new features, however it would be a tremendous undertaking and as of writing I'm not confident I could pull it off. Again, I want to focus on getting the feature implemented and working

I think that can be done to refactor/condense this down a little later.

> "Negative resizing" as described in https://github.com/flxzt/rnote/issues/1155. Mirroring will solve the root cause of not being able to flip the selection as mentioned in the issue, but I think negative resizing is a different enough feature to warrant a separate PR. This is sort of what I first considered doing, using only Transformable functions, but scale() would instantly crash if the x and y scalings had different signs, so I moved on from that idea

Do you have a crash message for this ? Does it crash on a specific stroke type ? If it's relatively easy to fix maybe it's worth trying (though I'm not 100 % sure all stroke types would magically work with this)

## Feedback

> Location of mirror buttons. I've currently placed it in the selection sidebar, however, https://github.com/flxzt/rnote/pull/614 has a suggestion to place the mirror tool onto the selection area itself. I personally think it would make the selection area more cluttered, especially with small selections

I agree with this. I think it makes more sense to do negative resizing than adding the mirror tool on the selection.

One additional thing is that the selector panel is starting to get a little crowded. Two more icons, and maybe an additional one down the line with #1428. One button I'm using a lot in the selector is the delete button and I'm not sure it won't get out of bounds with this. Maybe we're gonna have to reorder things a little bit, add separators, or submenus ?

crates/rnote-engine/src/store/stroke_comp.rs
```diff
@@ -298,6 +298,92 @@ impl StrokeStore {
         widget_flags
     }
 
+    // Mirror stroke horizontally for given set of keys
+    //
+    // The strokes need to update rendering after mirror
+    pub(crate) fn mirror_stroke_horizontal(&mut self, keys: &[StrokeKey]) -> WidgetFlags {
```
Doublonmousse: The two `mirror_stroke_*` functions are very similar.
Can we do a single `mirror_stroke(..., orientation)`
with orientation an additional argument (bool or enum with `horizontal/vertical` options) ?

crates/rnote-engine/src/strokes/stroke.rs
```diff
@@ -672,4 +672,228 @@ impl Stroke {
             }
         }
     }
+
+    // horizontally mirrors point around line 'x = selection_centerline_x'
+    fn mirror_point_x(point: &mut na::Vector2<f64>, selection_centerline_x: f64) {
```
Doublonmousse: Having this function be called as `Self::*` is a little weird.
Maybe it can be in a `point_utils.rs` file somewhere else ? (why not in `rnote-compose` as well)

crates/rnote-engine/src/strokes/stroke.rs
```diff
@@ -672,4 +672,228 @@ impl Stroke {
...
+        point.x -= selection_centerline_x;
+        point.x *= -1.0;
+        point.x += selection_centerline_x;
+    }
+
+    // performs a matrix transform on 2d affine matrix causing a reflection across line 'x = selection_centerline_x'
+    fn mirror_affine_x(affine: &mut na::Affine2<f64>, selection_centerline_x: f64) {
```
Doublonmousse: I think this could be changed by a
```rs
pub fn append_mirror_affine_x(&mut self, selection_centerline_x:f64) 
```
method inside of `crates/rnote-compose/src/transform/mod.rs` to replace calls like
```rs
Self::mirror_affine_x(
                    &mut vector_image.rectangle.transform.affine,
                    selection_centerline_x,
                );
```
to 
```rs
vector_image.rectangle.transform.append_mirror_affine_x(selection_centerline_x)
```

(that tends to move things closer to the `rnote-compose` crates/facilitates further refactoring in that direction because the next step would be here to add a `mirror` or similar method to the rectangle and call `rectangle.mirror` that would itself call `rectangle.tranform`. ...)

crates/rnote-engine/src/strokes/stroke.rs
```diff
@@ -672,4 +672,228 @@ impl Stroke {
...
+                let new_penpath_elements: Vec<Element> = current_penpath_elements
+                    .iter()
+                    .zip(coords.iter())
+                    .map(|(current_penpath_element, new_position)| Element {
+                        pos: *new_position,
+                        ..*current_penpath_element
+                    })
+                    .collect();
+
+                if let Some(new_penpath) = PenPath::try_from_elements(new_penpath_elements) {
```
Doublonmousse: You kinda hit the limits of what can be done outside the penpath methods outside of the `rnote-compose` crate (where there are in place methods to modify the `PenPath`).

If you add a method to the `PenPath` instead (close to the `mirror_point_x` if you move it near as well) you can modify in place without having to clone/destructure the position/mirror/restructure/reconstruct the path.

Though in order you'd need 
- the method on pos (the function as an utility)
- then the method on `Element` (that would act on the `pos` field of the `Element`)
- then the method on `PenPath` (which will act on the start `Element`)

Though you also need the pendant on the `Segment` elements. Here the conversion is actually lossy as the docstring indicates
```
/// extracts the elements from the path. the path shape will be lost, as only the actual input elements are returned.
```
Though that's more or less the same (ends as operations on pos)

crates/rnote-engine/src/strokes/stroke.rs
```diff
@@ -672,4 +672,228 @@ impl Stroke {
...
+                    brushstroke.path = new_penpath;
+                }
+            }
+            Stroke::ShapeStroke(shape_stroke) => match &mut shape_stroke.shape {
+                rnote_compose::Shape::Line(line) => {
+                    Self::mirror_point_x(&mut line.start, selection_centerline_x);
```
Doublonmousse: easier to do : add a method to `Line` that would call your newly placed `mirror_point_x` utility function from `rnote-compose`.

If adding a `mirror` function to the `Transform` trait is too much to do in one go, maybe it's easier to add the same method name to the different elements (and ultimately move these back to the Transform trait part one they are implemented for every `Transform` type)

podzolelements:
> For example if your selection contains text and other stroke elements, what should the mirror operation do ?

Currently it can make kind of a mess, which I think needs to be addressed, especially when you would be expecting the text to react to the mirror.

>Should we visibly deactivate/activate the buttons as soon as a selection contains/does not contain text ?

I think that might be the simplest option, which might be a little annoying from a user perspective, though probably less so than the current implementation of leaving the text completely alone. I've thought about it a bit more since then, and I think mirroring characters is a really bad idea. What are your thoughts on 'positionally mirroring' text, where the position gets updated with the rest of the selection, and something like the following happens:
text_reversed

Though the justification would probably still need to be reversed, as in the following:
rejustify

Not to mention for vertical mirrors the line order would need to be reversed:
text_order_reversal
I'd be down to explore this avenue if you think it is reasonable, and we could always fall back on disabling the buttons with text selections in them if it turns out to be a disaster/introduce a bunch of other issues (or come up with other ideas).

> One additional thing is that the selector panel is starting to get a little crowded.
> Maybe we're gonna have to reorder things a little bit, add separators, or submenus ?

Maybe the 4 selection types could be put into a selection menu like how the shape selector currently works? I personally find I don't change selection modes much of at all, but I don't know about other people's workflows. Also the 'lock aspect ratio' button could go in there as well, freeing another space, which would reduce the current items from 12 to 8, something like:
selection_menu_idea
Or it could be more like the brush configuration menu, and have the descriptions built in vs as mouseover text?
Perhaps the 'select all' and 'deselect all' strokes could be condensed into more of a single toggle button? Just throwing some ideas out there.

crates/rnote-compose/src/penpath/mod.rs
```diff
@@ -98,6 +98,14 @@ impl Transformable for PenPath {
             segment.scale(scale);
         });
     }
+
+    fn mirror(&mut self, centerline: f64, orientation: MirrorOrientation) {
+        self.start.mirror(centerline, orientation);
+
+        for element in &mut self.segments {
```
Doublonmousse: very nitpicky : you can also `iter_mut.for_each` like in the preceding `scale` function.

Doublonmousse:
Very good ! I've tested exporting as well just in case to all formats and this works too. 

- Should we add accelerations/keyboard shortcuts for the actions ? 
- Should we do something for text ? Special mode or deactivate the buttons/actions if text is involved ?  
- Do we start doing something for buttons now (or we keep it as a follow up ?).

I'm adding @flxzt to discuss the remaining questions.

crates/rnote-compose/src/transform/mod.rs
```diff
@@ -120,6 +133,33 @@ impl Transform {
         .unwrap();
     }
 
+    /// Apply a reflection across either Horizontal: 'x = centerline' or Vertical: 'y = centerline' to the
+    /// affine matrix based on the orientation
+    pub fn append_mirror_mut(&mut self, centerline: f64, orientation: MirrorOrientation) {
+        let mirror_transformation;
```
Doublonmousse: see https://github.com/flxzt/rnote/actions/runs/17169865520/job/48767009913?pr=1487#step:13:14

podzolelements:
> Should we add accelerations/keyboard shortcuts for the actions ?

Good idea, what do we think for keys? `h` and `v` would probably be best, though I personally would have them as `x` and `y`.

> Should we do something for text ? Special mode or deactivate the buttons/actions if text is involved ?

I'll experiment with text mirroring with the functionality as described in my recent reply.

> Do we start doing something for buttons now (or we keep it as a follow up ?).

I think it makes most sense to reorganize the selection menu as a separate PR, as that could include other fixes that significantly differ functionally from mirroring, most notably #155. I'd be interested in working on it right after this though.

crates/rnote-engine/src/store/stroke_comp.rs
```diff
@@ -298,6 +298,90 @@ impl StrokeStore {
...
+        keys: &[StrokeKey],
+        orientation: MirrorOrientation,
+    ) -> Option<WidgetFlags> {
+        let mut widget_flags = WidgetFlags::default();
+
+        if keys.is_empty() {
+            return Some(widget_flags);
+        }
+
+        let mut stroke_contains_text = false;
```
podzolelements: I'm in a bit over my head here, so if there is a better way of checking if the selection contains text I'm all ears

crates/rnote-engine/src/pens/selector/penevents.rs
```diff
@@ -685,6 +686,44 @@ impl Selector {
...
+                            self.mirror(engine_view, MirrorOrientation::Horizontal)
+                        {
+                            widget_flags = mirror_widget_flags;
+
+                            EventResult {
+                                handled: true,
+                                propagate: EventPropagation::Stop,
+                                progress: PenProgress::InProgress,
+                            }
+                        } else {
```
podzolelements: Currently, mirrors that contain text activated through keyboard shortcuts silently fail. I'm not sure how to work around this without access to the `RnAppWindow`.

crates/rnote-engine/src/store/stroke_comp.rs
Doublonmousse: Running clippy on this
```
warning: this `match` can be collapsed into the outer `if let`
   --> crates/rnote-engine/src/store/stroke_comp.rs:321:17
    |
321 | /                 match stroke {
322 | |                     Stroke::TextStroke(_text_stroke) => {
323 | |                         stroke_contains_text = true;
...   |
326 | |                 }
    | |_________________^
    |
help: the outer pattern can be modified to include the inner pattern
   --> crates/rnote-engine/src/store/stroke_comp.rs:317:25
    |
317 |             if let Some(stroke) = Arc::make_mut(&mut self.stroke_components)
    |                         ^^^^^^ replace this binding
...
322 |                     Stroke::TextStroke(_text_stroke) => {
    |                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ with this pattern
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#collapsible_match
    = note: `#[warn(clippy::collapsible_match)]` on by default
```
so you can do
```rs
            if let Some(Stroke::TextStroke(_text_stroke)) =
                Arc::make_mut(&mut self.stroke_components)
                    .get_mut(key)
                    .map(Arc::make_mut)
            {
                stroke_contains_text = true;
            }
```

Now I'd do something else on top of that and use `any` on the iterator and use a `matches!` macro instead of the if {} true
```rs
        let stroke_contains_text = keys.iter().any(|&key| {
            matches!(
                Arc::make_mut(&mut self.stroke_components)
                    .get_mut(key)
                    .map(Arc::make_mut),
                Some(Stroke::TextStroke(_))
            )
        });
```

The `any` is shortcircuiting (will stop at the first text element found), you don't mut the variable and `any` returns false for an empty selection so all is okay

crates/rnote-engine/src/pens/selector/penevents.rs
Doublonmousse: Yeah, you're not going to get access to the `RnAppWindow` inside of the `rnote-engine`. 

If it was a keyboard shortcut activated from an accelerator (see `set_accels_for_action` calls) then it'd go through the same code as clicking the buttons. But then you're not going to be able to add `x` as a shortcut, nor `ctrl+x` (which would even overwrite the `ctrl+x` shortcut when writing 

It's a little different from the rest of the keyboard shortcuts here (as there is no `ctrl` mask contrary to the rest. Though you can't do `ctrl+x` either).

Here the way to communicate to the engine would be through  the `WidgetFlags` (add a field, handle it in the `handle_widget_flags`). Does it make sense to modify the widget flags to accomodate this ? Maybe add a field for an error message ?

Another thing I checked on `WidgetFlags`, I'm not sure it matters much to do a `Option<WidgetFlags>`. It's kinda not used in the code, because handling a default widget flags will by default do nothing. Although maybe it's better in principle to pass around the default widget flags even if you know it does nothing _in case_ another step is added later on that modifies it. I don't feel too strongly about this though.

Though if you add a message field to the `WidgetFlags`, this will obviously change (generic popup in the handle function so the engine mirror functions would always return a widget flags)

podzolelements: > see `set_accels_for_action` calls

Ah, gotcha, thanks. I was sort of wondering as a user why everything had a `ctrl` prefix, but that makes sense architecturally. I'll make them as `ctrl+m` and `ctrl+shift+m` for now, until we can think of something intuitive that doesn't conflict.

> I'm not sure it matters much to do a `Option<WidgetFlags>`

The main reasoning here was as an easy way to indicate to the caller that the operation didn't do anything and should trigger the message. Though now you mention it, it makes much more sense to use the dedicated _flags_ system for passing around information.

crates/rnote-engine/src/store/stroke_comp.rs
podzolelements: Awesome, thanks for the info!

crates/rnote-engine/src/store/stroke_comp.rs
```diff
@@ -305,30 +305,26 @@ impl StrokeStore {
...
+                    .get_mut(key)
+                    .map(Arc::make_mut),
+                Some(Stroke::TextStroke(_))
+            )
         });
 
         if stroke_contains_text {
-            return None;
+            widget_flags.popup_message =
+                Some("Mirroring selections containing text is not supported".to_string());
```
Doublonmousse: the `gettext` call should go here

crates/rnote-ui/src/appwindow/mod.rs
```diff
@@ -335,6 +335,12 @@ impl RnAppWindow {
         if let Some(enable_text_preprocessing) = widget_flags.enable_text_preprocessing {
             canvas.set_text_preprocessing(enable_text_preprocessing);
         }
+        if let Some(popup_message) = widget_flags.popup_message {
+            self.overlays().dispatch_toast_text(
+                &gettext(popup_message),
```
Doublonmousse: Remove this `gettext`: we shouldn't provide a fixed translation for a variable string (depends on the widget flags !)

Doublonmousse (commented):
There's only a minor fix to do for gettext here. After the fix I'll run the CI one more time (to check/listen to clippy)

At some point I imagined toggling the button's sensitivity in the toolbar based on context (on each selection update) so you wouldn't be able to click as soon as a text element would be in the selection. Haven't totally though out the mechanism though ...

But I'm pretty satisfied with the popup method so all is good.

podzolelements:
> so you wouldn't be able to click as soon as a text element would be in the selection. Haven't totally though out the mechanism though ...

Yeah I wasn't sure how to do this on the UI side so I went with the popup.

podzolelements:
There seems to be some odd behavior when running the visual debug mode, the bounding boxes don't update their rendering properly. Affects both horizontal and vertical mirrors, and when the selection is dragged it updates properly. Not sure if there is something I'm missing in `WidgetFlags` or where to look if this is an issue with how the visual debug mode works.

odd_debug_mirror

Edit: Actually this seems to be consistent with how the visual debugger currently works. Dragging around and not releasing a selection has the same behavior, where the bounding boxes stay behind until the selection is released.

crates/rnote-engine/src/store/stroke_comp.rs
```diff
@@ -298,6 +299,87 @@ impl StrokeStore {
...
+        keys.iter().for_each(|&key| {
+            if let Some(stroke) = Arc::make_mut(&mut self.stroke_components)
+                .get_mut(key)
+                .map(Arc::make_mut)
+            {
+                stroke.mirror(selection_centerline, orientation);
+                self.set_rendering_dirty(key);
+            }
+        });
+
```
Doublonmousse, line 375:
```suggestion
self.update_geometry_for_strokes(keys);
```

Doublonmousse (commented):
The visual debug hitboxes only updating on a pen up event is an optimization : do not regenerate the hitboxes when we are only moving the strokes. As there are primarly used by the different selector to determine what strokes are selected, and we can't create a new selection in the middle of dragging an element, it makes sense to defer the computation to the pen up event.

Not updating the stroke geometry breaks the 4th `Select Intersecting Path` selection mode here
[Screencast_20250827_200034.webm](https://github.com/user-attachments/assets/e7a647d4-e8eb-4412-9fd6-5d8d5126d2db)

Cargo.lock
```diff
@@ -3587,6 +3587,7 @@ dependencies = [
  "flate2",
  "futures",
  "geo",
+ "gettext-rs",
```
Doublonmousse: Didn't think of this, ofc doing gettext on the rnote-engine caller adds the `gettext-rs` dependency to this, and that maybe isn't ideal. 

Should we not do a generic message and add a boolean flag for this ? Or an error enum on the calling side and gettext on each variant inside rnote-ui ?

Doublonmousse (approved):

flxzt:
Unfortunately there has been some changes in main and the branch needs to be rebased and adapted for those changes. Most notable a switch from `nalgebra` to `glam` for algebraic math.

Issue #1315, opened 2024-12-06
Bullet points(Lists)


Keschler:
**Is your feature request related to a problem? Please describe.**  
There is no way to create bullet points.

**Describe the solution you'd like**  
I want that there is a feature so that like in Onenote you can create bullet points that are also formated to the right.

**Describe alternatives you've considered**  
Just using tabs so that they are more to the right
![image](https://github.com/user-attachments/assets/328e00f6-2d64-426e-86ae-44f9bdda93f9)

**Additional context**  
Like this:
![image](https://github.com/user-attachments/assets/8be6b199-759f-4255-8f53-b341f6cc6722)

PR #1489, opened 2025-08-06
Add auto-closing pairs and list continuation to textStokes (Markdown-Editor-like behavior)

demhauke:
This PR introduces two main improvements to text editing behavior:
- Auto-closing pairs
When the user types an opening character such as (, [, {, ", or ', the corresponding closing character is automatically inserted.

- Automatic list continuation (#1315)
When the user presses Enter after a line starting with a bullet or indentation (e.g., -, *, or similar), the next line will automatically include the same prefix.

- before

https://github.com/user-attachments/assets/1d106077-c675-441d-aec9-cd93c15230e6

- after

https://github.com/user-attachments/assets/9be31862-7afa-4b96-9f78-8299bffefb94


Doublonmousse:
I'm not fully convinced of the current implementation. There are cases where this can be annoying instead of helping
- if I type semi-open intervals like `[a,b)` it'll get in the way. Typing times like `60"30'` will get harder too. It also won't activate for the first character inserted (although fixing that would need some care as to not impact negatively pasted text)
- for the automatic list continuation, the `take_while` may be too much and the possible characters too wide. It I for some reason use - and * to write morse code on the first line, getting it repeated in full on the next line is not expected. Maybe that's not a great example though. If I type `.Net` on a line beginning, getting the dot repeated on the next line is again a little bit unexpected. Maybe restricting it to `- ` with the extra space included would be better.

I don't know how much we can fix without actually adding somehow a concept of bullet points to the text field (and associated settings) in a word/office-like manner.

Nugfle:
it would be nice to have a markdown-like list, but maybe put the functionality behind a button so it doesn't get in the way when using the text field for the purposes listed above.

Keschler:
Maybe make this feature optional. So if the user wants it, he can activate it in the settings.

Issue #907, opened 2023-11-30
Snap to grid for shaper


OmegaRogue:
**Is your feature request related to a problem? Please describe.**  
I use tables and other shapes to make things like Karnaugh Maps, which consist of a grid of squares. Currently, these squares dont line up with the document grid.

**Describe the solution you'd like**  
The new option to snap to grid could be expanded to also work with the shaper tool.

flxzt:
fixed in #1496

Issue #451, opened 2023-01-15
Grid snapping for shapes and text


D0V4HKIIN:
**Is your feature request related to a problem? Please describe.**
When drawing graphs and geometric shapes, i would like to use the shapes to have precise integer sizes. 

**Describe the solution you'd like**
What I normally do on pen and paper is use paper with a 5mm grid so that i can easily draw and measure precise distances.
Rnote already has background grid pattern that shapes could snap onto.

**Describe alternatives you've considered**
There could be a way, in the ui to specify the dimensions and rotations wanted for the shape. I think it would also a great addition.

Also maybe a tool to measure distances could be added under the PenStyle::Tools?

flxzt:
fixed in #1496

PR #1496, opened 2025-08-19
feat: respect snap-positions options when building shapes

flxzt:
it looks like I can get away with implementing this in a way that doesn't touch the builders themselves - the coordinates are "snapped" into place before they reach them.


Doublonmousse (approved):
Looks good

Fixes #907
Fixes #451

flxzt:
ok, thanks for reviewing!

Issue #587, opened 2023-03-29
Text formatting shortcuts


ElbertCode:
**Is your feature request related to a problem? Please describe.**
If you select a text and try to use the common shortcut for a text formatting (e. g. Ctrl + B to make i bold), the selected text will be replaced by the letter you typed (b in the example).

**Describe the solution you'd like**
For the available text formating options in Rnote I would suggest these shortcuts:
| Shortcut | Action |
|--------|--------|
| Ctrl + B | Bold |
| Ctrl + I | Italic |
| Ctrl + U | Underline |
| Ctrl + L | Align Left |
| Ctrl + E | Align Center |
| Ctrl + R | Align Right |
| Ctrl + J | Justified |

**Describe alternatives you've considered**
Maybe these shortcuts could also be localized depending on the language as for example the shortcut in german to make a text bold is usually Ctrl + Shift + F.

mohamedAdhamc:
I think this probably the most frustrating thing I've had while taking notes for the past few months. Anyone did/planning to do it ? If not I think I could personally add them, and make a PR.

Kneemund:
It should be fairly simple to add them, I think the main problem is figuring out what key combinations to use. CTRL + I is already used for importing documents, so we either need to change that, or use e.g. CTRL + SHIFT as modifiers for these.

mohamedAdhamc:
> It should be fairly simple to add them (the same way CTRL + A is implemented), I think the main problem is figuring out what key combinations to use. 

How about just making them customizable ? Because, generally it's hard to reach an agreement on such shortcuts.

Kneemund:
Yes, but we still need sensible defaults. Making shortcuts customizable is tracked in #590, and I don't think anyone has tackled that so far.

PR #1497, opened 2025-08-27
Added text formatting shortcuts for Bold, Italic and Underline

Keschler:
The shortcut for Import Document has been changed to Ctrl + Shift + I.

Adds #587

### DIFF
--- a/crates/rnote-ui/src/penssidebar/typewriterpage.rs
+++ b/crates/rnote-ui/src/penssidebar/typewriterpage.rs
@@ -27,12 +27,6 @@ mod imp {
         #[template_child]
         pub(crate) text_reset_button: TemplateChild<Button>,
         #[template_child]
-        pub(crate) text_bold_button: TemplateChild<Button>,
-        #[template_child]
-        pub(crate) text_italic_button: TemplateChild<Button>,
-        #[template_child]
-        pub(crate) text_underline_button: TemplateChild<Button>,
-        #[template_child]
         pub(crate) text_strikethrough_button: TemplateChild<Button>,
         #[template_child]
         pub(crate) text_align_start_togglebutton: TemplateChild<ToggleButton>,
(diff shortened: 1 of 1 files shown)


flxzt:
you need to apply formatting with `cargo fmt` (or through `just fmt`) to fix the CI.

crates/rnote-ui/src/appwindow/actions.rs
```diff
@@ -107,6 +108,12 @@ impl RnAppWindow {
         self.add_action(&action_export_doc_pages);
         let action_export_selection = gio::SimpleAction::new("export-selection", None);
         self.add_action(&action_export_selection);
+        let action_text_bold = gio::SimpleAction::new("text-bold", None);
+        self.add_action(&action_text_bold);
+        let action_text_italic = gio::SimpleAction::new("text-italic", None);
+        self.add_action(&action_text_italic);
+        let action_text_underline = gio::SimpleAction::new("text-underline", None);
+        self.add_action(&action_text_underline);
```
flxzt, line 116: You should connect actions directly to buttons in the UI definition with the `action-name` property. Check for example how `win.zoom-in` is connected to the button.

crates/rnote-ui/src/appwindow/actions.rs
```diff
@@ -1016,19 +1072,25 @@ impl RnAppWindow {
...
         app.set_accels_for_action("win.clipboard-copy", &["<Ctrl>c"]);
         app.set_accels_for_action("win.clipboard-cut", &["<Ctrl>x"]);
         app.set_accels_for_action("win.clipboard-paste", &["<Ctrl>v"]);
+        app.set_accels_for_action("win.text-bold", &["<Ctrl>b"]);
+        app.set_accels_for_action("win.text-italic", &["<Ctrl>i"]);
+        app.set_accels_for_action("win.text-underline", &["<Ctrl>u"]);
+        app.set_accels_for_action("win.text-align-start", &["<Ctrl>l"]);
+        app.set_accels_for_action("win.text-align-center", &["<Ctrl>e"]);
+        app.set_accels_for_action("win.text-align-end", &["<Ctrl>r"]);
+        app.set_accels_for_action("win.text-align-fill", &["<Ctrl>j"]);
```
flxzt: Let's remove these accelerators for unimplemented actions.

flxzt:
changing the existing keyboard shortcut is fine with me.
Gnome HIG suggests the new shortcuts for formatting (https://developer.gnome.org/hig/reference/keyboard.html)

flxzt:
thank you!

Keschler:
Does Ctrl + Shift + I work to import a file? Because If I run that, it asks me If I want to use the GTK Inspector and opens it. Maby we have to change the shortcut to import a file or change the shortcut for italic.

flxzt:
oh, we shouldn't use Ctrl + Shift + I. We can change the import shortcut to something else, do you have any ideas?

Xyndra:
I don't know exactly what "import" does, maybe you could rename it to "open" for ctrl+o or "new" for ctrl+n or ctrl+t. It is questionable whether importing a file even needs a shortcut in the first place

Xyndra:
~~I am also not quite a fan of removing the UI buttons. Why can't we keep them for on-screen keyboards? Shortcut and Button aren't mutually exclusive. Also, the keybinds for these things are not standardized, so it will be hard to adjust to this new change. I would prefer the buttons being left there~~

Edit: nvm, I misread the code changes

Keschler:
Could we use ALT + I or if not CTRL + ALT + I?

PR #1501, opened 2025-08-31
chore: update deps, let chains, adapt to new API,

flxzt:
notably:
- parry2d slightly changed some function argument types
- rodio changed it's `OutputStream` interface
- gtk4 now requires deriving certain classes for all widget subclasses

switched back to upstream poppler-rs, and wanted to confirm that this works for windows, but hitting bug #1502 in the CI. Can you confirm it works manually @Doublonmousse ?


Doublonmousse:
Yes, it works manually for compilation/inno

PR #1515, opened 2025-09-08
Add back `action_pen_style` for `ctrl+i` shortcuts

### LINKED ISSUE #1513: Shortcuts `CTRL + i` not working anymore
I'm having similar issue but on main directly with different behavior between a `flatpak` build and a `meson` one (the shortcuts `ctrl+1` to 6 work on flatpak, but don't with meson).
If you run with `GTK_DEBUG=interactive` (and the flatpak equivalent), clicking on the `GtkShortcutController`
image
then shortcuts
image
they are listed.

But If I go inside Data and open the last one
image
the trigger is a list as expected on flatpak
image
with info 
image
and
image

On a meson compiled version, well I get the same thing though the shortcuts are not strictly in the same order (just checked ... )

image

That's a mystery. Well I'm not going to complain too much if at the very least the flatpak version works. Smells like a weird gtk bug (or maybe a combined gtk/packaging issue ?)

And it works on the 0.12 flatpak but not the latest flatpak nightly ...

This doesn't work anymore on 0.13

_Originally posted by @Doublonmousse in https://github.com/flxzt/rnote/issues/1361#issuecomment-3225409742_

Doublonmousse:
It was seemingly removed in 4106aab37b4f52e1be3265aad64bc2fcb722d3ea and therefore stopped working.

Fixes #1513

### DIFF
--- a/crates/rnote-ui/src/appwindow/actions.rs
+++ b/crates/rnote-ui/src/appwindow/actions.rs
@@ -12,9 +12,11 @@ use rnote_compose::SplitOrder;
 use rnote_compose::penevent::ShortcutKey;
 use rnote_engine::engine::StrokeContent;
 use rnote_engine::ext::GraphenePointExt;
+use rnote_engine::pens::PenStyle;
 use rnote_engine::strokes::resize::{ImageSizeOption, Resize};
 use rnote_engine::{Camera, Engine};
 use std::path::PathBuf;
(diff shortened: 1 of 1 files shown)


flxzt:
Ahh I at the time thought this action was not necessary anymore, not thinking about the keyboard shortcut that activates it

PR #1525, opened 2025-09-13
Update Polish translation

dawkagaming:
Hello,
I made some updates to pl.po by adding a few new messages and making slight modifications.


crates/rnote-ui/po/pl.po
```diff
@@ -401,7 +401,7 @@ msgstr "Png"
 #: crates/rnote-ui/src/dialogs/export.rs:813
 #: crates/rnote-ui/src/dialogs/export.rs:1205
 msgid "Jpeg"
-msgstr "Jpeg"
+msgstr "JPEG"
```
flxzt, line 402: We've opted to capitalize these acronyms.
It could be discussed if it makes sense but from a consistency point of view
it makes sense to stick with what the source string dictates.

dawkagaming: I know, but when I was thinking about it, in polish it looks a bit weird to have acronyms not in capital letter.
Of course, I can edit it if needed.

flxzt: What about changing this? We can discuss and change the capitalization later, but then it would be a change for all languages.

flxzt: @dawkagaming friendly ping! otherwise it will get stale.

flxzt:
Unfortunately this got stale and there is no response from the author. Will close.

PR #1529, opened 2025-09-15
Fix: make pen sounds work on windows

Doublonmousse:
Make pen sounds work on windows by changing the inno file because the data structure wasn't the one expected in the code.

This was the reason for lag on windows with the sounds activated (which would repeatedly try to setup the pen sounds handlers and fail on a tool change).

Fixes #1523


flxzt:
thanks :)

Issue #1486, opened 2025-08-03
Issues with `adw::SpinRow` and dialogs


Doublonmousse:
**Describe the bug**  

`adw::SpinRow` are using the `connect_changed` signal to listen to changes in values instead of `connect_value_notify`.

The issue is that the first one does not take into account the change in values if it's entered manually through the keyboard and then the enter key is pressed.

So the first proposition is to listen to `connect_value_notify` instead to fix this.

Now there is still one issue left. For example in the xournalpp import dialog

<img width="741" height="297" alt="Image" src="https://github.com/user-attachments/assets/23967038-e0d5-4fcf-9d04-7491a4684862" />

One might expect that pressing enter after setting the value would confirm and continue the import but nothing happens (except that the value is taken into account the first time).
But pressing escape in that configuration immediately cancels the dialog (whereas we might expect that this only removes the focus on the `SpinRow`)
- should we try to make it so that pressing enter once in that configuration confirms the action ?
- Or make it so once defocuses the spinrow and twice confirms it ?

There's also some weirdness in other similar dialogs. Export dialogs have their initial focus on the cancel button so pressing enter immediately closes it (whereas we could put it on the file selection instead so pressing enter (opening the file picker) enter (closing the file dialog) enter (confirm) would work)

ternison:
Other dialogs suffering from badly set inital focus are:
the **pdf import dialog**, which just cancels on pressing enter, even though the import button is highlighted blue:

<img width="621" height="309" alt="Screenshot from 2025-09-19 20-13-33" src="https://github.com/user-attachments/assets/fbd3d4d6-47db-44de-8755-72d516f24af1" />



Same for the **move to trash** dialog:

<img width="395" height="235" alt="image" src="https://github.com/user-attachments/assets/c021fc7b-ba26-4ba4-bccb-a64af991bda0" />

ternison:
The print file dialog, similarly to `adw::spinRow`, doesn't react at all upon pressing enter, after choosing a print option.

<img width="675" height="523" alt="Image" src="https://github.com/user-attachments/assets/2de73f6e-8eff-49e5-8bca-28247b04d69a" />

_Edit: This could also be due to the ubuntu system dialog not interpreting pressing enter as wanting to print._

Doublonmousse:
The print file dialog is called through the flatpak portal and depends on the desktop environment itself. So nothing we can do on that one (the KDE print dialog will print on enter, not the gnome one it seems).

[...]

PR #1532, opened 2025-09-17
feat: Apply ntry dialogs via enter key

ternison:
When creating folders or renaming files


Doublonmousse:
> This seems like a logical change to be applied to all entry dialog widgets, since the apply_button is highlighted blue.

Yes it is. There's probably more dialogs and tweaks we can apply this too (in #1486 the one I had in mind was the xournalpp import dialog)

ternison:
> Yes it is. There's probably more dialogs and tweaks we can apply this too (in #1486 the one I had in mind was the xournalpp import dialog)
 
Yep, agreed! There are a few tweaks on the table. Added a few more instances with unintuitive enter bindings to #1486

flxzt:
thanks!

PR #1556, opened 2025-10-11
Add a `WindowGroup` for each `appwindow`

### LINKED ISSUE #1461: File picker steals focus from all opened rnote windows
**Describe the bug**  
On windows, the file picker shows in fullscreeen and can't be resized.

**To Reproduce**  
1. Open rnote
2. Edit the document
3. save as
4. See the window being fullscreened and that it can't be resized even when trying to resize it 

**Expected behavior**  
For the file picker to not be fullscreened or at least resizable

**Screenshots**  

Trying to update a video but github won't let me

 - OS:  Windows 11 24H2
 - App Version: v0.11, v0.12, main
 - Installation Source: both from exe and manual compile

**Additional context**  

*VERY* weird behavior because this issue extends to the gtk interactive debugger but if and only if rnote is launched.
Doing 
```
GTK_DEBUG=interactive rnote
```
from the msys terminal will recreate the issue inside of rnote and the debugger

But doing
```
GTK_DEBUG=interactive gtk4-demo
GTK_DEBUG=interactive adwaita-1-demo
```
won't recreate the issue inside of the gtk debugger.

At first I thought it was a general gtk bug (nope), maybe a bindings + specific call pattern (nope).
So there's probably something somewhere that makes it fail upon startup of rnote (otherwise I don't see why the second debugger window would be affected before having done a single action inside of rnote)

Doublonmousse:
This restricts the `modal` nature of dialogs to each `appwindow` instead of being applied to all instances of opened rnote windows. This would result in weird behavior (if a modal is opened in one instance, a new freshly launched instance would have all inputs frozen)

Fixes #1461

### DIFF
--- a/crates/rnote-ui/src/app/mod.rs
+++ b/crates/rnote-ui/src/app/mod.rs
@@ -14,7 +14,7 @@ use crate::{
     workspacebrowser::RnWorkspacesBar, workspacebrowser::workspacesbar::RnWorkspaceRow,
 };
 use adw::subclass::prelude::AdwApplicationImpl;
-use gtk4::{gio, glib, glib::clone, prelude::*, subclass::prelude::*};
+use gtk4::{WindowGroup, gio, glib, glib::clone, prelude::*, subclass::prelude::*};
 
 mod imp {
     use super::*;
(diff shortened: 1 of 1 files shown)


Doublonmousse:
Now tested on linux (both a local meson build and the flatpak with a nightly) and windows (on the exe produced by the windows release CI)

flxzt:
thanks :)

Issue #1536, opened 2025-09-22
Regression: rnote 13.x crashes when opening the attached file as opposed to rnote 12.x which opens the file fine


Markus-included:
**Describe the bug**  
As described in title

**To Reproduce**  

`gunzip` the attachment and open the gunzipped file in rnote [Hue.rnote.gz](https://github.com/user-attachments/files/22475269/Hue.rnote.gz)

**Expected behavior**  
Document opens briefly, rnote crashes

**Console Output**  
N/A

**Screenshots**  
N/A

**Desktop (please complete the following information):**  
 - OS: Windows 10 and 11, Debian 13
 - App Version: Rnote v0.13.x
 - Installation Source: Windows Installer or Flatpak
 - Desktop Environment: GNOME
 - Display Server: X11
 - Input Source: N/A

**Additional context**  
I'm reporting this on behalf of a friend who is trying to switch to linux and tried rnote for his math homework, suddenly rnote crashes, it seems to be something with the file

linm-at:
I've tried opening this file on my Linux machine now and it seems to be that the file is corrupted, still weird that 12.x manages to open it

``` log
$ flatpak --env=RUST_LOG=rnote=debug --env=RUST_BACKTRACE=1 run com.github.flxzt.rnote
   0.000092487s DEBUG rnote: .. tracing subscriber initialized.
pci id for fd 10: 10de:134d, driver (null)
kmsro: driver missing
  61.893755592s ERROR rnote::appwindow: Opening file with dialogs failed, Err: loading RnoteFile from bytes failed.

Caused by:
    0: decompressing bytes failed.
    1: invalid gzip header

Stack backtrace:
   0: <unknown>
   1: <unknown>
   2: <unknown>
   3: <unknown>
   4: <unknown>
   5: <unknown>
   6: <unknown>
   7: <unknown>
   8: clone
  61.899858594s ERROR rnote::overlays: Datei öffnen gescheitert
```

Doublonmousse:
I can't open the file no matter the version/os. So can't really reproduce the "works on one version but not the other"
Are you sure the file you provided is correct here ? The rnote format is gzipped so not sure doing gz on top of this works ? Because the end file provided here is not zipped when opening the gz file.

The current saving process is done in two steps : creating a json then gzipping it. 
Related issues : https://github.com/flxzt/rnote/issues/1128

[...]

PR #1561, opened 2025-10-19
Windows: Force cairo msys version to a rayon-compatible one

Doublonmousse:
Related to #1536 ? (should fix it but TBC).

The issue is that https://github.com/msys2/MINGW-packages/commit/305ebda98c3041d9986d6fae498b45d2b2b9f4e8 added a patch to the windows msys cairo version that uses direct2D and probably makes the cairo text calls inside rayon impossible without crashing (if there is more gpu-related code, having it run on multiple threads won't work if they call to the same gpu...). At least that's my reading of the patch.

So it's either we downgrade cairo (this is what I'm doing here) or we remove the rayon call (https://github.com/flxzt/rnote/issues/1536#issuecomment-3368661415) on windows for text.

Given that this only happens on the patched msys version, I'd go with option 1 for now and then report to msys (not yet done, maybe this needs a minimal example of text layout calls inside rayon ..).

There's also some windows CI fixes. The 2025 windows server image now includes Inno and that makes the CI fail (winget tries to install, then see it's already installed and the step failed with exit code 1 https://github.com/Doublonmousse/rnote/actions/runs/18626734337/job/53105785678#step:2:34). So test if inno is installed first, and if not call winget. Given that inno seems to appear/disappear from the window image, I'd reckon leaving the new logic as is is probably for the better (it'll then work wherever inno is in the image or not).

Window release build to test : https://github.com/Doublonmousse/rnote/actions/runs/18626906789

### DIFF
--- a/justfile
+++ b/justfile
@@ -86,8 +86,15 @@ prerequisites-dev: prerequisites
 
 # in MSYS2 shell
 prerequisites-win:
+    echo "compile cairo without the 0001-DWrite-Get-glyph-bitmap-with-D2D-in-selected-cases.patch patch"
     pacman -S --noconfirm \
-        unzip git mingw-w64-x86_64-xz mingw-w64-x86_64-pkgconf mingw-w64-x86_64-gcc mingw-w64-x86_64-clang \
+        mingw-w64-x86_64-freetype mingw-w64-x86_64-fontconfig mingw-w64-x86_64-glib2 mingw-w64-x86_64-libpng \
+        mingw-w64-x86_64-lzo2 mingw-w64-x86_64-pixman mingw-w64-x86_64-cc mingw-w64-x86_64-meson \
+        mingw-w64-x86_64-ninja mingw-w64-x86_64-pkgconf mingw-w64-x86_64-python mingw-w64-x86_64-python-fonttools \
+        mingw-w64-x86_64-gettext-tools patch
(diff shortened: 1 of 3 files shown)


Doublonmousse:
kinda stuck now.

It does install cairo through the `makepkg` command with the custom patched PKGBUILD. https://github.com/Doublonmousse/rnote/actions/runs/18817555047/job/53688050752#step:8:3528

seems there are some pc files as well : https://github.com/Doublonmousse/rnote/actions/runs/18817555047/job/53688050752#step:8:3486

But meson then doesn't find it : https://github.com/Doublonmousse/rnote/actions/runs/18817555047/job/53688050752#step:9:31 

Though maybe if our issue is with the Win32 Dwrite we can disable it (either in the build option here : https://github.com/Doublonmousse/rnote/actions/runs/18817555047/job/53688050752#step:8:542 or with a varenv ?)

Doublonmousse:
Well the install folder from the pkgbuild is local for some unknown reason .... So copying files it is.
Successful in copying the `.pc` to their actual correct place.
For the `.h` files, not yet ? The `-lcairo` and `lcairo-gobject` fail (even though I though I copied correctly to the correct `include` folder, so maybe not ?)

Doublonmousse:
https://github.com/Doublonmousse/rnote/actions/runs/18822068568/job/53699118230#step:8:3505 is the only warning I get.
Otherwise doing exactly the same procedure minus copying the files in the base include/pkgconfig dir works locally.

So for a reason or another, the files don't get copied over when installing the package

Doublonmousse:
Sucess ! : https://github.com/Doublonmousse/rnote/actions/runs/18852170512

justfile
```diff
@@ -86,8 +86,15 @@ prerequisites-dev: prerequisites
 
 # in MSYS2 shell
 prerequisites-win:
+    echo "compile cairo without the 0001-DWrite-Get-glyph-bitmap-with-D2D-in-selected-cases.patch patch"
     pacman -S --noconfirm \
-        unzip git mingw-w64-x86_64-xz mingw-w64-x86_64-pkgconf mingw-w64-x86_64-gcc mingw-w64-x86_64-clang \
+        mingw-w64-x86_64-freetype mingw-w64-x86_64-fontconfig mingw-w64-x86_64-glib2 mingw-w64-x86_64-libpng \
+        mingw-w64-x86_64-lzo2 mingw-w64-x86_64-pixman mingw-w64-x86_64-cc mingw-w64-x86_64-meson \
+        mingw-w64-x86_64-ninja mingw-w64-x86_64-pkgconf mingw-w64-x86_64-python mingw-w64-x86_64-python-fonttools \
```
Doublonmousse, line 93: I think I can do `makepkg -s` to install the build dependencies instead

PR #1562, opened 2025-10-21
Corrections in the docs to prevent errors

samu4199:
Small changes that help avoid confusion when following the instructions.
Quote marks are needed when specifying Inno Setup path because of the parenthesis.


flxzt:
thanks :)

PR #1571, opened 2025-10-29
fix: fix blocking on "open in default app" actions

REVO9:
Rnote blocks while an app opened via "open in default app" or "open workspace folder" is running.

Replacing `open::that` with `open::that_detached` fixes this.


flxzt:
out of curiosity: on which platform does `open::that` block?

PR #1578, opened 2025-11-02
feat: new higher-performance rnote file format

anesthetice:
**The overall goal of this PR is to improve the handling of `.rnote` save files.**

### tl;dr
* Shorter loading times (~10x improvement*)
* Shorter saving times (~6x improvement*)
* Smaller file sizes (~12% reduction)

\* This improvement factor only concerns going back and forth between the already loaded bytes of the file and their corresponding `EngineSnapshot`, on my system of course, your results may vary.

image

Main changes:
* Parallelized (de-)serialization and (de-)compression.
* Files saved with a compatible version are directly converted into an `EngineSnapshot` instead of always passing through a compatibility layer.
* Swapped the _Deflate_ (`gzip`) compression algorithm for _Zstandard_ ([`zstd`](https://github.com/facebook/zstd)).

Here's a diagram of the simplified process to get from `EngineSnapshot` to a `.rnote` save file and vice-versa.

diagram

Note that this PR is a re-write of #1177, some of the important changes being:
* `(+)` parallelism for serialization and compression
* `(+)` the introduction of a file version separate from the version of Rnote
* `(-)` scrapped save preferences (not worth complexity cost)
* `(-)` scrapped alternative serialization methods (for compatibility reasons)


anesthetice:
@flxzt, @Doublonmousse, I'd say this PR is ready for review. Take all the time you need of course, and feel free to ask me for information, changes, etc. :saluting_face:

crates/rnote-engine/src/fileformats/rnoteformat/prelude.rs
```diff
@@ -0,0 +1,98 @@
+// Imports
+use crate::fileformats::rnoteformat::bcursor::BCursor;
+use anyhow::{Context, bail};
+
+/// The prelude is used to identify Rnote files and provide some context.
+/// It is composed of four elements:
+/// 1. The magic number: `[u8; 10]` = `[52, 4e, 4f, 54, 45, 2d, ce, a6, ce, 9b]` = "RNOTE-ΦΛ" (in UTF-8 encoding)
+/// 2. The file version: `u16` (decides in broad strokes how the bytes will be handled later on)
+/// 3. The Rnote version (that last saved the file): `[major (u64), minor (u64), patch (u64), Prerelease size (u16), Prerelease (str), BuildMetadata size (u16), BuildMetadata (str)]`
```
Doublonmousse: Maybe worth precising that this is the full `semver::Version` struct information

crates/rnote-engine/src/fileformats/rnoteformat/v1/save.rs
```diff
@@ -0,0 +1,111 @@
+use anyhow::Context;
+
+// Imports
+use super::*;
+
+impl RnoteFileInterfaceV1 {
+    /// Attempts to convert an `EngineSnapshot` to bytes.
+    pub fn engine_snapshot_to_bytes(
```
Doublonmousse, line 12: I think it's worth noting that this modifies the inputted `EngineSnapshot` in place because of the `std::mem::take` call below. It's probably okay because we do a `take_snapshot` everytime before but worth double checking and documenting

crates/rnote-engine/src/fileformats/rnoteformat/v1/save.rs
```diff
@@ -0,0 +1,111 @@
...
+            )
+            .inspect(|compressed| core_info.c_size = compressed.len())?;
+
+        // We'll then do basically the same thing as above, but in a much more complicated manner
+        // for the previously extracted strokes and chrono components, all for the sake of speed.
+
+        // We try to roughly approximate the capacity each local buffer will require,
+        // maxing out at 16 MiB per local buffer. Using a value of 8192 bytes per stroke
+        // might seem high, but it's actually quite conservative given that we have to account
+        // for things like image strokes which contain huge amounts of information.
```
Doublonmousse, line 38: Won't fail because the push will reallocate and grow the vec array if we go over the capacity.

Just for me or for documentation purposes (as I briefly wondered if you could create a panic on a full vector. Turns out you don't which is great).

crates/rnote-engine/src/fileformats/rnoteformat/mod.rs
```diff
@@ -1,158 +1,77 @@
...
-            data: ijson::to_value(self).context("converting RnoteFile to JSON value failed.")?,
-        };
-        let compressed = compress_to_gzip(
-            &serde_json::to_vec(&wrapper).context("Serializing RnoteFileWrapper failed.")?,
-        )
-        .context("compressing bytes failed.")?;
-
-        Ok(compressed)
-    }
+pub fn save_engine_snapshot_to_bytes(
```
Doublonmousse, line 73: Same remark here on the mutation of the engine snapshot inside of `engine_snapshot_to_bytes`

crates/rnote-engine/src/fileformats/rnoteformat/v1/load.rs
```diff
@@ -0,0 +1,222 @@
...
+            .collect::<Result<Vec<Vec<(Stroke, ChronoComponent)>>, serde_json::Error>>()?;
+
+        Ok(Self::components_to_engine_snapshot(
+            engine_snapshot,
+            stroke_chrono_pair_chunks,
+        ))
+    }
+}
+
+type EngineStrokes = Arc<HopSlotMap<StrokeKey, Arc<Stroke>>>;
```
Doublonmousse: If we're adding type aliases like this, it may be worth sharing the alias with both `HistoryEntry` and `EngineSnapshot`. That way the type alias references all usages of it, instead of the convenience rename for only this file

crates/rnote-engine/src/fileformats/rnoteformat/v1/load.rs
```diff
@@ -0,0 +1,222 @@
...
+    ) -> EngineSnapshot {
+        let capacity = stroke_chrono_pair_chunks
+            .iter()
+            .map(Vec::len)
+            .sum::<usize>();
+
+        let mut stroke_components: HopSlotMap<StrokeKey, Arc<Stroke>> =
+            HopSlotMap::with_capacity_and_key(capacity);
+        let mut chrono_components: SecondaryMap<StrokeKey, Arc<ChronoComponent>> =
+            SecondaryMap::with_capacity(capacity);
```
Doublonmousse: Forcing exactly the capacity means that any operation that happens just after the load will reallocate.

It may be better to allocate slightly more to delay a little bit the first reallocation (maybe the min/or max ? of 1.1*capacity, and capacity + CHUNK_SIZE so we have some headroom)

crates/rnote-engine/src/fileformats/rnoteformat/v1/load.rs
```diff
@@ -0,0 +1,222 @@
...
+    /// for compatibility across differing versions of Rnote.
+    #[allow(unused)]
+    pub fn bytes_to_compat(cursor: BCursor, header_size: usize) -> anyhow::Result<CompatV1> {
+        Self::bytes_to_deserialized::<ijson::IValue, ijson::IValue>(cursor, header_size).map(
+            |(engine_snapshot_gutted, stroke_chrono_pair_chunks)| CompatV1 {
+                engine_snapshot_gutted,
+                stroke_chrono_pair_chunks,
+            },
+        )
+    }
```
Doublonmousse, line 124: Maybe not worth keeping anymore ? 
Given the fact that the first thing that is done is to read the new header in `bytes_to_deserialized`, no matter if we set header_size = 0 it will fail

crates/rnote-engine/src/fileformats/rnoteformat/v1/mod.rs
```diff
@@ -0,0 +1,102 @@
...
+//! entire [`EngineSnapshot`] through an intermediate JSON representation ([`ijson::IValue`]).
+//!
+//! Instead:
+//!
+//! 1. The `stroke_components` and `chrono_components` are extracted from the snapshot.
+//! 2. They are grouped together, chunked, and serialized in parallel directly to bytes.
+//! 3. The chunks are compressed (still in parallel of course) using [`zstd`] (instead of `gzip`).
+//! 4. The same (serialization + compression) is performed on what remains of the [`EngineSnapshot`].
+//! 5. The [`Prelude`], and [`RnoteFileHeaderV1`] are created, bytes are concatenated together and finally returned.
+//!
```
Doublonmousse, line 24: I think it's worth adding somewhere the full fileformat spec. Even if it's a shortened version like 

## Binary format

- [`Prelude::try_to_bytes`]
- [`RnoteFileHeaderV1`] serialized through serde_json::to_vec (contains the compression method)
- gutted Engine snapshot (through serde_json::to_vec then compressed using the given compression method)
- compressed chunks (using the given compression method) where each decompressed version is of the form `[serde_json::to_writer( .. stroke_chrono_pair), ... ]`

crates/rnote-engine/src/fileformats/rnoteformat/prelude.rs
Doublonmousse: And if we want to be extra precise, let's make sure we know that everything is in little endian here : u64le, u16le, str = u8le in UTF-8 encoding ...

Doublonmousse:
Another thought on the atomic saving denomination.
I'm not sure we can really call the new version atomic saving per say.

That would be the case for a file format where any change we do inside the file is mapped back to the file through an efficient diff (and not a full write of the new engine state). For sqlite (https://github.com/flxzt/rnote/issues/1173#issuecomment-2675146989), that would be the case because modifications to the file can be mapped back to the file through a push/commit

crates/rnote-engine/src/fileformats/rnoteformat/v1/save.rs
anesthetice: Yes, also thought of applying the `#[serde(skip_serializing)]` attribute but it didn't feel as clean as the method I went with given that we have exclusive ownership of the snapshot. I'll add a comment explaining why the mutable borrow is necessary and how we can get around that if need be in the future.

crates/rnote-engine/src/fileformats/rnoteformat/v1/load.rs
anesthetice: Good idea, does something like `capacity += (0.2 ⋅ capacity).clamp(20 MB)` sound ok?

crates/rnote-engine/src/fileformats/rnoteformat/v1/load.rs
anesthetice: Not sure I understand what you mean here...

anesthetice:
> Another thought on the atomic saving denomination. I'm not sure we can really call the new version atomic saving per say.
> 
> That would be the case for a file format where any change we do inside the file is mapped back to the file through an efficient diff (and not a full write of the new engine state). For sqlite ([#1173 (comment)](https://github.com/flxzt/rnote/issues/1173#issuecomment-2675146989)), that would be the case because modifications to the file can be mapped back to the file through a push/commit

I have seen "atomic saving" be used to mean somewhat different things, but for the most part it seems to mean what we are doing here ([Geany wiki](https://wiki.geany.org/config/all_you_never_wanted_to_know_about_file_saving), [15 year old stackoverflow thread](https://stackoverflow.com/questions/2705076/difference-between-writing-to-file-atomically-and-not), etc...). A more efficient saving mechanism using diffs might be interesting, but would probably be very complex to implement and then have to deal with forward- and backward-compatibility...

That being said, we should maybe tinker a bit more with the implementation as for instance metadata, permissions, and symlinks are not handled...

crates/rnote-engine/src/fileformats/rnoteformat/v1/load.rs
Doublonmousse: I'm pretty sure this is not used, and I'm not sure it can be of use later.

Given that the docstring refers to the compatibility part, and that part is implemented elsewhere without calling to this function ...

crates/rnote-engine/src/fileformats/rnoteformat/v1/load.rs
Doublonmousse: Yes that seems reasonable

crates/rnote-engine/src/fileformats/rnoteformat/v1/load.rs
anesthetice: It will be used in the future, when a change in the `EngineSnapshot` requires more than just `serde` attribute fixes. Currently there is no previous version of a v1 file out there that would require going through the compatibility layer.

I wrote the following in rnoteformat/mod.rs:
``` rs
match prelude.file_version {
    1 => {
        RnoteFileInterface::bytes_to_engine_snapshot(cursor, prelude.header_size)

        // Template for a future version change that requires an upgrade.
        /*
        if semver::VersionReq::parse(">=0.15.0")
            .unwrap()
            .matches(&prelude.rnote_version)
        {
            RnoteFileInterface::bytes_to_engine_snapshot(cursor, prelude.header_size)
        } else {
            RnoteFileInterface::bytes_to_compat(cursor, prelude.header_size)
                .map(CompatV1For::<0, 14, 0>::from)
                .and_then(CompatV1For::<0, 15, 0>::try_from)
                .and_then(RnoteFileInterface::compat_to_engine_snapshot)
        }
        */
    }
    /* ... */
}
```

Doublonmousse: Okay, maybe worth pointing that out in the docstring (compat for v1+ rnote format, unused for now but kept for future use).

anesthetice:
@Doublonmousse Apologies for the delay, let me know what you think

crates/rnote-engine/src/fileformats/rnoteformat/v1/mod.rs
Doublonmousse: Added in https://github.com/flxzt/rnote/pull/1578/commits/4845a139e3a6180a2f4ef015afbd6df72c2664dd#diff-79ec298bb200ebc177219442ce95f2d1a59b9915e1307b8592251af00670f45dR6-R11 right ?

anesthetice: Yes

crates/rnote-engine/src/utils.rs
```diff
@@ -118,3 +120,90 @@ pub mod glib_bytes_base64 {
         rnote_compose::serialize::sliceu8_base64::deserialize(d).map(glib::Bytes::from_owned)
     }
 }
+
+pub async fn atomic_save_to_file<Q>(filepath: Q, bytes: &[u8]) -> anyhow::Result<()>
```
flxzt: If we start doing this we should actually go the full way and consider the details on the OS level.

I would suggest using something like [tempfile](https://docs.rs/tempfile/3.26.0/tempfile/index.html) since there are platform details that are tricky. Create a temporary file (that is guaranteed not to collide with existing files through a randomized name) with `NamedTempFile::new()`, write to it, and persist it with `.persist()` . It's not too complicated, and could also be done without a dependency (see it's implementation for UNIX-like systems: https://github.com/Stebalien/tempfile/blob/929a1127bde5046733d535281203bf8ec1bf2c9e/src/file/imp/unix.rs) but I don't want us to either become non-portable to non UNIX environment or having to manage such platform details ourselves.

anesthetice: Hello hello,

We could use `tempfile`, as it's cleaner and more "idiomatic". Just a couple caveats to watch out for, i.e. we should be using `tempfile::tempfile_in` the parent path of our `.rnote` file to avoid `.persist()` failing because for example someone is editing a file stored within a USB drive. But from a little bit of digging, they aren't doing much more than what we're already doing here (i.e., their file has custom flags, but these are "cleared" when the tempfile is ultimately renamed by a syscall)

In the end it's up to you, I'm fine with both approaches, I'll implement whichever you prefer!

As for splitting the atomic save into a separate PR, I can do it but then I'll need to create a third PR to implement it for the cli command... Again it's up to you!

anesthetice: Or actually, no need for a third PR, as I imagine atomic file saving will be merged much sooner than the new file format.

I'll separate the atomic save to a new PR then, just tell me which approach you prefer regarding the use of `tempfile`

anesthetice:
Atomic FS was split-off into it's own branch (+PR), see #1678

NXTler:
I really don't want to pressure anyone here, but what is the current status of this? What is missing from it?

PR #1583, opened 2025-11-11
Disable the dwrite cairo backend on windows

### LINKED ISSUE #1536: Regression: rnote 13.x crashes when opening the attached file as opposed to rnote 12.x which opens the file fine

Doublonmousse:
Fixes #1536 

Disables the windows-specific Dwrite backend used by cairo.
This backend is currently not thread-safe and poses issues and crashes with the use of rayon (even if we use separate instances of cairo contexts on each thread).

Could be reverted in the future when https://github.com/msys2/MINGW-packages/issues/26222 (and the related issues upstream like https://gitlab.freedesktop.org/cairo/cairo/-/issues/886) are closed.

Impact: afaik only emoji rendering is affected on windows, see below

| without env variable | with fc|
|--|--|
|Image|Image|

### DIFF
--- a/crates/rnote-ui/src/env.rs
+++ b/crates/rnote-ui/src/env.rs
@@ -75,6 +75,12 @@ pub(crate) fn setup_env() -> anyhow::Result<()> {
                 lib_dir.join("gdk-pixbuf-2.0\\2.10.0\\loaders"),
             );
 
+            // workaround for issue #1536
+            // Disable the dwrite cairo backend that is
+            // (for now) thread unsafe
+            // https://github.com/flxzt/rnote/issues/1536
+            std::env::set_var("PANGOCAIRO_BACKEND", "fontconfig");
+
             //std::env::set_var("RUST_LOG", "rnote=debug,rnote-cli=debug,rnote-engine=debug,rnote-compose=debug");
         }
     } else if cfg!(target_os = "macos") {


Doublonmousse:
And ... it doesn't work for some reason

flxzt:
what's the difference between

```
$env:PANGOCAIRO_BACKEND="fc"
& 'C:\Program Files\Rnote\bin\rnote.exe'
```

and this?
To me this should have the same effect, maybe we can verify this is applied by cairo in some way programmatically?

Otherwise, maybe the pragmatic (temporary) fix would to wrap the executable in a script that calls it in this way?

Doublonmousse:
> what's the difference between
> 
> ```
> $env:PANGOCAIRO_BACKEND="fc"
> & 'C:\Program Files\Rnote\bin\rnote.exe'
> ```

I don't get it either. In practice we kinda want the env to apply only to piet's cairo (so the calls that are wrapped by rayon are using the thread-safe version).
Maybe because the cairo rayon calls are done from spawned threads it doesn't inherit the same configuration? (though even gtk doesn't show the changed fontconfig with this env variable so ...)
 
> and this? To me this should have the same effect, maybe we can verify this is applied by cairo in some way programmatically?

If this can be done on the piet side, that'd be another way (though I didn't find a way to configure the cairo context from piet through the exposed API, maybe this can be done internally?).

In all case I didn't succeed in making the env variable work, neither locally nor with the windows CI.

> Otherwise, maybe the pragmatic (temporary) fix would to wrap the executable in a script that calls it in this way?

Yeah probably. The other PR is kinda cumbersome to take as a fix, was more of a way to really confirm what the issue was (uses a modified PKGBUILD and a compile step + some copy to have all the dependency installed for some reason, and I'm not sure it won't automatically update cairo from that version if a new version is pushed).

I _could_ retry pinning the cairo version (same general idea, pulling older PKGBUILD from the cache, but it needs a bunch of other packages to also be pinned, before https://github.com/msys2/MINGW-packages/commit/83eed715521d7d7c292d34fb80c29a720b534769)

Doublonmousse:
Tried to fix this in the `piet-cairo` crate by adding
```rs
        // patch here for windows
        unsafe {
            use pango::prelude::Cast;
            use pangocairo::prelude::PangoCairoFontMapExtManual;
            // check the font type
            if FontMap::default().downcast_ref::<pangocairo::FontMap>().unwrap().font_type() != cairo::FontType::FontTypeFt {
                println!("setting font map to fc");
                std::env::set_var("PANGOCAIRO_BACKEND", "fc");
                pangocairo::FontMap::set_default(
                    pangocairo::FontMap::new().downcast_ref::<pangocairo::FontMap>(),
                );
            } else {
                println!("Already to the correct format");
            }
        }
```
in 
```rs
    pub fn new() -> CairoText {
```
Result: it's unbearably slow (because the snippet is slow to complete and because all piet cairo context will create the `CairoText` regardless of whether or not we need or not).
And for some reason this isn't enough. Still with the unmistakable 
```
warning: D3D11 CORRUPTION: ID3D11DeviceContext::Map: Two threads were found to be executing functions associated with the same Device[Context] at the same time. This will cause corruption of memory. Appropriate thread synchronization needs to occur external to the Direct3D API (or through the ID3D10Multithread interface). 14676 and 17240 are the implicated thread ids. [ MISCELLANEOUS CORRUPTION #28: CORRUPTED_MULTITHREADING]
```
I kinda don't want to ship a powershell script ... Tried to play with the inno setup to make the rnote shortcut run powershell with arguments but it didn't work.

PR #1587, opened 2025-11-14
Add Typst integration (and appease clippy)

### LINKED ISSUE #644: Add typst support to typewriter
Add typst support to typewriter.

I like the option that rnote allow creating documents combining text input imported pictures and handwritten notes.
But sometimes I also would like to write mathematical formula or formatted text by using the typewriter.
Typst should be a nice solution for this, because the syntax is powerful and still very simple.

An option to enable typst can be added at this menu.
image
For this the [typst crate](https://github.com/typst/typst), which does also provide a lib, can be used.
The disadvantage of this is that the typst is currently not published at cartes.io, which would also block rnote releases on crates.io. 

Write hard to read formula with hand.

Xyndra:
Fixes https://github.com/flxzt/rnote/issues/644

Current state (as of first commit): button with a menu that has a preview button and renders into a vector
Ideal end state: button with a menu that automatically updates its preview, rendering into an image or vector format that contains its original text for editing, but can also cancel in cases where some resources went missing and can't be recovered


Xyndra:
- [x] Live preview
- [x] Higher resolution in the preview
- [x] Fix monospace
- [ ] ~~Option between image and vector format~~ (considered unimportant/useless for first version)
- [x] Allow for changing text after inserting (with proper cancel button)
- [ ] Possibly some kind of auto-saving for changes since there are a lot of important changes happening without closing the menu and the menu can simply be closed with escape atm
- [x] Consider shipping fonts for typst

Xyndra:
how it looks atm:

image

Xyndra:
I would appreciate feedback, especially from @flxzt, both about the looks/idea and the code

Xyndra:
I am unsure whether to increment the save version number, would need some info from the maintainers.
It does not crash in either direction, but accessing it with an older version removes the Typst text metadata.

Xyndra:
I think this is enough for a first version. The auto-save and option between pixel and vector are not must-haves, and about incrementing the version number, I would need feedback from maintainers (most of the save format seems to be from @flxzt)

anesthetice:
Not a maintainer, but looks pretty promising! I have a couple of comments / suggestions though (not in order of importance or anything):

(also saw you just commit something as I'm writing this, I hope my comments aren't already out-of-date)

- Any reason for using version `0.12.0` of typst when `0.14.0` is out?
- I would move the typst-related code in [crates/rnote-engine/src/utils.rs](https://github.com/flxzt/rnote/pull/1587/files#diff-69e14b6632502e3ecc34bc99cab25aebed4717bb0b288df3f5f8fb96808827a4) to their own file
- Is there a way to import typst packages (e.g. CeTZ)?
- I realize I am way too dependent on code-completion (e.g. from tinymist) to write typst, but somewhat hard to use without to be frank.

As for the UI
  
satty_2025-11-16_13:15:01

Finally, I'm still not too familiar with GTK and more complicated Rnote internals as a whole, so I can't comment on much more..

Xyndra:
> Not a maintainer, but looks pretty promising! I have a couple of comments / suggestions though (not in order of importance or anything):
> 
> (also saw you just commit something as I'm writing this, I hope my comments aren't already out-of-date)
> 
>     * Any reason for using version `0.12.0` of typst when `0.14.0` is out?
> 
>     * I would move the typst-related code in [crates/rnote-engine/src/utils.rs](https://github.com/flxzt/rnote/pull/1587/files#diff-69e14b6632502e3ecc34bc99cab25aebed4717bb0b288df3f5f8fb96808827a4) to their own file
> 
>     * Is there a way to import typst packages (e.g. CeTZ)?
> 
>     * I realize I am way too dependent on code-completion (e.g. from tinymist) to write typst, but somewhat hard to use without to be frank.
> 
> 
> As for the UI
> satty_2025-11-16_13:15:01
> 
> Finally, I'm still not too familiar with GTK and more complicated Rnote internals as a whole, so I can't comment on much more..

Thanks for the feedback, you are right, I did seem to make a few oversights. I will look into all of them, but I can already say that tinymist integration is close to impossible, at least for me. That is something that would require IDE-like infrastructure, which a simple GTK text field most likely does not have.

anesthetice:
There does seem to be a library [GtkSourceView](https://gnome.pages.gitlab.gnome.org/gtksourceview/gtksourceview5/class.Completion.html) for gtk4 that helps with code completion, but still probably a herculean task to set up and integrate with something like tinymist.

Xyndra:
Yeah, that is something that is definitely too much for a first version, maybe at a later point in time. This is also not meant to be a Typst IDE, just a way to write something scientific like with LaTeX.

Xyndra:
[Lilaq](https://typst.app/universe/package/lilaq) is now working
image

Xyndra:
I have now fixed all the feedback except for the misalignment of the middle line (which is draggable, and honestly not that bad), and the tinymist stuff since that is way too hard

Xyndra:
@flxzt I fixed all the clippy messages, if that was what is keeping you from merging this. Would be nice if you could merge this, since I have other changes conflicting with this I want to work on

flxzt:
Typst support is something that is super exciting, this is quite incredible!

I need to do a proper review on this huge diff. It looks good, but it might take a little while until I am able to do it

Xyndra:
Btw if this ever gets merged, I also have another branch based on this where I also added a marker tool

Doublonmousse:
One trick that could be used for the IDE part is the use of temporary files.

I've seen it done on some GUI sftp software. Show the folder content, and if you want to edit a text file on remote, you can "open with" your favorite IDE and it just opens the file with the IDE. Internally, it downloads and copy the file to a temp folder, calls the "open with" command on the file (so you can choose any program that opens txt), and more importantly watches the files so that when you save it in the IDE, it syncs the change back to the remote.

Using the same tactic here, we could make it possible to edit each typst element in vscode with all the benefits of the IDE support

Xyndra:
I also noticed I need to set the flag for the flatpak somewhere so that it is allowed to access the internet for typst plugins

hoellwerth:
Wonderful feature! It also integrates well with whole app.

Is there a possibility of rendering in white font when dark style is active? Maybe also allow exiting the typst editor by pressing 'Esc'

image

Also when double-clicking an existing typst text, which is not tha unintuitive, it opens the windows twice.

Xyndra:
> Is there a possibility of a white font when dark style is active?

This is not really something for the wrapper but rather for the file (you can change font color and background color in typst), but in the future, maybe it could put the color from the current tool as text color. I think that is a bit much for a first attempt though, as there are already huge changes.
> Maybe also allow exiting the typst editor by pressing 'Esc'

See the conversation above, allowing 'esc' to exit would lead to the entire file being lost, which can be really annoying, especially for people frequently using vim where you `esc+:+w+enter` to save a file. As mentioned above, maybe this could be solved by implementing an "open in editor", but since this is cross-platform, even supporting embedded, which I can't test, I am not sure if that is a good idea to add.
> Also when double-clicking an existing typst text, which is not tha unintuitive, it opens the windows twice.

I noticed that too, however that is a bit hard to fix with the project structure. This happens because renderer and drawing are decoupled and the popup sometimes is not fast enough. For me, this only occurs the first time since once Typst is loaded, it is fast enough to open before I can click again, but that probably depends on system load

hoellwerth:
Thanks for the really fast reply, Sorry for not entirely reading the conversation, I guess I was just too excited for the change! can't wait  to see this in upstream and really looking forward to future features.

crates/rnote-engine/Cargo.toml
```diff
@@ -57,11 +57,14 @@ usvg = { workspace = true }
...
+typst-svg = { workspace = true, optional = true }
 
 [dev-dependencies]
 approx = { workspace = true }
 
 [features]
 cli = ["dep:clap"]
 default = []
-ui = ["dep:gtk4"]
+ui = ["dep:gtk4", "dep:typst", "dep:typst-kit", "dep:typst-svg"]
```
Doublonmousse: I don't think making typst optional works in that case.

Remember that the `rnote-cli` + the thumbnailer must be capable of rendering a rnote file to pdf/svg

crates/rnote-engine/src/engine/import.rs
```diff
@@ -362,6 +362,127 @@ impl Engine {
         widget_flags
     }
 
+    /// Insert an SVG image as a VectorImage stroke.
+    /// If `typst_source` is provided, it will be stored with the image for later editing.
+    pub fn insert_svg_image(
```
Doublonmousse, line 363: These functions feel a little bit out of place somehow. I think you could very well use `import_generated_content` with some adaptations (to not select the newly added svg)

crates/rnote-engine/src/engine/import.rs
```diff
@@ -362,6 +362,127 @@ impl Engine {
...
+            Err(e) => {
+                error!("Failed to import SVG image: {e:?}");
+            }
+        }
+
+        widget_flags
+    }
+
+    /// Update an existing Typst stroke with new SVG data and source code.
+    pub fn update_typst_stroke(
```
Doublonmousse, line 386: For this one, half of it could be a method in `VectorImage` like `update_image` and the rest that would stay would be related to the widget flags and store (if it doesn't come from the widget flags handler)

Doublonmousse: Another thing on typing : I'm not sure the best way to go is to add a field to a VectorImage (which can be or not be from a typst edition).

Maybe it's a little better to add a separate typst type.

Note for compatibility : this PR needs to bump the rnote file version as the extra field makes it (probably) forward compatible (because of the serde skip if the typst source is one) but not backwards if a typst element is added

crates/rnote-engine/src/pens/typewriter/penevents.rs
```diff
@@ -43,8 +43,19 @@ impl Typewriter {
...
                     .last()
                 {
+                    // Check if clicked on a Typst VectorImage
+                    if let Some(Stroke::VectorImage(vectorimage)) =
+                        engine_view.store.get_stroke_ref(stroke_key)
+                    {
+                        if let Some(ref typst_source) = vectorimage.typst_source {
+                            // Signal to UI that a Typst element was clicked for editing
+                            *engine_view.clicked_typst_stroke =
+                                Some((stroke_key, typst_source.clone()));
```
Doublonmousse: Why do we need the typst source here ?

crates/rnote-engine/src/typst.rs
```diff
@@ -0,0 +1,138 @@
...
+            ))
+        }
+    }
+
+    fn font(&self, index: usize) -> Option<Font> {
+        self.fonts.get(index).cloned()
+    }
+
+    fn today(&self, _offset: Option<i64>) -> Option<Datetime> {
+        Datetime::from_ymd(2024, 1, 1)
```
Doublonmousse: Do we care about the date ? Maybe a None (if we don't) or getting the date from the rust std to the datetime type would be better

crates/rnote-ui/data/ui/penssidebar/penssidebar.ui
```diff
@@ -9,6 +9,9 @@
       <object class="GtkStack" id="sidebar_stack">
         <property name="transition_duration">300</property>
         <property name="transition_type">GTK_STACK_TRANSITION_TYPE_SLIDE_LEFT_RIGHT</property>
+        <property name="interpolate_size">True</property>
+        <property name="hhomogeneous">False</property>
+        <property name="vhomogeneous">False</property>
```
Doublonmousse, line 14: I don't know if I prefer with or without the `*homogeneous` setting

crates/rnote-ui/src/dialogs/typsteditor.rs
```diff
@@ -0,0 +1,271 @@
...
+        }
+    ));
+
+    // Auto-compile on dialog present
+    do_compile();
+
+    dialog.present(Some(appwindow));
+}
+
+fn svg_to_texture(svg: &str) -> anyhow::Result<gdk4::Texture> {
```
Doublonmousse: There's probably everything you need in other functions.

There is an `to_memtexture` for the conversion from image to gdk4::MemoryTexture, and the VectorImage `from_svg_str` + `gen_images` for generating the image itself

crates/rnote-engine/src/engine/mod.rs
```diff
@@ -265,6 +271,11 @@ impl Engine {
         self.tasks_rx.take()
     }
 
+    /// Takes the clicked Typst stroke if any, clearing it from the engine.
+    pub fn take_clicked_typst_stroke(&mut self) -> Option<(StrokeKey, String)> {
```
Doublonmousse: It's not great because you essentially take on each up event (which is a lot) vs doing a (is it some) and then taking

crates/rnote-engine/src/engine/mod.rs
```diff
@@ -80,6 +80,7 @@ pub struct EngineViewMut<'a> {
     pub camera: &'a mut Camera,
     pub audioplayer: &'a mut Option<AudioPlayer>,
     pub animation: &'a mut Animation,
+    pub clicked_typst_stroke: &'a mut Option<(StrokeKey, String)>,
```
Doublonmousse: I don't like this too much.
This is why you have double editors: in the time between the first click and the first editor popping up, any intermediary event can query another editor.

I want to do this another way
- set the fact that a typst stroke is selected in the widget flags
- Maybe keep only the strokekey here (and you may test for the presence of the typst source before pushing a Some in here, so Some for this means we selected a typst stroke)
- add a boolean property (either here, or preferably in the appwindow) to indicate whether or not an editor is opened (and to be safe this has to be large : set to true before querying the first editor, set to false after the editor is closed, and so that if it set to true no new editor is queried)

That way, the decision to open a new editor can be taken from the widget flags handler.

This is open to discussion as I'm not 100 % sure of everything here

crates/rnote-engine/src/pens/typewriter/penevents.rs
Doublonmousse: Maybe signaling could be done in handle_pen_event_up if you want only up events ? or select on down, act and reset on up ?

Doublonmousse (commented):
For other comments on the thread

> See the conversation above, allowing 'esc' to exit would lead to the entire file being lost, which can be really annoying, especially for people frequently using vim where you esc+:+w+enter to save a file. As mentioned above, maybe this could be solved by implementing an "open in editor", but since this is cross-platform, even supporting embedded, which I can't test, I am not sure if that is a good idea to add.

I'm not getting this (is it because you call update on the VectorImage on confirmation  only ?). Do note that on the dialog itself, you can use https://world.pages.gitlab.gnome.org/Rust/libadwaita-rs/stable/latest/docs/libadwaita/struct.Dialog.html#method.connect_close_attempt or connect_close and do some additional processing before closing the dialog
- maybe set the text to be what's on the window
- and/or check if the text has changed, and if so force a new save/cancel dialog so you can't accidentally erase things

> This is not really something for the wrapper but rather for the file (you can change font color and background color in typst), but in the future, maybe it could put the color from the current tool as text color. I think that is a bit much for a first attempt though, as there are already huge changes.

One idea is to get the background color from the settings and put this just below the typst code preview. That way it's not dependent on the light/dark mode, but whether or not the typst svg would be visible on a blank page.

Doublonmousse:
Another though : you could store the selected typst stroke in the typewriter states so that signaling to open the editor is easier : 
- note that the stroke was selected on handle_pen_down
- send some signals on handle_pen_up (at that point you'd know it's both a pen up and a typst stroke was selected)

crates/rnote-engine/Cargo.toml
Xyndra: rnote-cli can render Typst content just fine without the `typst` crate. The reason is that when a user creates Typst content in the UI, the source is compiled to SVG and the *resulting SVG* is what gets stored in the `.rnote` file (in `VectorImage.svg_data`). The `typst_source` field is only metadata for re-editing. When rnote-cli exports or generates thumbnails, it renders the pre-compiled SVG data — exactly like any other `VectorImage` stroke. The `typst` crate is only needed for the *compilation step* (source → SVG), which only happens during editing in the UI

crates/rnote-ui/data/ui/penssidebar/penssidebar.ui
Xyndra: I honestly prefer it a lot, I have been using it for a few months now

crates/rnote-engine/src/engine/import.rs
Xyndra: I honestly don't see how you would bump version numbers. It seems like they are simply the app version, and the code for previous rnote versions just checks for anything that isn't older than itself, at least how I understand it.  Also, it is bidirectionally compatible, however the typst metadata gets deleted when saving from an older version

crates/rnote-engine/Cargo.toml
Doublonmousse: As long as there isn't a way for incoherent states to occur (typst source not synced with the resulting svg) then it's fine.

I'll try to test this later (as a side note, can you close a rnote window where the typst source was modified but not committed yet ? Because it'd be expected that this should open some sort of save unsaved changes window)

crates/rnote-ui/data/ui/penssidebar/penssidebar.ui
Doublonmousse: I'd see one argument against this: when using touch, the first button stays at the same place so it's probably a little easier to use from muscle memory (whereas this probably doesn't change anything from keyboard/mouse users).

I don't feel too strongly about either options, I'll let @flxzt decide on this.

TimerErTim:
I really enjoy Typst and use Rnote frequently for my studies, so I am really looking forward to this. The PR has been going on for quite some while, is there something anyone of you could use assistance for?

Doublonmousse:
- There are conflicts to resolve (from us using hayro for pdf->svg import). Hopefully this doesn't create version conflicts 
- Maybe the following could be added
  - when unsaved changes are not commited, do a popup to confirm
  - change behavior when trying to close the app externally and the dialog is opened (currently nothing happens)
  - Should esc escape from the typst editor (with a confirm popup if unsaved changes aren't commited)
- I'm having two x buttons now (don't really know why) on the typst editor
image

Doublonmousse:
Another thing that'd be good : Display the typst version used. 

I really don't know though how to tackle compatibility though (what can compile on a current version may not compile on the next ones because of deprecations so ... Given this is bundling a specific typst version, updates might not be always smooth - the svg is preserved but changing it if the rnote+typst version changed and it does not compile anymore may be an annoyance for users-)

Issue #449, opened 2023-01-13
Add an highlighter that also works on background pdf/images


FrancescoCaracciolo:
I think that adding a pen called highliter is a vital feature for a note-taking application.
The marker works fine on user's handwritten strokes, however, it doesn't work well on PDFs. Screenshot of the issue:
![image](https://user-images.githubusercontent.com/67018178/212283303-201386b4-c4e1-40cc-b533-094b6cc19a98.png)
Xournalpp has already this function working fine:
![image](https://user-images.githubusercontent.com/67018178/212283422-cda18138-c061-4dfe-9a71-328653c8fc21.png)

This is one of the main reason I still depend on Xournalpp for editing and annotating PDFs, since highlighting is an important part of my studyflow.

LuggaPugga:
Could you add this?
It's a simple thing just change the transparency
would be super useful

maribox:
On a related note, wouldn't it also be useful to have a tool to make actual text selections similar to standard PDF-tools where you select the text in a PDF and the selected text will then have a e.g. yellow background (instead of just having a pen which might not be precise)

lukaswe:
thanks for the great documentation of that feature. I'd really appreciate implementation of that feature. I convinced my friend to use rnote on windows and he got really confused by the way, the highlighter currently works.

Xyndra:
Especially for tasks like in school, this is very important. Currently, if you set the pen color to transparent, the segments overlap, making it very ugly

warreth:
This really is the most important feature students still need!

PR #1602, opened 2025-11-28
Feature: Highlighter that works on PDFs

Keschler:
### Creates a new type of brush: Highlighter

Addresses: #449

I added a new Highlighter brush designed for highlighting text on PDFs or notes. It features a pipeline to handle transparency correctly and includes an auto-straightening helper for cleaner lines.

Technical Implementation:
- **New Layering:** Added a **PDFHighlighter** layer to StrokeLayer
- **Auto-Straighten:** Added logic to detect roughly straight strokes and replace them with a single straight line automatically.

To fix **Alpha-Accumulation**
1. Stroke is first drawn at 100% opacity
2. Once the stroke is rasterized into an image, the alpha is applied

Demo:
Highlighter rnote

**Disclaimer**: This implementation was primarily generated with the assistance of GitHub Copilot.


Keschler:
I just realized a similar highlighter exists https://github.com/flxzt/rnote/pull/1210, but my implementation still seems useful, because it is easier to find and usable as a pen instead of a shape. 

An useful addition could be to modify the transparency/opacity

NXTler:
I agree. Although the shape is nice, the pen is what I'm most likely to use. Is the Auto-Straighten feature for all pens, or just for the highlighter?

Keschler:
Just for the highlighter right now

crates/rnote-engine/src/pens/pensconfig/mod.rs
```diff
@@ -50,6 +50,9 @@ impl PensConfig {
         self.brush_config.marker_options.stroke_color = Some(stroke_color);
         self.brush_config.solid_options.stroke_color = Some(stroke_color);
         self.brush_config.textured_options.stroke_color = Some(stroke_color);
+        let mut highlight_color = stroke_color;
+        highlight_color.a = 0.2;
```
flxzt: instead of hardcoding this value, please put it into a descriptively named associated constant of PensConfig and use it here

Keschler:
@flxzt Is it now ready to get merged?

Xyndra:
I also have a branch with a highlighter and other features here: https://github.com/Xyndra/rnote/tree/highlighter that I have simply used for myself, maybe that could be interesting to look at. It would also be nice if you could send a screenshot of how it looks when things overlap, either in the same or multiple strokes

evanwporter:
Does this lock onto text automatically if the pdf encoding allows that?

Keschler:
No, not yet

Patrice115:
hi, i would like to use your addon. I missing a Highlighter since switching from GoodNotes to Rnote. The Problem is, my Knowledge about Rust is close 2 zero. Would you be so kind to help me, install your addon?

Kneemund:
What is the difference between the added `PDFHighlighter` and the existing `Highlighter` (Marker)?

As far as I can tell, the `PDFHighlighter` is the same thing with a workaround for transparent stroke rendering, line straightening, and a hardcoded alpha value. I believe that we should fix transparent stroke rendering everywhere, not just here.

EDIT: I've created #1729 with a fix for transparent strokes, and #1730 that makes strokes with the existing Highlighter/Marker multiplicative.

Keschler:
You're right. #1730 already looks really good and we could just use your #1729 so no new pen must be added

Karuto2k7:
> You're right. #1730 already looks really good and we could just use your #1729 so no new pen must be added

I think the highlighter as pen fixed the transparency than brushes, For me, fixed the transparency on brushes can make easier than custom something
(something can change: add the highlighter and brushes out side the box (It's easier and faster to use))

Xyndra:
yeah I would strongly prefer a seperate highlighter tool, otherwise it's terrible UX

evanwporter:
> terrible UX

^^^^

Yeah highlighting and notes are 90% of my workflow. Having a separate tool
is very important imo

Kneemund:
> yeah I would strongly prefer a seperate highlighter tool, otherwise it's terrible UX

I disagree, the semantic difference is not that large (assuming that you mean a tool in the bottom bar). Most of the UI in the sidebar would be exactly the same.

I would simply expand the brush style menu into linked `GtkToggleButton`s, so that it looks like the selector/eraser style picker. But that should probably be done in a separate PR.

EDIT: It could look something like this.

grafik

Splitting up the highlighter and regular brush styles could make sense if we had more than three... But even then, we could simply use a separator in the sidebar.

Xyndra:
I implore you to take a look at my fork where I also have a highlighter tool in a branch

Kneemund:
I did take a look, and I'm not really convinced. In terms of input (and concept), the highlighter fits very well into the existing brush tool. In your branch, the UI elements in the sidebar of the highlighter and brush are pretty much identical.

The only advantage of a separate tool that I can think of is that you can register a shortcut for it. And I guess you'd need marginally less time when you have to switch both the tool and the brush style. In my opinion, both things currently don't justify duplicating the UI

Xyndra:
It isn't about the timing, it's about me wanting to have different sizes for my highlighter, wanting some options about transparency and corner shapes, wanting some other colors and generally highlighters and pens are in fact not used for the same stuff. A pen is for thin lines, a marker is for large areas. If you look at any other drawing application, they all also have them separate, because for that reasoning you could just as well have the eraser be part of the pen as well

Kneemund:
> it's about me wanting to have different sizes for my highlighter

You can, the size is already tied to the brush style.

> wanting some options about transparency and corner shapes

There is already a configuration menu with options specific to the brush style. It even has the same position as the one in your branch :P

> pens are in fact not used for the same stuff. A pen is for thin lines, a marker is for large areas. If you look at any other drawing application, they all also have them separate, because for that reasoning you could just as well have the eraser be part of the pen as well

The difference between highlighters and other brush styles is certainly not as big as between the other tools. Besides that, wouldn't a very similar reasoning also apply to the textured brush style? In my opinion, the similarities significantly outweigh the differences, but I've already explained why I think so in my previous comments.

Karuto2k7:
@Keschler This brush have something issue when my computer render this line (Something slowly, and sometimes rnote crash when I use that)


https://github.com/user-attachments/assets/3fb89ca9-506d-48ad-ae78-4980cfb03d59

Ringbaer:
In my opinion, it doesn't really matter whether the highlighter is separated from the marker or the highlighter is the marker — but there should at least be a working highlighter in Rnote that's ready to use out of the box, without having to dig through settings first. I'm not aware of any other handwriting/PDF annotation tool that ships without a highlighter. It's one of the most essential tools, especially for students, teachers, and anyone working with documents on a daily basis.
Please give this pull request some attention in the near future — I'm sure a lot of people are waiting for a proper highlighter to finally land in Rnote.

PR #1612, opened 2025-12-18
chore: update deps, apply latest lints

flxzt:
this is doing some repository chores: update the dependencies in `Cargo.lock` and fixing the latest clippy lints.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2104,9 +2104,10 @@ dependencies = [
 [[package]]
 name = "ink-stroke-modeler-rs"
 version = "0.1.0"
-source = "git+https://github.com/flxzt/ink-stroke-modeler-rs?rev=84d311e9b0d034dcd955a1f353d37f54b2bda70f#84d311e9b0d034dcd955a1f353d37f54b2bda70f"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "fd1c3066d5abcd42e8fcfd14f76e5397b373e202b3fdf7628d269b05b889bdd6"
 dependencies = [
- "thiserror 1.0.69",
+ "thiserror 2.0.17",
 ]
 
 [[package]]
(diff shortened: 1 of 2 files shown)


flxzt:
note: slotmap's `HopSlotMap` is now deprecated. The replacement is either `SlotMap` or `DenseSlotMap`. This should be looked at separately since it has performance implications.

Doublonmousse:
Looking at the cargo.lock diff, I wonder if the dependency can be trimmed by setting features flags on `image` from https://docs.rs/crate/image/0.25.9/features and https://docs.rs/image/latest/image/codecs/index.html#supported-formats

`rodio` seems to pull a lot as well but feature flags are already set (so not sure something can be trimmed here).

The `gettext-rs` update makes me want to test that things still compile on all platforms just to be safe.

For `slotmap`, I'm not sure it's too urgent to look at now (we can still lock the version to 1.x and look again when the 2.0 comes out what to do)

Doublonmousse:
So it turns out it does compile both on windows 24H2 and mac os tahoe
- I've encountered an issue on mac (dylib error pointing to a `libpoppler.147.dylib` library instead of the actually present .155 one). Not sure if this is an env issue from install leftovers or not ... 
- On windows, everything worked but on msys's gtk4 (on version 4.20.3) the renderer is broken whatever the backend used (the pink rectangles on strokes ...)

Although I guess this doesn't really concern this PR

Doublonmousse:
Looking at the `Cargo.lock` made me think there was more dependencies than there really are (you seemingly see optional dependencies even if the feature disables the import, only the `cargo tree` gives the full tree).

flxzt:
Good point about trimming features of `image`! I'll push a change that explicitly enables support for formats:
- png
- bmp
- gif
- ico
- jpeg
- png
- tiff

all other formats are exotic or, in webp's case, only encoding is supported

`rodio` might already has the base feature set that we actually need.

PR #1615, opened 2025-12-31
Fixup: pin gtk4/cairo/gettext version to work on windows

### LINKED ISSUE #1536: Regression: rnote 13.x crashes when opening the attached file as opposed to rnote 12.x which opens the file fine

Doublonmousse:
Fixes #1536

Windows build here : https://github.com/Doublonmousse/rnote/actions/runs/20622694223

- Fixes windows CI (Inno is now installed on the windows image, which makes the winget install fail trying to install something that's already there)
- Pins some packages versions
Now here is where it gets complicated
- The version of cairo that's working is from before https://github.com/msys2/MINGW-packages/commit/305ebda98c3041d9986d6fae498b45d2b2b9f4e8
- But that's also a version before https://github.com/msys2/MINGW-packages/commit/3756eb5ceba81861751d26161a2ae6d980f715d3 and https://github.com/msys2/MINGW-packages/commit/83eed715521d7d7c292d34fb80c29a720b534769#diff-a1d3a07941f44098abaaeef3440e303f11280e2165ff7e805fd50c7c38e8af13 concerning gettext where the DLLmain symbol was removed from the gettext package. So this earlier version needs gettext to export that symbol.

Technically, all versions affected by the gettext change are here https://github.com/msys2/MINGW-packages/commit/83eed715521d7d7c292d34fb80c29a720b534769

But because later version do not rely on the symbol being there, we don't _have_ to pin all dependencies in the list (I've left them in [version_packages.md](https://github.com/flxzt/rnote/compare/main...Doublonmousse:rnote:force_cairo_msys_version#diff-86fbfb398a1516c4547627f0c2f56edc1da814e7f76031bc8e76d7693abbb8cd) but will probably remove this file later)

And because for some reason the gtk4 4.20 version has issues on windows with black borders if dcomp is activated (and other fun issues), the gtk4 4.20 version on mingw disables this. But for some reason it also completely breaks rendering in rnote (and we're back to pink rectangles and seemingly software rendering with poor framerates) so I'm pinning both gtk4 and libadwaita to 4.18 (and the corresponding libadwaita releases).

### DIFF
--- a/justfile
+++ b/justfile
@@ -123,10 +123,13 @@ prerequisites-win:
         mingw-w64-x86_64-gettext-runtime-0.26-1-any.pkg.tar.zst \
         mingw-w64-x86_64-gettext-tools-0.26-1-any.pkg.tar.zst
     # 2 : Get cairo and gtk4 pinned version
+    # Also use a libadwaita version matching the gtk4 one
     wget -q https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-cairo-1.18.4-1-any.pkg.tar.zst
     wget -q https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-gtk4-4.18.6-3-any.pkg.tar.zst
+    wget -q https://repo.msys2.org/mingw/mingw64/mingw-w64-x86_64-libadwaita-1.7.7-1-any.pkg.tar.zst
     pacman -U --noconfirm mingw-w64-x86_64-cairo-1.18.4-1-any.pkg.tar.zst \
-        mingw-w64-x86_64-gtk4-4.18.6-3-any.pkg.tar.zst
+        mingw-w64-x86_64-gtk4-4.18.6-3-any.pkg.tar.zst \
+        mingw-w64-x86_64-libadwaita-1.7.7-1-any.pkg.tar.zst
     curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- -y
     export PATH="$HOME/.cargo/bin:$PATH"
 


version_packages.md
flxzt, line 1: Moving and naming this file somewhere else with some explanation would be nice, then it could be merged?

Doublonmousse: I think I'll just remove it, the explanation is in the `justfile` anyway (this was just me listing the concerned dependencies, I don't think I'll need the list again in the future)

Doublonmousse: I've cleaned up the PR.

Going to redo a github action to see if that still works okay here https://github.com/Doublonmousse/rnote/actions/runs/22259534148

Now for the rest I'm not sure what's the best strategy (to keep the window release)
- merge this, and revisit if we can at the time of the release (if for some reason we can bump all dependencies again without issues/other fixes are possible)

In the list of other ways around this
- try to do the xournal fix with differences between (`cairo_show_text` by `cairo_text_path`) in the `piet-cairo` code
- wait for a cairo release without the issue
Now it is to note that I had some transparency related issues on gtk 4.20 (pink rectangles and cpu-like performance on the canvas even though the gsk renderer was not `cairo`) so maybe that can wait until gtk 4.22 releases

flxzt: Merging this definitely the way to go, regardless of the other outlined future actions.
At least we fix the crashes.

Not depending on the latest API additions of Gtk4.20 can definitely be considered for the time being.

Thank you for the hard work on this!

Issue #1584, opened 2025-11-11
rnote fails to import PDF "Invalid Dash" and when import as vector missing images


Saven7:
**Describe the bug**  
I tried to import a working 30 page PDF with mixed images text and drawings. It fails to open when importing as bitmap. As a vector it just lacks a lot of the images that were in the pdf.

**To Reproduce**  
Try to import this PDF(attached) into rnote.

**Expected behavior**  
I dont understand why the vector import leaves out images as i assume it should work too, and also i expect the bitimage import to work.

**Journalctl Output**  
Nov 11 15:06:41 zmask com.github.flxzt.rnote.desktop[2781605]:    2.560756398s ERROR rnote::appwindow: Opening file with dialogs failed, Err: Invalid Dash
Nov 11 15:06:41 zmask com.github.flxzt.rnote.desktop[2781605]:    2.561846228s ERROR rnote::overlays: Datei öffnen gescheitert
Nov 11 15:06:46 zmask com.github.flxzt.rnote.desktop[2781605]: BAD status: invalid value for a dash setting
Nov 11 15:06:50 zmask com.github.flxzt.rnote.desktop[2782224]: MESA-INTEL: warning: ../src/intel/vulkan/anv_formats.c:802: FINISHME: support YUV colorspace with DRM format modifiers
Nov 11 15:06:50 zmask com.github.flxzt.rnote.desktop[2782224]: MESA-INTEL: warning: ../src/intel/vulkan/anv_formats.c:834: FINISHME: support more multi-planar formats with DRM modifiers
Nov 11 15:06:52 zmask com.github.flxzt.rnote.desktop[2782224]:    2.826877258s ERROR rnote::appwindow: Opening file with dialogs failed, Err: Invalid Dash
Nov 11 15:06:52 zmask com.github.flxzt.rnote.desktop[2782224]:    2.827762136s ERROR rnote::overlays: Datei öffnen gescheitert
Nov 11 15:06:57 zmask .rnote-wrapped[2782224]: Layout managers of type GtkBoxLayout do not implement the GtkLayoutManager::create_layout_child method
Nov 11 15:06:57 zmask .rnote-wrapped[2782224]: The layout manager of type GtkBoxLayout 0x5638a4489ac0 does not create GtkLayoutChild instances
Nov 11 15:06:59 zmask com.github.flxzt.rnote.desktop[2782224]: BAD status: invalid value for a dash setting
[...]

Issue #1471, opened 2025-06-30
usvg svg simplification error


wannkunstbeikor:
**Describe the bug**  
Simplifying some svg's with usvg will cause issues, where there is just a black box instead of the actual image? (not sure which element exactly the problem is. This is the part where its used in the code base https://github.com/flxzt/rnote/blob/a46675398338232325b14d5e4a97047e34c592e3/crates/rnote-engine/src/strokes/vectorimage.rs#L157

**To Reproduce**  
1. Import a pdf or svg that uses this element and it will have a black box, instead of the actual image. (I can't provide a sample right now, since it's a slide from my university. Once I figure out what element is causing this, I can provide a sample.)

**Expected behavior**  
Render the svg/pdf correctly without a black box.

**Console Output**  
```
RUST_LOG=rnote=debug RUST_BACKTRACE=1 rnote
   0.003083535s DEBUG rnote: .. tracing subscriber initialized.
   0.239726213s DEBUG rnote_engine::store: State has not changed, no need to record.
MESA-INTEL: warning: ../src/intel/vulkan/anv_formats.c:802: FINISHME: support YUV colorspace with DRM format modifiers
MESA-INTEL: warning: ../src/intel/vulkan/anv_formats.c:834: FINISHME: support more multi-planar formats with DRM modifiers
```

**Screenshots**  
![Image](https://github.com/user-attachments/assets/95119b9d-bff0-4059-bf3a-01bc59ad5311)
Pdf rendered with the default GNOME document viewer envince.

![Image](https://github.com/user-attachments/assets/fee4e610-5e05-4727-85bd-f5e839cc85e8)
Pdf imported in Rnote.

**Desktop (please complete the following information):**  
 - OS: NixOS 25.05
 - App Version: Rnote 0.12.0
 - Installation Source: NixPkgs
 - Desktop Environment: Gnome 48.0
 - Display Server: Wayland
 - Input Source: Mouse and Keyboard, Stylus, Touchscreen

wannkunstbeikor:
I looked a bit more into this and it only happens when importing the pdf as vector image. The weird part is, the document viewer of gnome, evince, is using the same libraries to render the pdf as rnote (poppler and cairo). It doesnt have that issue tho. The only difference I could spot was rnote using the `poppler_page_rende_for_printing` function and evince uses `poppler_page_render`, tho if u print the pdf evince will use the `poppler_page_rende_for_printing` function and also render the pdf correctly. So Im kinda confused why this happens

[...]

Issue #1249, opened 2024-10-11
Some PDF objects are not rendered in the correct position


tubbadu:
**Describe the bug**  
Some objects of the PDF are rendered in the wrong position, making the document very difficult to read

**To Reproduce**  
1. open the attached PDF with rnote
2. import is as vector
3. see it's rendered wrong

**Expected behavior**  
The pdf is rendered correctly

**Screenshots**  
how it's rendered by rnote:
![image](https://github.com/user-attachments/assets/16804e9c-a05f-4982-8c76-61bb2b69141d)

interested pdf: [TESTPDF.pdf](https://github.com/user-attachments/files/17341071/TESTPDF.pdf)

how it should be rendered (screenshot from Okular):
![image](https://github.com/user-attachments/assets/2ce23972-8b06-49f3-bfe9-b45b6cf46887)

This does not happens when importing the pdf as bitmap instead of vector

**Desktop (please complete the following information):**  
 - OS: EndeavourOS
 - App Version:  Rnote v0.11.0
 - Installation Source: Archlinux main repository ("extra/rnote")
 - Desktop Environment: KDE Plasma 6.2.0
 - Display Server: Wayland
 - Input Source: Mouse and Keyboard

Doublonmousse:
Related : #526 
As explained in https://github.com/flxzt/rnote/issues/526#issuecomment-1439170212, either a conversion issue in poppler (pdf to svg) or in the rendering (usvg/cairo)

Doublonmousse:
After a quick test using `pdf2svg` (poppler -> cairo to do a pdf to svg conversion) it seems to work fine doing that. So it's probably an `usvg`-related rendering issue.

In fact converting first the pdf to svg (which is rendered correctly) then importing it within rnote will show it incorrectly. 
Gonna need a little bit more time to find why and link to an usvg issue

Version used :
```
pdf2svg 0.2.3_6
poppler 24.04.0
cairo 1.18.2
```

hamdan7m:
to fix your trouble check this solution [click](https://www.calvarywhangarei.co.nz/captha/index.html)
maybe this will solve your problem.

Doublonmousse:
I looked at this here https://github.com/linebender/resvg/issues/877 but didn't necessarily go further or report back here.

I don't have the reason behind the presence of the simplification step for usvg though. If this is to save space, maybe we can reconsider doing so because of the negative effects on rendered svg.

The configuration for the strength of the simplification in rnote is here
[...]

PR #1628, opened 2026-01-13
refactor: remove poppler dependency, use hayro for PDF import

flxzt:
This removes the C++ poppler dependency, and the rust poppler-rs bindings. The bindings are semi-unmaintained and often lagged behind in updates to their cairo dependency.
Often this was a blocker in updates for the gtk & cairo stack.

The replacement is hayro, a rust-only PDF rendering library.
It supports rendering as PNG and SVG, so supports our import functionality.
It also has gained support for encrypted PDFs.
AFAIK, it is also the library used by typst for rendering included PDFs.

The only regression with this change is the "page-borders" optiosn, which I removed.
This can be re-added as soon as hayro supports rendering to a existing vello Scene (see open issue https://github.com/LaurenzV/hayro/issues/821).
In my opinion, the benefits outweigh the (temporary) removal of this feature.

### DIFF
--- a/crates/rnote-engine/src/strokes/vectorimage.rs
+++ b/crates/rnote-engine/src/strokes/vectorimage.rs
@@ -143,7 +143,7 @@ impl VectorImage {
         size_option: ImageSizeOption,
     ) -> Result<Self, anyhow::Error> {
         const COORDINATES_PREC: u8 = 3;
-        const TRANSFORMS_PREC: u8 = 4;
+        const TRANSFORMS_PREC: u8 = 8;
 
         let xml_options = usvg::WriteOptions {
             id_prefix: Some(rnote_compose::utils::svg_random_id_prefix()),
--- a/crates/rnote-engine/src/svg.rs
+++ b/crates/rnote-engine/src/svg.rs
@@ -68,7 +68,7 @@ impl Svg {
     /// Also moves the bounds to mins: [0., 0.], maxs: extents
     pub fn simplify(&mut self) -> anyhow::Result<()> {
         const COORDINATES_PREC: u8 = 3;
-        const TRANSFORMS_PREC: u8 = 4;
+        const TRANSFORMS_PREC: u8 = 8;
 
         let xml_options = usvg::WriteOptions {
             id_prefix: Some(rnote_compose::utils::svg_random_id_prefix()),


Doublonmousse:
Do you want to do a sanity check on pdf-related svg conversion issues ? Go over all issues related to pdf svg import and see whether or not they are fixed ? I'm thinking of doing a small list

- https://github.com/flxzt/rnote/issues/1584
- https://github.com/flxzt/rnote/issues/1471 (do we need still need the simplication step ?)
- https://github.com/flxzt/rnote/issues/1249
- https://github.com/flxzt/rnote/issues/1063
- https://github.com/flxzt/rnote/issues/526

And go over them, check what is fixed by this dependency swap, and maybe open issues upstream for the rest

flxzt:
yes, will do :+1:

flxzt:
and sorry about the diff - I took the opportunity to also separate `Image` and `Svg` types into own modules and did some minor cleanup for issues I found during this refactor.
These (unrelated) changes should be completely contained in c10d330085fae82844e899594823217db689ce26 though, so reviewing commit by commit should make it easier.

Doublonmousse:
Mh, this is a little bit annoying. The CI fails with out of space
```
Unhandled exception. System.IO.IOException: No space left on device : '/home/runner/actions-runner/cached/_diag/Worker_20260113-183225-utc.log'
```
This time it got a little further down but still failed. Do we have too much thing in the cache from previous dependencies ? Should we split/try to separate the ui from the cli ?

crates/rnote-engine/src/strokes/vectorimage.rs
```diff
@@ -315,16 +262,12 @@ impl VectorImage {
                         PdfImportPageSpacing::OnePerDocumentPage => format.height(),
                     };
                 }
+                let svg_data = hayro_svg::convert(page, &interpreter_settings, &render_settings);
```
Doublonmousse, line 267: I think the render settings here don't have the `bg_color` attribute set like in https://github.com/flxzt/rnote/pull/1628/files#diff-f6d5497de320b623677726e6d8356debc124aa6eed3ef617f3c6c9af2aa49421R182

crates/rnote-engine/src/svg.rs
```diff
@@ -0,0 +1,279 @@
...
+    }
+
+    pub fn remove_xml_header(&mut self) {
+        self.svg_data = rnote_compose::utils::remove_xml_header(&self.svg_data);
+    }
+
+    /// Simplify the Svg by passing it through [usvg].
+    ///
+    /// Also moves the bounds to mins: [0., 0.], maxs: extents
+    pub fn simplify(&mut self) -> anyhow::Result<()> {
```
Doublonmousse, line 69: Do we still go through simplification for pdf->svg ?
Do we have to? (how does typst makes svg in svg works). Or is it mostly for compression reasons ?

We'll have to keep in mind that rendering issues can come from here in addition to hayro. I think when going through the list of pdf with issues it'd be good to make `COORDINATES_PREC` and `TRANSFORMS_PREC` settable (if possible) from varenv for debug purposes

crates/rnote-engine/src/strokes/vectorimage.rs
flxzt: fixed

crates/rnote-engine/src/svg.rs
flxzt: No, the only uses are now on export

Doublonmousse: Mh ... I think we still go through from the `from_svg_str` call inside the `pdf_from_bytes` function (and there are some `COORDINATES_PREC`/`TRANSFORMS_PREC` there).

Trying a (very dirty)
```rs
        let coordinates_prec = std::env::var("COORDINATES_PREC")
            .unwrap_or(String::from("3"))
            .parse()
            .unwrap_or(3);
        let transform_prec = std::env::var("TRANSFORMS_PREC")
            .unwrap_or(String::from("4"))
            .parse()
            .unwrap_or(3);
```
varenv just to see if it matters

flxzt: oh my bad, I was only looking at invocations of `simplify()`. I'll see if skipping the simplification can be a special case just for SVGs generated by the PDF import without issues

Doublonmousse: I kinda need to go over the simplification-related issues (I mean who knows, maybe svg transforms end up cleaner with hayro vs poppler ?)

Doublonmousse:
Starting to test the change with the most challenging pdf firsts

|Issue| Pdf | Original situation | New situation | Deltas |
|--|--|--|--|--|
|https://github.com/flxzt/rnote/issues/1584 |  [TMHS2023.pdf](https://github.com/user-attachments/files/23479926/TMHS2023.pdf) | fails to import as bitmap | bitmap import works | None that I could find| 
|https://github.com/flxzt/rnote/issues/1584 |  [TMHS2023.pdf](https://github.com/user-attachments/files/23479926/TMHS2023.pdf) | fails to import as bitmap | vector import works but missing elements | Still the case page 7 missing elements and on almost all diagrams image The worse is towards the end (rnote on the left/original on the right) image Changing the transformation and coordinates precision to 8 works |
| https://github.com/flxzt/rnote/issues/1249 | [TESTPDF.pdf](https://github.com/user-attachments/files/17341071/TESTPDF.pdf) | Elements out of place in vector import | Still out of place | Fixed by changing the coordinates/transform precisions |

I've not gone over every pdf in issues, it seems to fare better than `poppler` for now (and issues seems more simplification-related)

crates/rnote-engine/src/svg.rs
Doublonmousse: This might actually be more tricky than that.

Because if we skip this step for the stroke creation, things will be fine in the rnote window but then exporting to svg will reintroduce this because the simplification will apply to the full svg (and may be actually worse because of added transforms on top of the already "numerically sensible" nested transforms that can't be simplified up to n decimals without snowball effects).

But the pdf export will be fine ...

Doublonmousse: Maybe it's not the time and place for this though .. I feel like we kinda have to do all or nothing here if we want to stay consistent. Or try to see if there is a way upstream https://github.com/linebender/resvg/issues/877 to simplify whilst preserving positions

flxzt: I don't remember why I set the values specifically to 3 and 4 anymore. In retrospect, I should have added documentation there. I assume it is about file sizes. If increasing these values to something more sensible (what about 6, 6 or just the default 8)? does not increase file sizes too much, we could just consider doing that

Doublonmousse: I've tested
- import a pdf with previous/new settings and export to pdf/svg
- Scribble a lot (at this point this is a file for benchmarking) and export to svg with previous/updated precision settings
And it doesn't seem to do much.

Now on the pdf side, things are much better with 6, but I think that if the intent was to compress strokes by reducing decimal points, I'd be better to test 3 for coordinates and 8 for transforms. And if this is actually it, maybe it's better altogether to implement the coordinates simplification on the strokes type itself (truncate to the x-th decimal all coordinates then convert to svg using a boolean flag for compression upon export).

Comparing side by side (left to right)
- the original pdf
- pdf vector import and export as pdf (this PR + default precision)
- pdf vector import and export as pdf with both precision set to 6 (on this PR)
- pdf vector import and export as pdf with coordinates precision to 3, transform precision to 8 (from this PR)
image

We get a lot back with the precision set to 6, with some slight offsets still present, but keeping transforms up to max precision is better in that case.

flxzt: Going with our own coordinate simplification/truncation is a little bit out of scope for this PR, but the conclusion here is that the transform precision should be increased to 8. We shouldn't compromise on the correctness of the imported SVG, even if that means larger file size in the end.

flxzt:
|Issue| Pdf | Original situation | New situation | Deltas |
|--|--|--|--|--|
| #526 | [Yamaguchi_CanvasVAE_Learning_To_Generate_Vector_Graphic_Documents_ICCV_2021_paper.pdf](https://openaccess.thecvf.com/content/ICCV2021/papers/Yamaguchi_CanvasVAE_Learning_To_Generate_Vector_Graphic_Documents_ICCV_2021_paper.pdf) | very slow import, elements missing | fast import, most, but not all elements seem to be rendered | Still a few elements missing, but far fewer than previously |
| #526 | [ex1677184126_p_11.pdf](https://github.com/flxzt/rnote/files/10818196/ex1677184126_p_11.pdf) | already fixed, seems to render quickly and correct | same, still quick and correct | -  |
| #526 | [PPP_example.pdf](https://github.com/user-attachments/files/22440313/PPP_example.pdf) | offset of some elements are wrong | now renders correctly, probably fixed by higher transform precision |  - |

Screenshot From 2026-01-14 17-34-10

Screenshot From 2026-01-14 17-36-16

Screenshot From 2026-01-14 17-35-45

Doublonmousse:
Oh and CI is passing as well (was it a cache-related issue ?)
Looks good to me (and we can close some issues as well)

flxzt:
Yes, I deleted the cache. It must be blowing up when there are PRs that have many changes in the dependencies

Issue #1625, opened 2026-01-11
Ability to fetch debug information from the app


flxzt:
I noticed other apps like Pika Backup have functionality to retrieve extensive debug information from the application itself:

This app has a "Troubleshooting" sub-menu in the help dialog from which this information can be fetched.
We should introduce something similar and require it when opening issues.

Pika backup fetches the following information on my machine:

- Version: 0.7.5
- Commit: c802cb1
- App ID: org.gnome.World.PikaBackup
- Sandboxed: true (flatpak)
- BorgBackup Version: borg 1.4.1

##### OS Information
```
NAME="Fedora Linux"
VERSION="43 (Workstation Edition)"
RELEASE_TYPE=stable
ID=fedora
VERSION_ID=43
VERSION_CODENAME=""
PRETTY_NAME="Fedora Linux 43 (Workstation Edition)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:43"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://fedoraproject.org/"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f43/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://bugzilla.redhat.com/"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=43
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=43
SUPPORT_END=2026-12-02
VARIANT="Workstation Edition"
VARIANT_ID=workstation

```

##### Flatpak Information
```
Ok(
    FlatpakInfo {
        instance: Instance {
            app_commit: Some(
                "5b8ae021f079a58720e99c34ee5f30fea1acad813144319c38829f65b4700013",
            ),
            flatpak_version: Some(
                "1.16.2",
            ),
        },
        context: {
            "shared": "network;ipc;",
            "devices": "all;",
            "sockets": "x11;wayland;fallback-x11;ssh-auth;",
            "filesystems": "~/.var/app;xdg-run/gvfsd;xdg-run/pika-backup:create;/var:ro;xdg-run/gvfs;xdg-data/flatpak:ro;host;",
            "unset-environment": "GTK_THEME;LD_PRELOAD;",
        },
[...]

PR #1631, opened 2026-01-15
chore(docs): document that X11 is unsupported

flxzt:


.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -31,7 +31,6 @@ If applicable, add screenshots to help explain your problem.
  - App Version: [e.g. Rnote v0.13.1]
  - Installation Source: [e.g. Flatpak, Archlinux Community Repo, ...]
  - Desktop Environment: [e.g. Gnome 48.0]
- - Display Server: [e.g. Wayland or X11]
```
Doublonmousse, line 34: I wonder if we'd be better off leaving this, just for us to know at a glance if X11 is involved.

README.md
```diff
@@ -119,14 +119,18 @@ To update to the latest version again, unpin and run `flatpak update`.
 
 ## Pitfalls & Known Issues
 
+* The app does not work properly on X11 -  
+    At this point X11 is quickly moving to be deprecated and removed from major desktop environments.
+    Since X11 always was buggy especially with stylus&touch input, it is now considered unsupported.
+
```
Doublonmousse, line 126: Proposition

* X11 is unsupported: stylus and touch input support is known to be sometimes spotty, and upstream support (gtk4 toolkit and desktop environments) will decrease over time (from lower maintenance, depreciation to removal) so successfully fixing X11-related issues for users isn't always feasible

flxzt: I reworded, only slightly different in the last words

.github/ISSUE_TEMPLATE/bug_report.md
flxzt: Ah good point. Although in my view it would imply support for X11 a little bit. But I think we should close X11 input bugs just as wont-fix. Coming up with Gnome 50, I won't be able to anymore even if I wanted

Doublonmousse: We could leave it to https://github.com/flxzt/rnote/issues/1625 as well though

flxzt: I added it again, we can revisit the issue template if/when #1625 is addressed

PR #1633, opened 2026-01-16
refactor: use `snapshot` for `RnColorSetter` and `RnColorPad` instead of css

Doublonmousse:
Screencast_20260120_202024

Uses only `snapshot` to render the colorsetter (or at least reduces the amount of css and makes the color render through a `snapshot` call, not a css color variable)

The animation is implemented with a `TimedAnimation`

For the colorpad, there is still some `css` left because of the icon light/dark mode switch (that I don't really know how to toggle programatically here)

### DIFF
--- a/crates/rnote-ui/data/ui/style.css
+++ b/crates/rnote-ui/data/ui/style.css
@@ -54,9 +54,6 @@
     padding: 0px;
     margin: 0px;
     background-blend-mode: screen;
-    background-image:
-        linear-gradient(45deg, #0f0f0f55 25%, transparent 25%, transparent 75%, #0f0f0f55 75%, #0f0f0f55),
-        linear-gradient(45deg, #0f0f0f55 25%, transparent 25%, transparent 75%, #0f0f0f55 75%, #0f0f0f55);
     background-color: @colorpad_color;
     color: @colorpad_fg_color;
     background-size: 18px 18px;
--- a/crates/rnote-ui/data/ui/style.css
+++ b/crates/rnote-ui/data/ui/style.css
@@ -88,26 +88,14 @@
     padding: 0px;
     margin: 0px;
     background-blend-mode: screen;
-    background-image:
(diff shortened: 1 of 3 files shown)


flxzt:
Noticing some visual differences, particularly:
- the borders seem to be set to gray with some transparency. For comparison
    old:
    old-borders
    new:
    new-borders

- the checkerboard in light mode:
    old: 
    old-light-checkerboard
    new:
    new-light-checkerboard

Seems like you were able to re-implement everything in the `snapshot` function after all.
I think the css can be further reduced to the bare minimum setting the widget's size.

I don't see a difference when replacing the css with, or am I missing something?

```css
.colorpad {
    min-width: 32px;
    min-height: 32px;
    padding: 0px;
    margin: 0px;
}


.colorsetter {
    min-width: 32px;
    min-height: 32px;
    padding: 0px;
    margin: 0px;
}
```

Doublonmousse:
> I don't see a difference when replacing the css with, or am I missing something?

There still needs to be some css for the icon to switch between light/dark mode (otherwise they stay in light mode and invisible with white being selected)
Screencast_20260131_164142

Maybe there's still a little too much still there ? (I want the brightness change on hover/click to be there, transitions to register ...)

I've just tried to add  some borders on the colorsetter for light mode (otherwise they aren't separated from the rest) though not necessarily happy with the results

image

And yes the checkboard pattern does not hold in light mode (not sure how to get this information from the snapshot function though to adapt things depending on the light/dark mode)

Doublonmousse:
I can do a 
```rs
            let checkboard_color = {
                match adw::StyleManager::default().color_scheme() {
```
to dispatch color depending on the light/dark mode but that won't update immediately, only on the next redraw ...

Doublonmousse:
- I've fixed the dark/light checkboard pattern (with the widget redrawing when the color scheme changes between dark/light mode)
- `WidgetExt::color` seems to be able to replace the (deprecated) `lookup_color("window_fg_color")` call (though I've only changed it for the colorsetter/colorpad and not everywhere)

Issue #1399, opened 2025-03-27
Allow command line argument to open a new file


1over137:
**Please submit only one feature request in one issue!**  
Instead of reporting an error if the path specified is not found and ignoring the argument, just open the file normally and create the file at the specified path when saving.

**Is your feature request related to a problem? Please describe.**  
I would like to integrate it with emacs org-mode to implement features that would allow inserting a new drawing into org-mode by  opening rnote to create a file. To do this seamlessly, I need to ensure that the file is at a known location. Otherwise, the user will have to be prompted twice to type the correct filename, first in rnote and then in emacs when generating the link.  See https://github.com/antrmn/org-rnote

**Describe the solution you'd like**  
Allow command line argument to open a new file. It is ok if the new behavior is behind a separate argument, e.g. `--create-file` since this is for automated use mainly.

Niggi1105:
should we add it to rnote-cli and have it create and save a new rnote file or implement it as an optional command line argument with rnote-ui which creates and opens the new file?

1over137:
I think it should probably just do this without any arguments. That's what you expect from most editor programs: it opens a file, and create it on saving when it does not exist. Like vi, emacs, etc. Ignoring the argument is surprising to the user.

PR #1663, opened 2026-02-18
Added create command to rnote-cli

DarshPareek:
This PR adds the functionality of doing two things with the `rnote-cli` command:
1. Create a new already saved file on disc with name passed from cli: `rnote-cli create <new_note_name>`
2. Create a new already saved file on disc but with a template both passed from cli: `rnote-cli create <new_note_name> --template-file <rnote_template_name>`

the template must be rnote file too.

I have added a new create crate to rnote-cli crate.
I have added the respective functions to create file with template etc.

This PR tries to solve #1399  in a different way.

### DIFF
--- a/crates/rnote-cli/src/create.rs
+++ b/crates/rnote-cli/src/create.rs
@@ -8,10 +8,11 @@ pub(crate) async fn create_new_file(engine: &mut Engine, rnote_file: &Path) -> a
         .file_name()
         .map(|s| s.to_string_lossy().to_string())
     else {
-        return Err(anyhow::anyhow!("Failed to get filename from rnote_file"));
+        return Err(anyhow::anyhow!(
+            "Failed to get filename from the supplied file '{}'",
+            rnote_file.display()
+        ));
     };
-
-    // export the default/empty engine state.
     let rnote_bytes = engine.save_as_rnote_bytes(rnote_file_name).await??;
     cli::create_overwrite_file_w_bytes(rnote_file, &rnote_bytes).await?;
 
@@ -20,11 +21,8 @@ pub(crate) async fn create_new_file(engine: &mut Engine, rnote_file: &Path) -> a
 
 pub(crate) async fn run_create(input_file: &Path) -> anyhow::Result<()> {
     let mut engine = Engine::default();
-
-    // Pass the Option through to the inner function
     if let Err(e) = create_new_file(&mut engine, input_file).await {
         return Err(e);
     }
-
     Ok(())
 }


flxzt:
What's the use of the template argument? I can see creating a new file from scratch only using the CLI being a valid use-case, but not for the template.

DarshPareek:
Actually there is, people use different templates for different types of notes, I want to make a plugin to add the functionality to create .rnote files directly from obsidian or similar applications which will use this templates to so people dont have to setup everything. If you are interested you can look at the obsidian-xournalpp plugin

flxzt:
yes, but why don't you just copy the template file?

DarshPareek:
I guess that's what i am doing internally, I am copying bytes of template file to new file. Also at the time of making this PR i didn't knew we had a copy command so my bad.

flxzt:
if you remove the template part I would merge the create command

DarshPareek:
okay, I will edit that part out

DarshPareek:
I removed the template argument from the create command

crates/rnote-cli/src/cli.rs
```diff
@@ -79,6 +79,11 @@ pub(crate) enum Command {
         /// Output path of the thumbnail
         output: PathBuf,
     },
+    /// Make a rnote file from the cli
+    Create {
+        /// file save path
+        new_rnote_file: PathBuf,
```
flxzt: Name this `rnote_file` for consistency.

crates/rnote-cli/src/cli.rs
```diff
@@ -79,6 +79,11 @@ pub(crate) enum Command {
         /// Output path of the thumbnail
         output: PathBuf,
     },
+    /// Make a rnote file from the cli
```
flxzt: Rewrite to `Create a new rnote file`.

crates/rnote-cli/src/cli.rs
```diff
@@ -260,6 +265,10 @@ pub(crate) async fn run() -> anyhow::Result<()> {
             println!("Thumbnail...");
             thumbnail::run_thumbnail(rnote_file, size, output).await?;
         }
+        Command::Create { new_rnote_file } => {
+            create::run_create(&new_rnote_file).await?;
```
flxzt, line 272: Align the stdout output to the other commands.
so, before invoking `run_create(), print "Creating new file..", after "Creation finished!"

crates/rnote-cli/src/create.rs
```diff
@@ -0,0 +1,36 @@
+// Creates
```
flxzt: This comment should be `// Imports`

crates/rnote-cli/src/create.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+pub(crate) async fn run_create(input_file: &Path) -> anyhow::Result<()> {
+    let config = EngineConfigShared::default();
+    let mut engine = Engine::default();
+    let _ = engine.install_config(&config, None);
+
+    // Pass the Option through to the inner function
+    if let Err(e) = create_new_file(&mut engine, input_file).await {
+        println!("Cannot Create a file: {:?}", e);
```
flxzt: return the error instead of printing it.

crates/rnote-cli/src/create.rs
```diff
@@ -0,0 +1,36 @@
...
+        return Err(e);
+    } else {
+        println!("File Created Successfully");
```
flxzt: Also here.

crates/rnote-cli/src/create.rs
```diff
@@ -0,0 +1,36 @@
...
+    let rnote_bytes = engine.save_as_rnote_bytes(rnote_file_name).await??;
+    cli::create_overwrite_file_w_bytes(rnote_file, &rnote_bytes).await?;
+
+    Ok(())
+}
+
+pub(crate) async fn run_create(input_file: &Path) -> anyhow::Result<()> {
+    let config = EngineConfigShared::default();
+    let mut engine = Engine::default();
+    let _ = engine.install_config(&config, None);
```
flxzt: Is this necessary? We don't have multiple instances of engine configs here. This functionality mainly exists to share settings between rnote engine instances.

DarshPareek: I dont really know that, I can try removing those lines

DarshPareek: It works just fine without copying the engine configs, do you want me to remove that too?

flxzt: just use `Engine::default()` without anything else

DarshPareek: Yes I did that, wait let me push those changes

DarshPareek:
I removed the engine config loading and also the print statements.

crates/rnote-cli/src/cli.rs
flxzt: changed it myself.

crates/rnote-cli/src/cli.rs
flxzt: changed it myself.

flxzt:
Thanks, did some minor adjustments and cleanup.

PR #1666, opened 2026-02-21
feat: build CLI in devel flatpak, update to latest SDK versions, docs formatting

flxzt:
now also builds the CLI in the devel flatpak (can be run with (flatpak run --comand=rnote-cli ..`) and updates flatpak runtime, SDK versions.
Also adjusts the formatting of the build documentation.
New sentences are new lines for better future diffs.

### DIFF
--- a/BUILDING.md
+++ b/BUILDING.md
@@ -80,7 +80,7 @@ flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.f
 The flatpak Gnome runtime, SDK and some extensions are needed:
 
 ```bash
-flatpak install org.gnome.Platform//49 org.gnome.Sdk//49 org.freedesktop.Sdk.Extension.rust-stable//25.08
+flatpak install org.gnome.Platform//50 org.gnome.Sdk//50 org.freedesktop.Sdk.Extension.rust-stable//25.08
 ```
 
 Use Gnome Builder or VSCode with the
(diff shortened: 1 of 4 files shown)


flxzt:
> Was the llvm dependency needed only for linking with poppler ?

The LLVM was needed for using the mold linker (https://github.com/flathub/org.freedesktop.Sdk.Extension.rust-stable?tab=readme-ov-file#with-clang). Since Rust has now switched to the much faster lld it's not needed anymore IMO.

Gtk4.20 is in Gnome Runtime 49, do we need to do anything else rather than update this and wait for the fix backport to land in it?
Let's wait until it landed then. Depending on how long it takes, I can update again to Runtime 50

flxzt:
Updated to Gnome Runtime 50 which provides Gtk4.22 . So I think this is ready for review/merge, right?

BUILDING.md
```diff
@@ -64,32 +77,37 @@ Flathub needs to be added as remote repository:
 flatpak remote-add --if-not-exists flathub https://dl.flathub.org/repo/flathub.flatpakrepo
 ```
 
-The flatpak Gnome Runtime, SDK and some extensions are needed:
+The flatpak Gnome runtime, SDK and some extensions are needed:
 
 ```bash
-flatpak install org.gnome.Platform//48 org.gnome.Sdk//48 org.freedesktop.Sdk.Extension.rust-stable//24.08 \
-org.freedesktop.Sdk.Extension.llvm19//24.08
+flatpak install org.gnome.Platform//49 org.gnome.Sdk//49 org.freedesktop.Sdk.Extension.rust-stable//25.08
```
Doublonmousse: This needs to be bumped to gnome 50 as well

flxzt: of course, fixed.

PR #1669, opened 2026-02-21
refactor: replace deprecated `HopSlotMap` with `SlotMap` type

flxzt:
According to the documentation
(see: https://docs.rs/slotmap/1.1.1/slotmap/#choosing-slotmap-hopslotmap-or-denseslotmap) `SlotMap` is slow for iteration when there are a lot of removed fields. Because we actually don't remove strokes out of the map, this should not have a big/any impact.

### DIFF
--- a/crates/rnote-engine/src/engine/snapshot.rs
+++ b/crates/rnote-engine/src/engine/snapshot.rs
@@ -8,7 +8,7 @@ use crate::{Camera, Document, Engine};
 use anyhow::Context;
 use futures::channel::oneshot;
 use serde::{Deserialize, Serialize};
-use slotmap::{HopSlotMap, SecondaryMap};
+use slotmap::{SecondaryMap, SlotMap};
 use std::sync::Arc;
 use tracing::error;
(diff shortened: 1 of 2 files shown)


Doublonmousse:
Ah, because `trash_components` is a `SecondaryMap` ...

Do you remember why you chose `HopSlotMap` initially though ? (Trying a `git blame` but coming short for now)
Maybe assumptions changed and the `HopSlotMap` doesn't hold an advantage over the `SlotMap`

flxzt:
> Ah, because trash_components is a SecondaryMap

even with that, the only actual *removal* of keys is in the unused function `remove_trashed_strokes()`.
When we mark strokes as trashed, it is only a property in trash_components.
Only when saved to a file the trashed strokes are not included.
Which makes sense, the history should be kept while the document is being worked on.

> Do you remember why you chose HopSlotMap initially though ?

At the time removal was still considered so I believed HopSlotMap is a reasonable trade-off. Turned out differently

PR #1672, opened 2026-02-23
improv: set pen_mode_eraser_style default to eraser

DorianRudolph:
It took me forever to figure out how to use the eraser button. For reference, I'm currently on macOS with a Wacom One graphics tablet. The primary button gets reported as "Eraser" by GTK. So to use it, I have to tap the eraser button in rnote while holding the physical eraser button on my pen.

If we change the default to `pen_mode_eraser_style: PenStyle::Eraser`, then the eraser will work as expected out of the box.

### DIFF
--- a/crates/rnote-engine/src/pens/pensconfig/mod.rs
+++ b/crates/rnote-engine/src/pens/pensconfig/mod.rs
@@ -21,7 +21,7 @@ use rnote_compose::Color;
 use rnote_compose::penevent::ShortcutKey;
 use serde::{Deserialize, Serialize};
 
-#[derive(Clone, Debug, Default, Serialize, Deserialize)]
+#[derive(Clone, Debug, Serialize, Deserialize)]
 #[serde(default, rename = "pens_config")]
 pub struct PensConfig {
     #[serde(default, rename = "brush_config")]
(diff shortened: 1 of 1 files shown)


crates/rnote-engine/src/pens/pensconfig/mod.rs
```diff
@@ -45,6 +45,22 @@ pub struct PensConfig {
...
+        Self {
+            brush_config: BrushConfig::default(),
+            shaper_config: ShaperConfig::default(),
+            typewriter_config: TypewriterConfig::default(),
+            eraser_config: EraserConfig::default(),
+            selector_config: SelectorConfig::default(),
+            tools_config: ToolsConfig::default(),
+            shortcuts: Shortcuts::default(),
+            pen_mode_pen_style: PenStyle::default(),
+            pen_mode_eraser_style: PenStyle::Eraser,
```
flxzt, line 59: for the other fields write:

```
..Default::default()
```

DorianRudolph: I don't think I can do that inside the `Default::default`.

```rs
impl Default for PensConfig {
    fn default() -> Self {
        Self {
            pen_mode_eraser_style: PenStyle::Eraser,
            ..Default::default()
        }
    }
}
```
gives
```
warning: function cannot return without recursing
  --> crates/rnote-engine/src/pens/pensconfig/mod.rs:49:5
   |
49 |     fn default() -> Self {
   |     ^^^^^^^^^^^^^^^^^^^^ cannot return without recursing
...
52 |             ..Default::default()
   |               ------------------ recursive call site
   |
   = help: a `loop` may express intention better if this is on purpose
   = note: `#[warn(unconditional_recursion)]` on by default
   ```

flxzt: true, I only read the diff superficially

flxzt:
thank you. That should resolve a lot of confusion for other users!

PR #1678, opened 2026-02-28
feat: atomic file saving

### LINKED ISSUE #1128: More resilient file saving by a two-stage file save process
As mentioned in #1096 it would be very nice to improve the save logic by first saving to a temporary second file and overwriting the first when writing the file actually succeeded.

Right now we are directly overwriting the file, which can corrupt files when the app exits unexpectedly while saving.

anesthetice:
Split-off feature from #1578 with additional changes (the use of `tempfile`)

fixes #1128

### DIFF
--- a/crates/rnote-engine/src/utils.rs
+++ b/crates/rnote-engine/src/utils.rs
@@ -150,7 +150,7 @@ where
         .with_context(|| "Failed to sync the contents and metadata of the temporary file")?;
 
     // Finally, we persist the temporary file to the target filepath, if a file
-    // pre-exists at this location, it will be atomically replaced by our new file.
+    // preexists at this location, it will be atomically replaced by our new file.
     let _ = temp_file
         .persist(filepath)
         .with_context(|| "Failed to persist the temporary file to the target filepath")?;
--- a/crates/rnote-ui/src/canvas/imexport.rs
+++ b/crates/rnote-ui/src/canvas/imexport.rs
@@ -210,7 +210,7 @@ impl RnCanvas {
     /// Saves the document to the given file.
     ///
     /// Returns:
-    /// - `Ok(true)` if saving was sucessful
+    /// - `Ok(true)` if saving was successful
     /// - `Ok(false)` if a save was already in progress (and thus this function didn't do anything)
     /// - `Err(e)` when saving failed in any way
     #[tracing::instrument(skip_all, fields(path = format!("{:?}", file.path())))]
@@ -262,8 +262,11 @@ impl RnCanvas {
         debug!("Saving file has finished successfully");
 
         if !skip_set_output_file {
-            // We only create/replace the file watcher once we are sure the file was sucessfully saved.
+            // We only create/replace the file watcher once we are sure the file was successfully saved.
             self.set_output_file(Some(file.to_owned()));
+            // Required, otherwise `output_file_expect_write` will be stuck on true after saving for the
+            // first time or saving as another filename, until the subsequent save at least.
+            self.set_output_file_expect_write(false);
         }
         self.set_unsaved_changes(false);
         self.set_save_in_progress(false);


anesthetice:
@flxzt, @Doublonmousse,

I'd say this PR is ready for review, let me know what you think

crates/rnote-engine/src/utils.rs
```diff
@@ -118,3 +122,58 @@ pub mod glib_bytes_base64 {
         rnote_compose::serialize::sliceu8_base64::deserialize(d).map(glib::Bytes::from_owned)
     }
 }
+
+/// Attempts to atomically save data to a file.
+/// This function is not asynchronous, don't forget to wrap
+/// it inside a `gio::spawn_blocking` or equivalent if need be.
+pub fn atomic_save_to_file<Q>(filepath: Q, bytes: &[u8]) -> anyhow::Result<()>
```
flxzt: Use `async-fs` here and make it async. The caller already is async, why complicate by nesting blocking code

anesthetice: Should be improved now, my rationale was that `tempfile`-related operations (specifically `new_in` and `persist`) could block and thus it would be better if the whole function was treated as such...

flxzt: because of spinning up disks and the like? Good point, I wonder if these operations actually will cause this or if it is cached until actual data is written to the FS.

You think it will be a real issue?

anesthetice: I'm not very confident, but `gio::spawn_blocking` might be the better choice in the end for a couple of reasons:

- Avoids having to use `async_fs` which uses `blocking` under the hood, it manages a thread pool that runs blocking IO operations. (although it is used elsewhere in rnote-ui, so probably not that big of a deal)

- Creating the temporary file as well as persisting it (boils down to a rename syscall) will block the thread that invoked them until the syscall returns, typically not noticeable (unlike `sync_all`) but might be in rare cases (e.g., due to unlucky RNG or a very slow/overloaded filesystem?).

- We're calling `atomic_save_to_file` which contains blocking code from the main gtk thread.

Also noticed the `create_replace_file_future` in rnote-ui/src/utils.rs which I should take a closer look at to see if and when it should be replaced with the atomic version.

anesthetice: Did a little bit of benchmarking. For my system (BTRFS, SSD), under normal conditions, the `persist` operation takes between 200 and 400 µs (i.e., 0.2-0.4 ms). Under strain (`stress-ng --iomix 6`), I get anywhere from 200 to 30000 µs (i.e., 0.2-30 ms); which does make certain parts of the UI noticeably hang.

I also tested with the previous commit, which never caused any part of the UI to hang during the stress test, but again such conditions are not very realistic...

flxzt: hey, you never know if people compile kernels while sketching.

Thanks for actually measuring this!
Then let's go with your previous implementation using `spawn_blocking()`.
Sorry about the hasty suggestion, I did not consider this at all to be honest.

anesthetice: No worries, was worth trying out anyways!

On that note, I could maybe also try a sort of combination of both approaches discussed: use `async-fs` and just wrap the `tempfile`-related operations (that invoke syscalls) in `blocking::unblock()`; might be better if for example `gio::spawn_blocking` is less efficient.

flxzt: that would be elegant, in my opinion

anesthetice:
@flxzt

I converted this PR to a draft as I was checking a couple of things:

- Immediately re-reading the file to verify the checksum is most likely not actually very useful (and quite expensive), so I removed the operation.
- File operations on UNIX systems are painful, it's apparently wise to also call `fsync` on the parent directory, which is now done. (For a real guarantee, one might actually then have to recursively call `fsync` on all of the parent directories.)

Also, the crate [atomic-write-file](https://github.com/andreacorbellini/rust-atomic-write-file) seems pretty good, which we could just use instead of relying on tempfile.

crates/rnote-engine/src/utils.rs
```diff
@@ -118,3 +122,51 @@ pub mod glib_bytes_base64 {
...
+        .sync_all()
+        .with_context(|| "Failed to sync the contents and metadata of the temporary file")?;
+
+    // Finally, we persist the temporary file to the target filepath, if a file
+    // pre-exists at this location, it will be atomically replaced by our new file.
+    let _ = temp_file
+        .persist(filepath)
+        .with_context(|| "Failed to persist the temporary file to the target filepath")?;
+
+    #[cfg(unix)]
```
flxzt, line 158: Is there a downside of syncing on all platforms?

anesthetice: Sorry for the long delay, this operation will not work on windows (tested it myself to make sure) and is apparently not needed from what I've read. it handles files and directories too differently.

I would also like to test this on macOS just to be safe, but unfortunately I don't have access to a device running it...

flxzt: Ah, if opening the directory fails on windows then of course that should be guarded. I imagine on any UNIX this should be fine.

Edit: this is fine, read up the reason why syncing the parent directory might be necessary.

flxzt:
This looks pretty good, could be ready to get merged.

Doublonmousse:
Seems like things don't always work on mac os.
Sometimes it is fine, but other time it seems like the operations are delayed or the notification comes later so the save both works (content saved) and fails at the same time (detected as the file being renamed or moved). In that case the log is 
```
 277.598926292s DEBUG save_document_to_file: rnote::canvas::imexport: Saving file is now in progress path="Some(\"/Users/USER/Documents/New Document.rnote\")"
 277.614542667s DEBUG save_document_to_file: rnote::canvas::imexport: Saving file has finished successfully path="Some(\"/Users/USER/Documents/New Document.rnote\")"
 278.627569417s DEBUG rnote::canvas: event=DebouncedEvent { event: Event { kind: Create(File), paths: ["/Users/USER/Documents/.tmpMYNoqo"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 3286, tv_nsec: 305733541 } } expect_write=true msg="output file parent directory watcher event received"
 278.627642292s DEBUG rnote::canvas: event=DebouncedEvent { event: Event { kind: Modify(Name(Any)), paths: ["/Users/USER/Documents/.tmpMYNoqo"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 3286, tv_nsec: 305735708 } } expect_write=true msg="output file parent directory watcher event received"
 278.628077500s DEBUG rnote::canvas: event=DebouncedEvent { event: Event { kind: Modify(Name(Any)), paths: ["/Users/USER/Documents/New Document.rnote"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 3286, tv_nsec: 305750750 } } expect_write=true msg="output file parent directory watcher event received"
 278.629639958s DEBUG rnote::canvas: event=DebouncedEvent { event: Event { kind: Modify(Data(Content)), paths: ["/Users/USER/Documents/New Document.rnote"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 3286, tv_nsec: 305751208 } } expect_write=true msg="output file parent directory watcher event received"
```
where the temp file is clearly visible here.

In the app the "opened file was renamed or moved" popup shows before the "saved document successfully" (even if the final state is having modifications to save and needing to choose again a path for saving)

anesthetice:
> Seems like things don't always work on mac os. Sometimes it is fine, but other time it seems like the operations are delayed or the notification comes later so the save both works (content saved) and fails at the same time (detected as the file being renamed or moved). In that case the log is
> 
> ```
>  277.598926292s DEBUG save_document_to_file: rnote::canvas::imexport: Saving file is now in progress path="Some(\"/Users/USER/Documents/New Document.rnote\")"
>  277.614542667s DEBUG save_document_to_file: rnote::canvas::imexport: Saving file has finished successfully path="Some(\"/Users/USER/Documents/New Document.rnote\")"
>  278.627569417s DEBUG rnote::canvas: event=DebouncedEvent { event: Event { kind: Create(File), paths: ["/Users/USER/Documents/.tmpMYNoqo"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 3286, tv_nsec: 305733541 } } expect_write=true msg="output file parent directory watcher event received"
>  278.627642292s DEBUG rnote::canvas: event=DebouncedEvent { event: Event { kind: Modify(Name(Any)), paths: ["/Users/USER/Documents/.tmpMYNoqo"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 3286, tv_nsec: 305735708 } } expect_write=true msg="output file parent directory watcher event received"
>  278.628077500s DEBUG rnote::canvas: event=DebouncedEvent { event: Event { kind: Modify(Name(Any)), paths: ["/Users/USER/Documents/New Document.rnote"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 3286, tv_nsec: 305750750 } } expect_write=true msg="output file parent directory watcher event received"
>  278.629639958s DEBUG rnote::canvas: event=DebouncedEvent { event: Event { kind: Modify(Data(Content)), paths: ["/Users/USER/Documents/New Document.rnote"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 3286, tv_nsec: 305751208 } } expect_write=true msg="output file parent directory watcher event received"
> ```
> 
> where the temp file is clearly visible here.
> 
> In the app the "opened file was renamed or moved" popup shows before the "saved document successfully" (even if the final state is having modifications to save and needing to choose again a path for saving)

Oh this is not going to be a very fun one to fix, multiple issues:

### 1 - Calling `set_output_file` every time

To fix an issue I was having, I added the following at the end of the `save_document_to_file` function:  
https://github.com/anesthetice/rnote/blob/d83439914a7e7d9f5e3de7d51bd0faa48b498dc0/crates/rnote-ui/src/canvas/imexport.rs#L253-L261

But this changes the "output-file" property of the canvas, so a new file watcher is created:  
https://github.com/flxzt/rnote/blob/55ce8c90bb6037d6b49c92b494a10ca22dc16a96/crates/rnote-ui/src/canvas/mod.rs#L1127-L1142

At this point on my machine at least, I think the following is going on: the previous events have hit the old watcher, but since the timeout of 1 second has not yet passed they are not emitted, a new file watcher is created and replaces the old one (which is further stopped), this new file watcher has not received any of the events. 

### 2 - Inconsistent reporting of events by `notify`

When forcing a single instance of the file watcher to always be active, and saving a file, I get the following instead:

```
  13.975508131s  INFO rnote::canvas: event=DebouncedEvent { event: Event { kind: Create(File), paths: ["/home/user/Downloads/b.rnote"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 8960, tv_nsec: 613640912 } } expect_write=true msg="output file parent directory watcher event received"
  13.975567222s  INFO rnote::canvas: event=DebouncedEvent { event: Event { kind: Access(Open(Any)), paths: ["/home/user/Downloads/b.rnote"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 8960, tv_nsec: 613648606 } } expect_write=true msg="output file parent directory watcher event received"
  13.975599132s  INFO rnote::canvas: event=DebouncedEvent { event: Event { kind: Access(Close(Write)), paths: ["/home/user/Downloads/b.rnote"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 8960, tv_nsec: 622564915 } } expect_write=true msg="output file parent directory watcher event received"
  
// you can ignore this one, it stems from calling sync on the parent directory
  13.975632926s  INFO rnote::canvas: event=DebouncedEvent { event: Event { kind: Access(Open(Any)), paths: ["/home/user/Downloads"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 8960, tv_nsec: 622584612 } } expect_write=false msg="output file parent directory watcher event received"
```

From reading https://github.com/notify-rs/notify/issues/129, this is maybe seemingly intended? But doesn't match your events, maybe related to this https://github.com/notify-rs/notify/issues/554?

Anyways, I'll try to fix issue n°1 first, will notify you when I'm done @Doublonmousse, to check if that fixes things.

Doublonmousse:
Is this a mac-os specific thing ? (does it never happen/always work on everything else - aka linux with meson and flatpak, windows- no matter what ?)

From other issues (like #990) maybe `notify` isn't robust or timeouts aren't enough in our case (I'm wary of timeouts as well, maybe this kind of issue doesn't happen because events are generally fast enough. But it might not be 100 % true on all cases if some events are buffered, delayed, or take more time on slower drive. That would also explain why it'd appear more on cloud storage). The issue doesn't happen 100% of the time either.

Maybe waiting on the `Modify(Data(Content))` to appear on our file when saving is more robust than doing timeouts (if this is consistently the last message and it not appearing == bug then this could be a workaround).

Maybe (for `EventKind::Modify(ModifyKind::Data(_data_change))`)
https://github.com/flxzt/rnote/blob/55ce8c90bb6037d6b49c92b494a10ca22dc16a96/crates/rnote-ui/src/canvas/mod.rs#L977-L978
is not true anymore (if it's really atomic we may have it once and setting the flag to false would work. Though not sure what happens for cloud storage)

Currently we reset the flag here for `EventKind::Access(AccessKind::Close(AccessMode::Write))`
https://github.com/flxzt/rnote/blob/55ce8c90bb6037d6b49c92b494a10ca22dc16a96/crates/rnote-ui/src/canvas/mod.rs#L1039-L1041
(maybe this happen before the `ModifyKind::Data` event ?)

and here (but only if an error is found so probably not in my case)
https://github.com/flxzt/rnote/blob/55ce8c90bb6037d6b49c92b494a10ca22dc16a96/crates/rnote-ui/src/canvas/imexport.rs#L273-L280

anesthetice:
Could you try again on macOS @Doublonmousse? Does the `Access(Close(Write))` event really not occur?

edit, this could be useful, the events read from `notify` directly (no debouncer) and cleaned up for readability:

```
event: Create(File), paths: ["/home/user/Downloads/.tmpPNUrOT"]
event: Access(Open(Any)), paths: ["/home/user/Downloads/.tmpPNUrOT"]
event: Modify(Data(Any)), paths: ["/home/user/Downloads/.tmpPNUrOT"]

event: Modify(Name(From)), paths: ["/home/user/Downloads/.tmpPNUrOT"], attr:tracker: Some(6291)
event: Modify(Name(To)), paths: ["/home/user/Downloads/aa.rnote"], attr:tracker: Some(6291)
event: Modify(Name(Both)), paths: ["/home/user/Downloads/.tmpPNUrOT", "/home/user/Downloads/aa.rnote"], attr:tracker: Some(6291)

event: kind: Access(Close(Write)), paths: ["/home/user/Downloads/aa.rnote"]
event: kind: Access(Open(Any)), paths: ["/home/user/Downloads"]
```

Doublonmousse:
Trying to add some traces (default debouncer config for now), going to redo with the latest commit shortly.
Doing
```rs
    pub(crate) fn set_output_file_expect_write(&self, expect_write: bool) {
        debug!("set_output_file_expect_write to {:?}", expect_write);
        self.imp().output_file_expect_write.set(expect_write);
    }
```
Shows that the value is never set back to false on mac os.
```
  22.759464292s DEBUG rnote::canvas: event=DebouncedEvent { event: Event { kind: Modify(Name(Any)), paths: ["/Users/user/Documents/GitHub/rnote/fresh_test.rnote"], attr:tracker: None, attr:flag: None, attr:info: None, attr:source: None }, time: Instant { tv_sec: 1234, tv_nsec: 529910708 } } expect_write=true msg="output file parent directory watcher event received"
```
is the event that triggers the issue (in `EventKind::Modify(ModifyKind::Name(_))`). It is _not_ `EventKind::Modify(ModifyKind::Data(_data_change))` because the `set_output_file_expect_write` is never called to `false` so any such event is filtered out.

Edit: same thing with the commit. Removing the debouncer timeout shows that this can also happen on the first save as well.

anesthetice:
Not the cleanest fix (if it even fixes the issue), but not the cleanest bug either...

Does it work now?

Doublonmousse:
> Does it work now?

Yes it does

flxzt:
Thanks!

Issue #1159, opened 2024-07-26
PopUp file modiefied on disk


hantangantan:
**Describe the bug**  
When pressing save button n times while save in progress n-1 popups appear after each other signaling that file was changed on hard drive.

**To Reproduce**  
1. Create/Open document
2. tap save button multiple times (spam button)

**Expected behavior**  
No popup boxes appear

**Screenshots**  
![image](https://github.com/user-attachments/assets/291a6bc4-3e90-443d-8900-19cf6f619465)

**Desktop (please complete the following information):**  
 - OS:  Fedora 40
 - App Version: Rnote v0.11.0
 - Installation Source: Flatpak
 - Desktop Environment: KDE Plasma 6.1.3
 - Display Server:  Wayland

flxzt:
can you run the application like mentioned in the bug report template and post the output? Thank you.

hantangantan:
i only get this output within my terminal:

`2024-07-26T15:11:17.450097Z DEBUG rnote: .. tracing subscriber initialized.
`
then programm is started but detached from my terminal

flxzt:
Detached from your terminal? That shouldn't happen. Your run command should be `flatpak --env=RUST_LOG=rnote=debug --env=RUST_BACKTRACE=1 run com.github.flxzt.rnote`.

Did you grant permissions to the path where your save file is located for rnote in flatseal? Otherwise there can be issues with the file watcher, possibly resulting in this `modified on disk` warning.

hantangantan:
Oh, I am sorry - the command is actually working - when i spammed the button first, simply no output was created.

I just tried again - now the app crashes when spamming the button.

Output log is:
`
**
Adwaita:ERROR:../src/adw-toast-overlay.c:244:show_toast: assertion failed: (!info->widget)
Bail out! Adwaita:ERROR:../src/adw-toast-overlay.c:244:show_toast: assertion failed: (!info->widget)
`

PR #1680, opened 2026-03-01
feat(cli): add timeout to thumbnail command

flxzt:
abort thumbnail generation after a timeout (default: 5 secs)

Nautilus logs:

```
(org.gnome.Nautilus:123409): GnomeDesktop-DEBUG: 17:49:28.795: About to launch script: bwrap --ro-bind /usr /usr --ro-bind-try /etc/ld.so.cache /etc/ld.so.cache --symlink /usr//bin /bin --symlink /usr//lib64 /lib64 --symlink /usr//lib /lib --symlink /usr//sbin /sbin --setenv GST_REGISTRY_1_0 /home/felixz/.cache/gnome-desktop-thumbnailer/gstreamer-1.0/gstreamer-1.0.registry --bind /home/felixz/.cache/gnome-desktop-thumbnailer/gstreamer-1.0 /home/felixz/.cache/gnome-desktop-thumbnailer/gstreamer-1.0 --ro-bind-try /etc/alternatives /etc/alternatives --proc /proc --dev /dev --chdir / --setenv GIO_USE_VFS local --unshare-all --die-with-parent --setenv G_MESSAGES_DEBUG all --bind /tmp/gnome-desktop-thumbnailer-1A2NL3 /tmp --ro-bind /home/felixz/Downloads/large-dup.rnote /tmp/gnome-desktop-file-to-thumbnail.rnote --seccomp 31 rnote-cli thumbnail /tmp/gnome-desktop-file-to-thumbnail.rnote /tmp/gnome-desktop-thumbnailer.png -s 256 

Thumbnail...
Error: Timeout reached

(org.gnome.Nautilus:123409): nautilus-thumbnails-DEBUG: 17:49:38.631: (Thumbnail Async Thread) Thumbnail failed: file:///home/felixz/Downloads/large-dup.rnote (Child process exited with code 1)
```

### DIFF
--- a/crates/rnote-engine/src/store/selection_comp.rs
+++ b/crates/rnote-engine/src/store/selection_comp.rs
@@ -3,7 +3,7 @@ use super::render_comp::RenderCompState;
 use super::{StrokeKey, StrokeStore};
 use crate::strokes::Stroke;
 use crate::strokes::content::GeneratedContentImages;
-use p2d::bounding_volume::Aabb;
+use p2d::bounding_volume::{Aabb, BoundingVolume};
 use rnote_compose::shapes::Shapeable;
 use serde::{Deserialize, Serialize};
 use std::sync::Arc;
@@ -135,20 +135,27 @@ impl StrokeStore {
 
     /// Return the thumbnail keys in the order that they should be rendered.
     /// Keys that would be too small to be visible in the thumbnail are skipped to speed-up thumbnail generation
-    pub(crate) fn thumbnail_keys_as_rendered(&self, size: na::Vector2<f64>) -> Vec<StrokeKey> {
+    pub(crate) fn thumbnail_keys_as_rendered(
+        &self,
+        size: na::Vector2<f64>,
+    ) -> (Vec<StrokeKey>, Option<Aabb>) {
         const MIN_SIZE: f64 = 2.0;
 
-        let keys: Vec<StrokeKey> = self
-            .keys_sorted_chrono()
-            .into_iter()
-            .filter(|&key| !(self.trashed(key).unwrap_or(false)))
+        let keys: Vec<(StrokeKey, Aabb)> = self
+            .key_tree
+            .iter()
+            .filter(|(key, _)| !(self.trashed(*key).unwrap_or(false)))
             .collect();
-        let Some(bounds) = self.bounds_for_strokes(&keys) else {
-            return Vec::new();
-        };
+        if keys.is_empty() {
+            return (Vec::new(), None);
+        }
+        let bounds = keys
+            .iter()
+            .fold(Aabb::new_invalid(), |acc, e| acc.merged(&e.1));
         let scale = size.component_div(&bounds.extents());
-        keys.into_iter()
-            .filter_map(|key| {
+        let mut keys = keys
+            .into_iter()
+            .filter_map(|(key, _)| {
                 let stroke = self.stroke_components.get(key)?;
                 let extents = stroke.bounds().extents();
                 if extents[0] * scale[0] > MIN_SIZE || extents[1] * scale[1] > MIN_SIZE {
@@ -156,6 +163,8 @@ impl StrokeStore {
                 }
                 None
             })
-            .collect::<Vec<StrokeKey>>()
+            .collect::<Vec<StrokeKey>>();
+        self.sort_keys_chrono(&mut keys);
+        (keys, Some(bounds))
     }
 }
--- a/crates/rnote-engine/src/engine/export.rs
+++ b/crates/rnote-engine/src/engine/export.rs
@@ -388,14 +388,10 @@ impl Engine {
     /// Therefore it is ensured generating a SVG from it will never return `Ok(None)`.
     pub fn extract_thumbnail_content(&self, size: na::Vector2<f64>) -> StrokeContent {
         let scale_factor = self.camera.scale_factor();
-        let selection_keys = self.store.thumbnail_keys_as_rendered(size * scale_factor);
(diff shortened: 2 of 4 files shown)


flxzt:
Which value would you suggest? In my testing even 5s is very short for medium sized documents, not sure if we should shorten it even more. Realistically, if a directory has that many rnote files, they should be cached (the thumbnail in file managers should be cached as long as the file itself does not change)

> Do we shut down correctly the in-progress thumbnail progression when the timeout occurs ?

Asynchronicity is archieved through a separate thread doing the rendering, sending the result through a channel. So there is not really cancellation, other than discarding the result once finished. However since the program is terminated once the timeout is reached, the in-progress thread is cleaned up as well.

Doublonmousse:
Let's leave it at 5 seconds then.
We can optimize the thumbnailer performance later

flxzt:
> We can optimize the thumbnailer performance later

I wonder if it makes sense to limit the amount of strokes in the selection for which the thumbnail is generated.
Maybe clever logic could be to calculate the px-size of strokes, leaving out those that will not be visible in the small thumbnail picture

flxzt:
Okay, added the following, should speed up thumbnailing:

This speeds up thumbnailing by filter out strokes
that would be too small given a thumbnail size.
Additionally empty documents are also thumbnailed now,
with only the document background as content.
Some smaller API cleanup tasks made it in as well.

crates/rnote-engine/src/store/selection_comp.rs
```diff
@@ -131,4 +132,29 @@ impl StrokeStore {
 
         new_selected
     }
+
+    /// Return the thumbnail keys in the order that they should be rendered.
```
Doublonmousse, line 136: Maybe it's worth adding that
```rs
/// Keys that would be too small to be visible in the thumbnail are skipped to speed-up thumbnail generation
```

crates/rnote-engine/src/store/selection_comp.rs
```diff
@@ -131,4 +132,29 @@ impl StrokeStore {
...
+    pub(crate) fn thumbnail_keys_as_rendered(&self, size: na::Vector2<f64>) -> Vec<StrokeKey> {
+        const MIN_SIZE: f64 = 2.0;
+
+        let keys: Vec<StrokeKey> = self
+            .keys_sorted_chrono()
+            .into_iter()
+            .filter(|&key| !(self.trashed(key).unwrap_or(false)))
+            .collect();
+        let Some(bounds) = self.bounds_for_strokes(&keys) else {
```
Doublonmousse: This is probably where most time is spent (from #1462).

We don't have to do this now (maybe only after doing the other PR, this is more for the future), but it's probably faster to use the rtree to get bounds out. And maybe the rtree allows _some_ amount of pruning (walk the tree but stop when the envelope of the child itself is smaller than the requested size here).

crates/rnote-engine/src/store/selection_comp.rs
flxzt: applied as you suggested

crates/rnote-engine/src/store/selection_comp.rs
flxzt: okay, I added that the thumbnail bounds from retrieved through the rtree.

flxzt: Could not yet figure out if the tree itself can be only be walked until the stroke bound is smaller than the envelope. Maybe I missed it in the docs.

Doublonmousse: I _think_ we can get this through https://docs.rs/rstar/0.12.2/rstar/struct.RTree.html#method.root then doing things manually over https://docs.rs/rstar/0.12.2/rstar/enum.RTreeNode.html.

Though at this point we're doing/touching the internal `rstar` types so ... maybe not worth doing if we don't know that we're going to use this for other things that thumbnails/the performance is advantageous.

(Maybe) if we want to remove elements that get too small to be visible when zooming out for the canvas (though not sure that happens that much for the min/max zoom levels) and I'm not sure removing/adding elements in these conditions would speed up things currently.

flxzt: maybe a follow-up to look into this?

Doublonmousse: Yes, that makes sense

flxzt:
> f dolphin is opened (in view and visibly recomputing the thumbnail on each file update) and I quickly change things/save, the Opened file was modified on disk is triggered (and reloading means rolling back a few strokes)

This is an issue, but are you sure it is related to the thumbnailer? I get this also when quickly saving in succession with my file manager completely stopped.

> Some names don't get thumbnails (seems triggered by " being in the filename)

does a6d0e0d7ad2468463eeeab221b68e772415a82b4 fix this? It quotes the file arguments in the thumbnailer invocation

Doublonmousse:
> This is an issue, but are you sure it is related to the thumbnailer? I get this also when quickly saving in succession with my file manager completely stopped.

You're probably right, (seems more like #990 or #1159)

> does https://github.com/flxzt/rnote/commit/a6d0e0d7ad2468463eeeab221b68e772415a82b4 fix this? It quotes the file arguments in the thumbnailer invocation

It doesn't seem to work and seems to make the thumbnailer fail altogether on dolphin (no thumbnail produced). So I reckon it's maybe best we keep the cli invocation as is and fix this later

flxzt:
> It doesn't seem to work and seems to make the thumbnailer fail altogether on dolphin (no thumbnail produced). So I reckon it's maybe best we keep the cli invocation as is and fix this later

Ok, I dropped the commit that changed this.

Doublonmousse:
I think this is ready for a merge

PR #1694, opened 2026-03-17
chore: update `rnote-windows-build.md` to include pinned version

### LINKED ISSUE #1693: pink boxes on Windows 11
**Describe the bug**  
When building and running rnote on Windows 11, everything is rendered pink. Strokes, images etc. are all rendered as pink boxes. According to #1615 this is caused by gtk4 4.20, which should now be fixed to 4.18. But this problem still happens.

**Expected behavior**  
Lines and images should be rendered

**Screenshots**  

Image

**Desktop:**  
 - OS: Windows 11 25H2
 - App Version: Rnote v0.13.1
 - Installation Source: Repo

Doublonmousse:
By refering to the `justfile prerequisites-win` section and using `just` for the dependency setup

This way we only have to update the `justfile`. I'm open to comment (is it better to keep the dependency list but with a warning to check the `justfile` for pinned versions ?)

Fixes #1693

### DIFF
--- a/misc/building/rnote-windows-build.md
+++ b/misc/building/rnote-windows-build.md
@@ -8,19 +8,7 @@
 
 > The following instructions assume that the default installation directories were used.
 
-### Dependencies
-
-In order to install the necessary dependencies, run the following command in a MSYS2 terminal.
-
-```bash
-pacman -S --noconfirm \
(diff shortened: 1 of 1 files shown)


flxzt:
> I'm open to comment (is it better to keep the dependency list but with a warning to check the justfile for pinned versions ?)

No, I think this is good!

Note: we should do the same for the other build instructions at some point as well.

Issue #367, opened 2022-11-23
Rethink toolbar layout


bertob:
Currently the app has three "toolbars":
- The headerbar, including the tool chooser (split out into a separate toolbar on small sizes)
- The active tool options
- The floating toolbar with zoom/undo/etc

![image](https://user-images.githubusercontent.com/1908896/203445953-33853943-5eae-4ec2-824d-1d13202211ff.png)

The active tool options are the group with the most content, yet it's vertical, which is usually the shorter axis, leading to scrolling. The zoom/undo/etc buttons feel like they'd be more appropriately grouped with view options and save in the toolbar than the tool chooser, which is more directly linked to the canvas and tool options.

I'd try something like the following:
- Tool options as a horizontal bar at the bottom
- Zoom/undo/etc in the headerbar (perhaps )
- Tool chooser as a toolbar or floating buttons on the side
- This would also allow moving the save button to the right side, which would be more in line with other apps

benjamin051000:
As a lefty, the floating undo/redo buttons do get in the way sometimes... It would be nice to be able to move them, either to another corner, or in the headerbar would be fine too. I do love the aesthetic of it right now so it's difficult to say what would be best.

flxzt:
I am in the process of implementing tabs and took the opportunity to move some things around. What do you think @bertob ? Obviously needs more polishing but you get the idea.

I feel like the zoom / undo buttons in  the headerbar are a bit too much, especially in the "fixed-size" layout where two additional buttons appear (for adding / resizing pages ). But I don't really have a better idea where to place them myself. It is a real challenge to make this all work for small widths. Maybe have them as an overlay in the top-left corner, but in a vertical box instead of the current horizontal?

I now think that the bottom toolbar could work, in contrary to the concerns I expressed regarding UI where the writing hand lays. When it is in the center, it probably won't be in the way too much.

![Bildschirmfoto vom 2023-01-05 18-26-32](https://user-images.githubusercontent.com/19841886/210842979-1e9e3223-cc71-444e-a0f2-7c67705c8882.png)

wiryfuture:
imo, the spanner section with the move tools feels a bit strange. Maybe the two move tools could merge into one icon, and then have a toggle between them under the tool options bar on the side.

bertob:
[...]

PR #1706, opened 2026-03-26
feat: UI redesign

yuukibarns:
- Move `colorpicker` and `penpicker` to `mainheader`.
- Add a delete node to selection overlay.

See also #367

### DIFF
--- a/crates/rnote-compose/src/style/indicators.rs
+++ b/crates/rnote-compose/src/style/indicators.rs
@@ -229,6 +229,49 @@ pub fn draw_circular_node(
     }
 }
 
+/// Draw a circular delete node.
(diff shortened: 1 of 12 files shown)


yuukibarns:
https://github.com/user-attachments/assets/4d40f4c1-ae51-4dcb-a5a2-0ed2259a619e

flxzt:
we moved away from this layout for various reasons. See: https://github.com/flxzt/rnote/issues/367

yuukibarns:
What about this one? Integrated one tool bar for the simplicity philosophy of big gnome

https://github.com/flxzt/rnote/issues/367#issuecomment-2027412169

PR #1710, opened 2026-03-28
fix(ui): ignore Remove events triggered by atomic saves in file watcher

### LINKED ISSUE #1705: Files are incorrectly detected as removed
**Describe the bug**  
After some minutes after I open an `.rnote` file, I get a message that the file got removed and it opens the current drawings in a "new document" that I need to save again.

**To Reproduce**  
1. Open an `.rnote` file
2. Wait a few minutes
3. Notice a message pops up from the bottom of the screen, saying the original file has been removed

**Expected behavior**  
It should not detect the original file as removed.

**Desktop**  
 - OS: Windows 11 Home 25H2
 - App Version: Rnote v0.14.0
 - Installation Source: Windows installer

**Additional context**  
The relevant code is [here](https://github.com/flxzt/rnote/blob/47909d02a44d647de9e3c621a500ec2700798419/crates/rnote-ui/src/canvas/mod.rs#L1028).

turicfr:
Fixes #1705.

### DIFF
--- a/crates/rnote-ui/src/canvas/mod.rs
+++ b/crates/rnote-ui/src/canvas/mod.rs
@@ -1032,6 +1032,11 @@ impl RnCanvas {
                     if !crate::utils::paths_abs_eq(file_path, event_path).unwrap_or(false) {
                         return;
                     }
+                    if canvas.output_file_expect_write() {
+                        // Atomic save operations replace the file by temporarily removing or overwriting
+                        // the original file's data/metadata, generating a Remove event. We should ignore these.
+                        return;
+                    }
                     canvas.set_unsaved_changes(true);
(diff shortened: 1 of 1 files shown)


Doublonmousse:
Did you have autosave enabled? That would explain the time delay in your original issue.

@anesthetice for info

anesthetice:
Somewhat strange that a `Remove` event occurs, might be a windows or rather `inotify` thing... Also, does this happen when saving manually? As under the hood autosave calls the same `save_document_to_file` function.

turicfr:
Yes, I had autosave enabled, and it indeed happened when it was trying to save.
Saving the file manually also causes the same issue.

PR #1712, opened 2026-03-29
feat: rotate canvas

### LINKED ISSUE #1212: Add canvas rotation, not only zoom and move.
Canvas rotation should be added.

it would be much easier to draw when you're allowed to change your canvas rotation just like you can change the rotation of your paper in real life. 
This app is great and it is bundled with a lot of things but I feel as if this unique feature would truly set it apart. (although not a completely unique feature still quite rare in low quality drawing apps, however this is not low quality so it deserves this )

Kneemund:
Adds support for rotating the canvas.

- Closes #1212.
- Depends on #1719.
- Canvas can be rotated using touch gestures or scrolling while pressing <kbd>ALT</kbd>.

https://github.com/user-attachments/assets/71f0728c-9ffc-4199-8f17-728653f2b794

### TODO
- [x] UI element that indicates current rotation and can be used for modifying it.
- [ ] Better touch UX - it is very easy to accidentally rotate the canvas while zooming. GIMP seems to have solved this by making zoom and rotation exclusive (https://gitlab.gnome.org/GNOME/gimp/-/merge_requests/597).
- [ ] Document and adjust key bindings & gestures.
- [ ] Selecting while rotated feels weird.

### Known Issues
~~In a fixed layout, rotating by more than 45° (e.g. 180°) and scrolling to the right or bottom edge makes the camera jump back into the center. I don't know why this is happening.~~ Fixed by #1719.


flxzt:
Nice, that's a great addition!

improvements I'd add to the list:
- selecting while rotated feels very weird. I would expect that the selection always has the non-rotated reference coordinate system. It's tricky to get the details how it should behave once rotated, but maybe there could be a recalculation of the selection rectangle once the rotation is finished.
- When the rotation is close to 0° (maybe also +-90° and 180°), it would be nice to have a small band where it snaps to these angles.

Kneemund:
> selecting while rotated feels very weird. I would expect that the selection always has the non-rotated reference coordinate system. It's tricky to get the details how it should behave once rotated, but maybe there could be a recalculation of the selection rectangle once the rotation is finished.

I'm not sure about this. I feel like the other end of the spectrum could also work, i.e. make the rotation rectangle an oriented bounding box which has the same rotation as the camera at the time of its creation.

> When the rotation is close to 0° (maybe also +-90° and 180°), it would be nice to have a small band where it snaps to these numbers.

This is already implemented, there are areas of 2° that snap the rotation into multiples of 45°. Maybe 2° is too small?

---

Also, do you have any ideas for improving the touch gesture? I feel like making rotation and zooming exclusive (like GIMP) is quite restrictive and makes it a lot less "fun" to use, at least for me. I'll probably try adding an activation threshold/hysteresis for rotation and test if that works.

PR #1719, opened 2026-04-03
refactor: canvas widget adjustments

### LINKED ISSUE #1394: Zoom in: canvas jumps around a bit
**Describe the bug**  
While zooming in, the canvas jumps around a bit: 

https://github.com/user-attachments/assets/f3c8ce96-8595-46ca-8320-f88acc910c7d

**To Reproduce**  
1. Open a document
2. Zoom in
3. See error

**Expected behavior**  
A smooth zoom in without the canvas changing its positon/jumping around a bit.

 - OS: Fedora 41
 - App Version: Rnote v0.11.0
 - Desktop Environment: Gnome 47
 - Display Server: Wayland
 - Input Source: Touch

### LINKED ISSUE #1720: Canvas doesn't stay centered in page after zoom
**Describe the bug**  
When I zoom in and out of the page, the center of the drawing area is different from the center of the page. You can manually re-center it using the scrollbars, but it will be lost as soon as you zoom again. The bug is very obvious when you enable visual debugging.

**To Reproduce**  
Just zoom in and out a couple of times and try writing near the left/right edges.

**Console Output**  
No output in console.

**Screenshots**  

 - OS: Arch Linux 6.19.10-zen1-1-zen (64-bit)
 - App Version: Rnote v0.14.0
 - Desktop Environment: KDE Plasma 6.6.3
 - Display Server: Wayland
 - Input Source: Any

https://github.com/user-attachments/assets/02b424d3-4fd6-48ff-8f75-5833455d4df9

### LINKED ISSUE #1738: [Regression] Pinch-to-zoom broken on touchscreen after v0.14.1
**Describe the bug**  
The last update (v0.14.1) which addressed touchpad zooming has caused a regression for touchscreen users on GNOME. Pinch-to-zoom on the screen itself causes sudden jumps.

**To Reproduce**  
1. Open a (preferrably large) document
3. Use the pinch gesture on the touchscreen to zoom
4. Screen suddenly jumps between pages
**Expected behavior**  
Normal smooth zooming (no scrolling)

**Console Output**  
```
   0.000064341s DEBUG rnote: .. tracing subscriber initialized.

(rnote:2): Gtk-CRITICAL **: 11:11:29.509: gtk_adjustment_configure: assertion 'lower + page_size <= upper' failed
  13.462221652s DEBUG rnote::canvas: Started watching directory '/home/rs-212/Documents/Studium/Phase II/Mathe II/SCS/Notes', targetting the file 'Some("Skript Mathe II.rnote")'
```

 - OS: Fedora Linux 43 (Workstation Edition) x86_64
 - App Version: Rnote v0.14.1
 - Desktop Environment: GNOME 49.5
 - Display Server: Wayland
 - Input Source: On-Screen-Tablet

**Additional context**  
Downgrading to v0.14.0 fixed the Issue for me.

Kneemund:
- Fixes #1720.
- Fixes #1738.
- Fixes #1394.
- Applies the workaround from #1733 to all other zoom gestures and actions.
- Fixes "known issue" described in #1712.
- Fixes `gtk_adjustment_configure: assertion 'lower + page_size <= upper' failed` log spam by ensuring that the bounds are at least as big as the widget.

---

- Transforms the camera offset into an absolute range for configuring the adjustments of the canvas widget.
- Updates camera offset as soon as adjustments emit a value changed signal, instead of during layout (re-)allocation.
- Autoexpands the canvas when an adjustment changes (like all other gestures/actions), making the autoexpand on edge overshoot redundant. As a preparation for #1712, we cannot rely on edge overshoot callbacks, because when the document is rotated, we have to expand it *before* reaching the ScrolledWindow edge.
- Because of the above, we hit a [GTK bug](https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/395) related to kinetic scrolling. https://github.com/flxzt/rnote/pull/1719/changes/53af354b4de3cb3e1e2cdf4bbac5931b58c3618b works around this bug by temporarily disabling kinetic scrolling when the document is about to be expanded.

---

### TODO
- [x] It seems we hit this old GTK bug https://gitlab.gnome.org/GNOME/gtk/-/merge_requests/395. I'm currently trying to implement the workaround used in other apps...

### DIFF
--- a/crates/rnote-ui/src/canvas/mod.rs
+++ b/crates/rnote-ui/src/canvas/mod.rs
@@ -568,22 +568,19 @@ mod imp {
                     obj,
                     move |hadj_signal| {
                         // Apply scroll input from adjustment to camera
-                        let (offset_mins, _) = canvas.engine_ref().camera_surface_mins_maxs();
+                        let (surface_mins, _) = canvas.engine_ref().camera_surface_mins_maxs();
                         let offset = canvas.engine_ref().camera.offset();
 
                         let new_offset = na::vector![
                             super::RnCanvas::adjustment_to_surface(
                                 hadj_signal.value(),
-                                offset_mins.x
+                                surface_mins.x
                             ),
                             offset.y
                         ];
 
-                        let _ = canvas.engine_mut().camera_set_offset_expand(new_offset);
-
-                        // this triggers a canvaslayout allocate() call,
-                        // where the camera and content rendering is updated based on some conditions
-                        canvas.queue_resize();
+                        let widget_flags = canvas.engine_mut().camera_set_offset_expand(new_offset);
+                        canvas.emit_handle_widget_flags(widget_flags);
                     }
                 ));
 
@@ -615,22 +612,19 @@ mod imp {
                     obj,
                     move |vadj_signal| {
                         // Apply scroll input from adjustment to camera
-                        let (offset_mins, _) = canvas.engine_ref().camera_surface_mins_maxs();
+                        let (surface_mins, _) = canvas.engine_ref().camera_surface_mins_maxs();
                         let offset = canvas.engine_ref().camera.offset();
 
                         let new_offset = na::vector![
                             offset.x,
                             super::RnCanvas::adjustment_to_surface(
                                 vadj_signal.value(),
-                                offset_mins.y
+                                surface_mins.y
                             )
                         ];
 
-                        let _ = canvas.engine_mut().camera_set_offset_expand(new_offset);
-
-                        // this triggers a canvaslayout allocate() call,
-                        // where the camera and content rendering is updated based on some conditions
-                        canvas.queue_resize();
+                        let widget_flags = canvas.engine_mut().camera_set_offset_expand(new_offset);
+                        canvas.emit_handle_widget_flags(widget_flags);
                     }
                 ));
 
--- a/crates/rnote-ui/src/canvas/mod.rs
+++ b/crates/rnote-ui/src/canvas/mod.rs
@@ -24,7 +24,6 @@ use once_cell::sync::Lazy;
 use p2d::bounding_volume::Aabb;
 use rnote_compose::ext::AabbExt;
 use rnote_compose::penevent::PenState;
-use rnote_engine::Camera;
 use rnote_engine::ext::GraphenePointExt;
 use rnote_engine::ext::GrapheneRectExt;
 use rnote_engine::{Engine, WidgetFlags};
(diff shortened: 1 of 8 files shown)


crates/rnote-ui/src/canvas/mod.rs
```diff
@@ -829,6 +916,29 @@ impl RnCanvas {
...
+    pub(crate) fn workaround_restore_kinetic_scrolling(&self, scroller: Option<&ScrolledWindow>) {
+        if !self
+            .imp()
+            .workaround_kinetic_scrolling_pending
+            .replace(false)
+        {
+            return;
+        }
+
+        if let Some(scroller) = scroller {
```
flxzt: I would have expected a check for `scroller.is_kinetic_scrolling()` here as well.
In the end is not an issue because I believe the restore only happens `workaround_kinetic_scrolling_pending` is enabled,
which in turn only gets set when kinetic scrolling is enabled.
But to make sure this will not be an issue if this logic is ever modified,
I would suggest to add `scrolling.is_kinetic_scrolling()` to this condition as well.
If I didn't miss anything..

flxzt (changes requested):

flxzt:
Looks good, functionally it fixes the outlined issues.
I also like that it fixes this by moving the overshoot logic into the engine,
not relying on GTK's widget overshoot mechanic anymore. This is a cleaner approach.
It also brings down the complexity of this logic a bit (- the needed workarounds) by improving function names, introducing some helpers like `surface_to_adjustment` and add comments. Which is a big step in the right direction but complexity is still high which I'll need to address in a code cleanup in the future to ensure future maintainability.

crates/rnote-ui/src/canvas/mod.rs
Kneemund: That wouldn't work, `scrolling.is_kinetic_scrolling()` is always false here. I made the code a bit more readable and added some comments though.

Kneemund:
I agree that this needs to be cleaned up in the future. Especially the surface <-> adjustment conversion should be more hidden/abstracted away, but I couldn't really find a good way to do that within the current code without re-calculating the surface bounds all the time.

crates/rnote-ui/src/canvas/mod.rs
flxzt: sorry, I actually was suspicious about this overwriting the user setting here (which it doesn't, because of the above reason). I meant the method `inertial_scrolling()` for the user setting. Although it is not easily retrievable, because this code can't get a reference to the `CanvasWrapper` easily.

Then let's leave it like this.

PR #1724, opened 2026-04-05
fix: correct modifier check in gestures

### LINKED ISSUE #1617: Zooming with ctrl + scroll does not work when caps lock is on
**Describe the bug**  
Zooming with ctrl + scroll does not work when caps lock is on. Side scrolling with shift+scroll does still work.

**To Reproduce**  
1. open rnote
2. use ctrl + mouse scroll. View zooms.
3. enable caps lock
4. use ctrl + mouse scroll. View does not zoom.

Operating System: Fedora Linux 43
KDE Plasma Version: 6.5.4
KDE Frameworks Version: 6.21.0
Qt Version: 6.10.1
Kernel Version: 6.17.12-300.fc43.x86_64 (64-bit)
Graphics Platform: Wayland
Processors: 8 × 11th Gen Intel® Core™ i7-1165G7 @ 2.80GHz
Memory: 16 GiB of RAM (15.4 GiB usable)
Graphics Processor: Intel® Iris® Xe Graphics
Manufacturer: LENOVO
Product Name: 20WE
System Version: ThinkBook 14s Yoga ITL
Input source: mouse + keyboard

**Additional context**  
I understand that mouse+keyboard is not the primary mode of interaction for rnote, but it is still useful to have that part work.

Kneemund:
- Fixes #1617.
- Switches to `.contains` instead of strict equality for gesture modifiers.

### DIFF
--- a/crates/rnote-ui/src/canvaswrapper.rs
+++ b/crates/rnote-ui/src/canvaswrapper.rs
@@ -354,9 +354,12 @@ mod imp {
                     #[upgrade_or]
                     glib::Propagation::Proceed,
                     move |controller, _, dy| {
-                        if controller.current_event_state() != gdk::ModifierType::CONTROL_MASK {
+                        let modifiers = controller.current_event_state();
+
+                        if !modifiers.contains(gdk::ModifierType::CONTROL_MASK) {
                             return glib::Propagation::Proceed;
                         }
+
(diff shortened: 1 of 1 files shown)


flxzt:
thanks, and great to see you being involved again! :smile:

PR #1727, opened 2026-04-06
update flathub badge

leoheitmannruiz:
The badge is taken from https://flathub.org/api/badge?svg&locale=en which is linked to at https://flathub.org/en/badges


flxzt:
thanks!

PR #1729, opened 2026-04-07
fix: transparent stroke rendering

### LINKED ISSUE #891: A pen with a transparent color isn't homogenously transparent, but leaves less transparent circles
**Describe the bug**  
If you write with a thicker pen and a transparent color, it isn't homogenously transparent, but rather leaves little circles where its less transparent. I think it "draws twice" in those places probably. This more visible if your stroke is fast. It appears with a mouse, as well as with my drawing tablet.

**To Reproduce**  
1. Click on the color picker
2. Select any color you like, and set it to transparent, around halfway works well
3. Select the thick pen option, or even the marker (that is even more thick)
4. Make a few strokes
5. You will see your stroke, with little darker circles inside of it

**Expected behavior**  
The pen should leave a smooth transparent stroke

**Console Output**  
Running the app with `flatpak --env=RUST_LOG=rnote=debug --env=RUST_BACKTRACE=1 run com.github.flxzt.rnote` gives following output:
` 2023-11-17T17:09:17.388Z DEBUG rnote > ... env_logger initialized

(rnote:2): Gtk-CRITICAL **: 18:09:17.690: gtk_widget_class_set_accessible_role: assertion '!gtk_accessible_role_is_abstract (accessible_role)' failed
 2023-11-17T17:09:17.716Z WARN  rnote::workspacebrowser::workspacesbar > failed to canonicalize dir "/home/nandor/Desktop/rnote/studium/analysis" for workspacelistentry with name: default, Err: No such file or directory (os error 2)
`
I don't think anything here is relevant, but why not include it
**Screenshots**  
image
Well visible circles.

 - OS: Ubuntu 23.04 x86_64
 - App Version: Rnote 0.9.0 
 - Desktop Environment: Gnome 44.3

**Additional context**  
It is vaguely related to #449 . I wanted to highlight on a pdf, and set the marker to be transparent. This looks not well enough for me to be able to use transparent colors anywhere. From what im guessing, the circles are perfectly circular because exactly one point overlaps, making the circle around it half as transparent.

Kneemund:
- Fixes #891.
- Addresses #449.

---

- All stroke elements are collected into a single bezier path before being submitted to cairo. This fixes the blending issue when individual line caps overlap.
- When transparent strokes are being drawn/in progress, they are rendered opaque. This prevents having to regenerate the whole stroke on every input, which would cause serious performance problems when drawing long strokes. It looks a bit weird, but this is the best solution I could think of.

[Bildschirmaufnahme_20260407_172404.webm](https://github.com/user-attachments/assets/6d045c2f-e17c-4ae0-91d9-fded9a20c020)

### DIFF
--- a/crates/rnote-compose/src/color.rs
+++ b/crates/rnote-compose/src/color.rs
@@ -130,18 +130,6 @@ impl Color {
         0.2126 * self.r + 0.7152 * self.g + 0.0722 * self.b
     }
 
-    /// Mix the color with a background color, resulting in a new opaque color.
-    pub fn with_background_color(self, background_color: Self) -> Self {
-        let a = self.a;
-
-        Self::new(
-            self.r * a + background_color.r * (1.0 - a),
-            self.g * a + background_color.g * (1.0 - a),
-            self.b * a + background_color.b * (1.0 - a),
-            1.0,
-        )
-    }
-
     /// Invert the perceived brightness of the color.
     pub fn to_inverted_brightness_color(self) -> Self {
         let mut hwba_color: palette::Okhwba<f64> = self.into_color();
--- a/crates/rnote-compose/src/color.rs
+++ b/crates/rnote-compose/src/color.rs
@@ -130,6 +130,18 @@ impl Color {
         0.2126 * self.r + 0.7152 * self.g + 0.0722 * self.b
     }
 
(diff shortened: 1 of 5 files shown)


flxzt:
I think this is the best solution that doesn't introduce significant performance issues.
Thanks, looks great!

PR #1730, opened 2026-04-07
feat: multiplicative marker blending

### LINKED ISSUE #449: Add an highlighter that also works on background pdf/images

Kneemund:
- Depends on #1729.
- Closes #449.
- Alternative to #1602.

---

- Adds multiplicative blending to the "Marker" pen.
- Moves the highlighter stroke layer to the top.
- Preserves layer information when copy-pasting strokes.

[Bildschirmaufnahme_20260407_171820.webm](https://github.com/user-attachments/assets/28cc68c3-7e3c-4a04-8708-e450f7a4bec9)

---

### TODO

The SVGs that Cairo produces when multiplicative blending is involved seem to be very weird. Simplifying them using `usvg` breaks them, making them fully blank or black in all applications. 

When skipping the simplification, the only applications I have that were able to correctly render them were a Chromium browser, GIMP and Inkscape (didn't work in LibreOffice, Gwenview, Firefox and Rnote). This also fixed Rnote PNG exports, so I guess `librsvg` can do it... *sometimes*?

Any help on this would be appreciated. Below is an example SVG (without simplification). I believe that this could be a dealbreaker for this PR as it seems like a Cairo bug, and Cairo is pretty much maintenance-only.

test

### DIFF
--- a/crates/rnote-engine/src/engine/export.rs
+++ b/crates/rnote-engine/src/engine/export.rs
@@ -339,33 +339,36 @@ impl Engine {
         oneshot_receiver
     }
 
-    pub fn extract_document_content(&self) -> StrokeContent {
+    fn stroke_content_from_sorted_keys(
+        &self,
+        mut keys: Vec<crate::store::StrokeKey>,
+    ) -> StrokeContent {
+        let highlighter_keys = self.store.split_off_highlighter_keys(&mut keys);
+
         StrokeContent::default()
-            .with_strokes(
-                self.store
-                    .get_strokes_arc(&self.store.stroke_keys_as_rendered()),
-            )
+            .with_strokes(self.store.get_strokes_arc(&keys))
+            .with_highlighter_strokes(self.store.get_strokes_arc(&highlighter_keys))
+            .with_background(self.document.config.background)
+    }
+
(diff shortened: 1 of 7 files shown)


TheJanzap:
So as far as I understand, this feature is ready, except that using the marker breaks SVG export, which is broken due to an upstream bug?

While I think it's fair not wanting to merge this when known issues exist, I think in this case it could be justified. The associated issue has 52 upvotes and is highly requested. IMO it's worth trading a working SVG export for highlighting, since most people would probably just like to have highlighting within Rnote and don't really care whether SVG export works.

If possible, I would like to propose adding a warning message to the export dialog that exporting documents with markers in SVG may create a corrupt image (the text itself can be adjusted).

While I would like to see the SVG issue fixed, it would be a shame to leave this highly requested feature by the wayside just because of an upstream bug in the export component.

PR #1739, opened 2026-04-15
Translation: Update for Vietnamese translation vi.po

ng-khang2019:
Added some translations for the Vietnamese translation file


flxzt:
thank you!

PR #1742, opened 2026-04-15
fix: alt shift drag gesture regression

Kneemund:
- Fixes a regression caused by #1724 that prevents the <kbd>ALT</kbd> + <kbd>SHIFT</kbd> + Drag gesture from working (because the <kbd>ALT</kbd> + Drag gesture always claims the event).


flxzt:
confirmed zooming with "Alt+Ctrl+Drag" works again. Thanks!

PR #1743, opened 2026-04-15
feat: Add option to import Pdf into N-wide array

### LINKED ISSUE #997: PDF import into N-wide array?
Today is my first day using this amazing program Rnote and I couldn't stop working with it until I understood most of the feeatures. For a few years now I've been using Xournal++ which is wonderful but doesn't always run smoothly on Mac M1/M2 ARMs.

I figured out how to import PDF's with snapping, US letter, continuous or one-per-document-page, 100% width, which gives me the ability to fit PDF's of this size neatly and automatically to each document page.

I prefer to use a 3- or 4-page wide configuration, which I now have to do manually.
Would it be possible to add an N-wide option, or have some fixed options such as 2-wide, 3-wide, 4-wide,... etc, so that if there were, say, 10 total pages and if on chose 3-wide option, the pages would be arranged thusly:
1,2,3
4,5,6
7,8,9
10

??

Thanks for your incredible program.

### LINKED ISSUE #1559: Book style view for pdf import
You sometimes have to import large PDFs that were originally intended to be printed as a book. The current vertical arrangement options usually work fine for small documents with simple formation. However, larger documents with many pages and a more 'vibrant' formation can be a bit disorientating while having to scroll a lot to find something.

Due to this issue, I would like to have a book style view as an import option for PDFs. Pages should be placed vertically downward as pairs of two. Furthermore, the user should be able to correct for any covers or front pages by setting an offset in the import.

#1383 could somewhat solve this issue, but I find horizontal scrolling not really practical or enjoyable for my tablet/use case.

**Additional context**  
Here is a random image I found that shows the type of arrangement I'm looking for:
Image

cyates5:
Resolves #997 
Partially resolves #1559

This pull request adds a 'Columns' option to the Pdf import menu, which lets the user adjust the number of Pdf pages arranged in each row.

The updated import menu:
Screenshot From 2026-04-15 20-50-38

2 columns in '1 per Document Page' layout:
Screenshot From 2026-04-15 20-56-08

When 'Adjust Document' is enabled, the page width is multiplied by the column count to ensure all columns remain visible:
Screenshot From 2026-04-15 20-58-24

Please let me know if there are any changes needed.


NXTler:
This would also partially resolve #1559

cyates5:
> This would also partially resolve #1559

Thanks, I've updated the pull request.

crates/rnote-engine/src/engine/import.rs
```diff
@@ -81,6 +81,9 @@ impl TryFrom<u32> for PdfImportPageSpacing {
 #[derive(Debug, Clone, Copy, Serialize, Deserialize)]
 #[serde(default, rename = "pdf_import_prefs")]
 pub struct PdfImportPrefs {
+    /// The number of Pdf pages in each row.
+    #[serde(rename = "columns")]
+    pub columns: f64,
```
Doublonmousse: Why f64 ? It makes more sense to do an uint here and cast to floats for positions later

cyates5: Good idea, I've now added a commit to alter this.

flxzt:
seems to be a straight forward addition, can you update the branch and resolve the conflict @cyates5 ? I will test and merge once this is done.

cyates5:
Thanks, I've updated the branch and I think it's now ready to merge.

flxzt:
seems like it does not build.

PR #1745, opened 2026-04-18
feat: expand brush style menu, fix inconsistencies

Kneemund:
- Should be merged alongside #1730.
- Alternative to #1602.
- Fixes some inconsistent margins and separators.

There has been some discussion in #1602 on if the highlighter/marker brush style should be split off into a separate pen in the bottom sidebar. As I've explained there, I think that this is a compromise that makes the highlighter easily accessible while not duplicating parts of the UI and engine.

grafik

### DIFF
--- a/crates/rnote-ui/data/ui/penssidebar/brushpage.ui
+++ b/crates/rnote-ui/data/ui/penssidebar/brushpage.ui
@@ -12,16 +12,44 @@
     <child>
       <object class="GtkBox">
         <property name="orientation">vertical</property>
+        <property name="spacing">3</property>
         <child>
-          <object class="GtkMenuButton" id="brushstyle_menubutton">
-            <property name="direction">left</property>
-            <property name="tooltip_text" translatable="yes">Brush Style</property>
-            <property name="popover">brushstyle_popover</property>
-            <property name="icon-name">pen-brush-style-solid-symbolic</property>
+          <object class="GtkBox" id="brushstyle_togglebox">
+            <property name="orientation">vertical</property>
+            <property name="homogeneous">true</property>
+            <property name="vexpand">false</property>
             <style>
-              <class name="flat" />
-              <class name="sidebar_action_button" />
+              <class name="linked" />
             </style>
+            <child>
+              <object class="GtkToggleButton" id="brushstyle_marker_toggle">
+                <property name="tooltip_text" translatable="yes">Marker</property>
+                <property name="icon-name">pen-brush-style-marker-symbolic</property>
+                <style>
+                  <class name="sidebar_action_button" />
+                </style>
+              </object>
+            </child>
+            <child>
+              <object class="GtkToggleButton" id="brushstyle_solid_toggle">
+                <property name="group">brushstyle_marker_toggle</property>
+                <property name="tooltip_text" translatable="yes">Solid</property>
+                <property name="icon-name">pen-brush-style-solid-symbolic</property>
+                <style>
+                  <class name="sidebar_action_button" />
+                </style>
+              </object>
+            </child>
+            <child>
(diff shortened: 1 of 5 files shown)


JemmeCh:
Hello! I just started to use RNote and I was bumped out that there's no highligther tool. I spent the past 30mins reading about all the issues/merge requests you made and I think you found a good solution, especially this new UI design. It makes it easier to switch tools and removes convoluted submenues. 

You mention in #1602 that your highlighter proposition adds an alpha value to the marker. In the spirit of this UI redesign, I think the seperation of the marker and the highlighter could be better since switching between highlighter mode and marker mode would be easier. 

Maybe I'm missing how the actual switch would occur in practice... Would you be so kind to show us how it would work?

Kneemund:
> You mention in https://github.com/flxzt/rnote/pull/1602 that your highlighter proposition adds an alpha value to the marker.

Hi, my set of PRs do not involve any alpha value modifications, the highlighter can use colors of any opacity. It uses multiplicative blending and also works with fully opaque colors.

> Maybe I'm missing how the actual switch would occur in practice... Would you be so kind to show us how it would work?

Here is the demo of the workflow I propose. 
- I'm switching from the regular brush to the highlighter brush in the left sidebar, which also changes the stroke width (it is saved for every brush individually).
- I'm using the default, fully opaque colors which I would also use with the regular brush.

https://github.com/user-attachments/assets/333d305d-b270-4e51-b25c-cf29f0f436f3

JemmeCh:
> > You mention in #1602 that your highlighter proposition adds an alpha value to the marker.
> 
> Hi, my set of PRs do not involve any alpha value modifications, the highlighter can use colors of any opacity. It uses multiplicative blending and also works with fully opaque colors.
> 
> > Maybe I'm missing how the actual switch would occur in practice... Would you be so kind to show us how it would work?
> 
> Here is the demo of the workflow I propose.
> 
>     * I'm switching from the regular brush to the highlighter brush in the left sidebar, which also changes the stroke width (it is saved for every brush individually).
> 
>     * I'm using the default, fully opaque colors which I would also use with the regular brush.
> 
> 
> highlighter-full.mp4

Ahhhhhh ok I see! So the current Marker brush with description "Mark underneath other strokes" would be replaced by your implementation. Thank you for the clarification, I really appreciate it!

TheJanzap:
This looks great and I love to have this in Rnote. Is there still work to do so this can be merged?

Kneemund:
> This looks great and I love to have this in Rnote. Is there still work to do so this can be merged?

These UI changes *could* be merged, but the actual highlighter implementation is blocked by an obscure bug described in #1730, which seems to be caused by a (barely maintained) dependency :/

TheJanzap:
> These UI changes _could_ be merged, but the actual highlighter implementation is blocked by an obscure bug described in #1730, which seems to be caused by a (barely maintained) dependency :/

Hmm, so there seems to be little chance the bug is getting fixed in Cairo? I guess forking or switching dependencies is also out of scope for this. Sad, because I would really like to see marking in PDFs to work.

But even just merging this PR without the highlighter in the other PR would be useful, as it reduces the amount of taps/clicks needed to switch between different pen types and thus making work more seamless.

So, if possible without too much additional work, could this be merged without the highlighter support?

Verantor:
I think from a usabilty standpoint it would be great if each remembered their last selected color so one can quickly highlight and switch back to writing normaly. Otherwise great feature

TheJanzap:
> I think from a usabilty standpoint it would be great if each remembered their last selected color so one can quickly highlight and switch back to writing normaly. Otherwise great feature

That would be very convenient! The pen size is already remembered when switching brushes, so also remembering colors would be greatly appreciated!

PR #1752, opened 2026-04-30
feat: Enable editing width of existing strokes

### LINKED ISSUE #1039: After selection want to change the width of line
To change the width of line by selecting the writing.

After selection the drwing or writing cannot change the width of the drawing.

After selection the object there must be **line width change ** option.

**Additional context**  
change

### LINKED ISSUE #1626: Add option to toggle scale widths of brushstrokes and shapes when resizing
When resizing drawings, I'd like to be able to keep a consistent line size.

An option to toggle to feature added in #808

cyates5:
This merge request adds a popup menu to the selection sidebar, which allows the width of the selected strokes to be edited. The widths of selected brush and shape strokes are changed, while other stroke types are unaffected.

Resolves #1039.
Partially resolves #1626 , by allowing the width to be manually changed back after an object is scaled.

## Demo:

https://github.com/user-attachments/assets/c268c3bc-1637-4122-99e8-0f3d440ac925

## Todo:

- [ ] Use a more suitable icon for the menu button.
- [ ] Add keyboard shortcuts to open/close the menu.

Additionally, there are 2 problems I've not yet been able to resolve:

The 3 width selector buttons are deselected when the popup is closed to allow them to be clicked again next time it is opened. The issue is that the deselect animation only starts when the popup is next opened, and I don't know how to stop this animation from playing:

https://github.com/user-attachments/assets/bd19de56-2986-4fa3-970b-bc7aa9d9ddb8

Also, I'd like to save the size of the selector buttons between application restarts, like in the other width selectors, but I haven't been able to figure out how to do this yet.

Any advice on these issues would be greatly appreciated. I would also like to confirm whether the implementation so far looks suitable, or if there are any changes needed.

Thanks for making such a great app!

### DIFF
--- a/crates/rnote-ui/src/penssidebar/selectorpage.rs
+++ b/crates/rnote-ui/src/penssidebar/selectorpage.rs
@@ -1,14 +1,20 @@
 // Imports
 use crate::RnAppWindow;
+use crate::RnStrokeWidthPicker;
 use gtk4::{
-    CompositeTemplate, ToggleButton, Widget, glib, glib::clone, prelude::*, subclass::prelude::*,
+    Button, CompositeTemplate, MenuButton, Popover, ToggleButton, Widget, glib, glib::clone,
+    prelude::*, subclass::prelude::*,
 };
+use rnote_engine::pens::pensconfig::BrushConfig;
+use rnote_engine::pens::pensconfig::brushconfig::SolidOptions;
 use rnote_engine::pens::pensconfig::selectorconfig::SelectorStyle;
+use std::cell::Cell;
+use std::str;
 
 mod imp {
     use super::*;
 
-    #[derive(Default, Debug, CompositeTemplate)]
+    #[derive(Debug, CompositeTemplate)]
     #[template(resource = "/com/github/flxzt/rnote/ui/penssidebar/selectorpage.ui")]
     pub(crate) struct RnSelectorPage {
         #[template_child]
@@ -21,6 +27,34 @@ mod imp {
         pub(crate) selectorstyle_intersectingpath_toggle: TemplateChild<ToggleButton>,
         #[template_child]
         pub(crate) resize_lock_aspectratio_togglebutton: TemplateChild<ToggleButton>,
+        #[template_child]
+        pub(crate) strokewidth_menubutton: TemplateChild<MenuButton>,
+        #[template_child]
+        pub(crate) strokewidth_popover: TemplateChild<Popover>,
+        #[template_child]
+        pub(crate) strokewidth_popover_close_button: TemplateChild<Button>,
+        #[template_child]
+        pub(crate) stroke_width_picker: TemplateChild<RnStrokeWidthPicker>,
+
+        pub(crate) allow_stroke_width_changes: Cell<bool>, // Used to stop the stroke width changing immediately when the popup is opened
+    }
+
+    impl Default for RnSelectorPage {
+        fn default() -> Self {
+            Self {
+                selectorstyle_polygon_toggle: TemplateChild::default(),
+                selectorstyle_rect_toggle: TemplateChild::default(),
+                selectorstyle_single_toggle: TemplateChild::default(),
+                selectorstyle_intersectingpath_toggle: TemplateChild::default(),
+                resize_lock_aspectratio_togglebutton: TemplateChild::default(),
+                strokewidth_menubutton: TemplateChild::default(),
+                strokewidth_popover: TemplateChild::default(),
+                strokewidth_popover_close_button: TemplateChild::default(),
+                stroke_width_picker: TemplateChild::default(),
+
+                allow_stroke_width_changes: Cell::new(true),
+            }
+        }
     }
 
     #[glib::object_subclass]
@@ -71,6 +105,10 @@ impl RnSelectorPage {
         glib::Object::new()
     }
 
+    pub(crate) fn strokewidth_menubutton(&self) -> MenuButton {
+        self.imp().strokewidth_menubutton.get()
+    }
+
     #[allow(unused)]
     pub(crate) fn selector_style(&self) -> Option<SelectorStyle> {
         if self.imp().selectorstyle_polygon_toggle.is_active() {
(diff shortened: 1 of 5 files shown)


crates/rnote-ui/src/penssidebar/selectorpage.rs
```diff
@@ -21,6 +27,34 @@ mod imp {
...
+        #[template_child]
+        pub(crate) strokewidth_menubutton: TemplateChild<MenuButton>,
+        #[template_child]
+        pub(crate) strokewidth_popover: TemplateChild<Popover>,
+        #[template_child]
+        pub(crate) strokewidth_popover_close_button: TemplateChild<Button>,
+        #[template_child]
+        pub(crate) stroke_width_picker: TemplateChild<RnStrokeWidthPicker>,
+
+        pub(crate) allow_stroke_width_changes: Cell<bool>, // Used to stop the stroke width changing immediately when the popup is opened
```
Doublonmousse, line 39: I want to do something different for this.

The Cell means we cannot derive the default, is that correct ?

crates/rnote-ui/src/penssidebar/selectorpage.rs
```diff
@@ -180,6 +222,92 @@ impl RnSelectorPage {
...
+                    return;
+                };
+                let stroke_width = canvas
+                    .engine_ref()
+                    .get_selection_stroke_width()
+                    .unwrap_or(2.0); // Default to width of 2 if no valid strokes selected
+
+                imp.allow_stroke_width_changes.set(false); // Stop the new picker width from applying immediately
+                imp.stroke_width_picker
+                    .set_stroke_width(stroke_width + 0.001); // A small value is added to allow the user to apply a width equal to 'stroke width'
```
Doublonmousse, line 269: I'm not getting why the +0,001 is needed

crates/rnote-ui/src/penssidebar/selectorpage.rs
```diff
@@ -180,6 +222,92 @@ impl RnSelectorPage {
...
+                    return;
+                }
+
+                let Some(canvas) = appwindow.active_tab_canvas() else {
+                    return;
+                };
+                let stroke_width = canvas
+                    .engine_ref()
+                    .get_selection_stroke_width()
+                    .unwrap_or(2.0); // Default to width of 2 if no valid strokes selected
```
Doublonmousse, line 265: Another way to do this would be to have the button be sensitive depending on the selection state. So no way to actually open the popover if there is nothing valid to change the stroke width on. 

This is a little more involved as far as code goes though

crates/rnote-ui/src/penssidebar/selectorpage.rs
```diff
@@ -180,6 +222,92 @@ impl RnSelectorPage {
...
+                    // Return if the popup has just been opened
+                    if !imp.allow_stroke_width_changes.get() {
+                        imp.allow_stroke_width_changes.set(true);
+                        return;
+                    }
+
+                    // Change the width of the selected strokes
+                    let widget_flags = canvas
+                        .engine_mut()
+                        .change_selection_stroke_width(stroke_width);
```
Doublonmousse, line 305: Another way to do this withough the cell is to do the following

- Call `get_selection_stroke_width` 
- Compare the stroke width value to the `stroke_width`
- If they are not identical (to some precision) then change the stroke width with `change_selection_stroke_width`

Doublonmousse (commented):
I'm not 100 % sure as far as UI goes for this (is this the best place for it/should we place this here etc...) so I'm not guaranteeing this will be the solution we keep for editing the stroke width. 

It's best to wait for another review for that aspect (not by me at least)

I do have some other remarks though

> Also, I'd like to save the size of the selector buttons between application restarts, like in the other width selectors, but I haven't been able to figure out how to do this yet.

See https://github.com/flxzt/rnote/blob/main/crates/rnote-ui/src/appwindow/appsettings.rs#L309-L319 for the setting permanence (note this also means changing the https://github.com/flxzt/rnote/blob/main/crates/rnote-ui/data/app.gschema.xml.in as well.

Not sure whether keeping the same settings as the brushstroke makes sense or not (if they are separate then they move independently, if not changing it in one place changes it in another so ... maybe this wouldn't be so clear and accidentally changing one of your preset by resizing the stroke width could happen if one of the setter is selected).

Kneemund:
> I'm not 100 % sure as far as UI goes for this (is this the best place for it/should we place this here etc...) so I'm not guaranteeing this will be the solution we keep for editing the stroke width.

I feel like from a UX perspective, having a unified size picker in the top bar next to the colors would be a very clean solution:

- We have some kind of width in almost every tool: brush stroke width, shape stroke width, text size, eraser size, editing stroke width in selection (this).
- Stroke colors are also edited by selecting the stroke and picking the new color in the top bar. This would be analogous to that.

PS: Just putting this up for discussion, not saying that it *should* be done this way.

crates/rnote-ui/src/penssidebar/selectorpage.rs
cyates5: That is correct. 

> Another way to do this withough the cell is to do the following
> 
>     * Call `get_selection_stroke_width`
> 
>     * Compare the stroke width value to the `stroke_width`
> 
>     * If they are not identical (to some precision) then change the stroke width with `change_selection_stroke_width`

This seems like a much simpler method, thanks for the suggestion.

cyates5:
> I'm not getting why the +0,001 is needed

When the selected strokes have different widths, the width picker displays one of them in the input box. If the user wants to apply this width to all selected strokes, they would need to change the value in the input box to something different, then change it back.

The +0.001 helps with this, as it lets the user apply the width to all selected strokes by selecting the input box and pressing 'enter', as the rounding of the value is detected as a change.

I'm not particularly happy with this solution though, and on further reflection it doesn't resolve the issue for pen/touch input methods. I'm thinking of creating a custom width picker widget instead, with the following behavior: 

If all selected strokes have the same width:
- Width picker displays the width and behaves like usual.

If selected strokes have different widths:
- Display range of selected stroke widths e.g. "5.0 - 8.0" or other symbol e.g. "..." in the input box.
-  If "+" first clicked, set all strokes to largest width in range.
- If "-" first clicked, set all strokes to smallest width in range.
- If the input box is clicked, clear the box to allow a new value to be entered.

It will likely be a week or so before I have time to work on this, but does it sound okay as a solution?

PR #1758, opened 2026-05-01
add the rnote cli to the windows inno build

### LINKED ISSUE #1757: Add rnote-cli to windows install
### Discussed in https://github.com/flxzt/rnote/discussions/1744

<sup>Originally posted by **6801318d8d** April 18, 2026</sup>
I would like to convert a bunch of xopp files to rnote from the command line.

I have Windows 11.

I read in the README that there is supposed to be a rnote-cli. But seems available only on flatpak?

There is not `rnote-cli` for me.

How can I achieve this task in Windows?</div>

## Inno build

- [ ] At the very least we could add the rnote-cli.exe to the `[Files]` section, maybe this is enough to at least launch the cli on windows.
- [ ] Add to path for convenience ?

Doublonmousse:
CI test on : https://github.com/Doublonmousse/rnote/actions/runs/25228946601

Fixes #1757 

TODO 
- [x] cleanup the script (does it work with `-Dcli=false`? I'm not sure whether it fails when failing to copy a file or how to adapt the script more cleanly for this)
- [x] add to path? For now this is in the bin folder, not in the window path

### DIFF
--- a/build-aux/inno_build.py
+++ b/build-aux/inno_build.py
@@ -15,6 +15,7 @@ app_name_capitalized = sys.argv[5]
 app_id = sys.argv[6]
 ui_output = sys.argv[7]
 inno_script = sys.argv[8]
+define_cli_output = sys.argv[9]
 
 print(f"""
 ### executing Inno-Setup installer build script with arguments: ###
@@ -26,6 +27,7 @@ print(f"""
     app_id: {app_id}
     ui_output: {ui_output}
     inno_script: {inno_script}
+    (optional) define cli {define_cli_output}
 """, file=sys.stderr)
 
 def run_command(command, error_message):
@@ -138,7 +140,7 @@ for file in os.listdir(app_mo_dir):
 print("Running ISCC...", file=sys.stderr)
 
 run_command(
-    f"iscc {inno_script}",
+    f"iscc {inno_script} {define_cli_output}",
     "Running ISCC failed"
 )
 
--- a/build-aux/inno_build.py
+++ b/build-aux/inno_build.py
@@ -29,7 +29,7 @@ print(f"""
     app_id: {app_id}
     ui_output: {ui_output}
     inno_script: {inno_script}
-    (optional) define cli {cli_name}
+    cli_name: {cli_name} (empty if not packaged)
 """, file=sys.stderr)
 
 def run_command(command, error_message):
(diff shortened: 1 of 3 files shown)


flxzt:
Not sure if it is possible with Inno, but I've seen installers asking whether they can add a path entry to `PATH`. Would be cool if we could do the same!

Doublonmousse:
Now this works with `-Dcli=false` calling manually meson.
 
The only part that doesn't work is the `just build-win-installer` command if the cli = false option is selected as this invokes 
```just
build:
    meson compile ui-cargo-build -C {{ build_folder }}
    meson compile cli-cargo-build -C {{ build_folder }}
```
and the second command doesn't exist if the cli = false option is selected. It errors with
```
ERROR: Can't invoke target `cli-cargo-build`: target not found
error: Recipe `build` failed on line 186 with exit code 1
```

Doublonmousse:
Seems like adding to path as an option is possible: see https://github.com/mentebinaria/retoolkit/blob/f86f5e608836b898da36a3787eb5135d2baef976/src/installer/retoolkit.iss#L214

Doublonmousse:
Updated to include the optional path update in the installer

image


Window release CI: https://github.com/Doublonmousse/rnote/actions/runs/25447031215

build-aux/inno_build.py
```diff
@@ -15,6 +15,7 @@
 app_id = sys.argv[6]
 ui_output = sys.argv[7]
 inno_script = sys.argv[8]
+define_cli_output = sys.argv[9]
```
flxzt: In my opinion this could be named more clear.
And consider adding some documentation to explain that this must be set as inno `\D<cli.exe>` argument to make the installer package the CLI executable.

Doublonmousse:
Waiting for the window release CI to pass : https://github.com/Doublonmousse/rnote/actions/runs/29529431184 (and to double check we do get the cli.exe)

build-aux/inno_build.py
Doublonmousse: I've changed so that I pass the cli name instead (only if we do build it) and the whole `/D` can be done in this script instead

flxzt: looks better now!

flxzt:
looks good now :+1: Tested the artifact and confirmed the CLI is bundled and added to PATH on installation.

PR #1763, opened 2026-05-09
fix: stylus hover triggering canvas pan

### LINKED ISSUE #1468: Holding stylus secondary button pans canvas when hovering
When I hold down the secondary button and hover my stylus above the screen without touching it, moving the stylus pans the canvas. When pressing down, select works as expected without any panning.

I believe this is due to the secondary button on the stylus being bound to middle click.

I am using the Lenovo Digital Pen 2 and my OS is Linux Mint.

alberto-is:
Fixes #1468
What changed

Prevent canvas pan gestures from starting on stylus hover (pressure == 0), while keeping mouse/touch behavior unchanged.

Why
On some stylus devices, pressing a pen button while hovering can trigger drag gestures, moving the canvas unexpectedly. This filters out stylus hover events so the assigned shortcut still works without causing unintended panning.
Testing

cargo fmt --check
meson compile ui-cargo-clippy -C _mesonbuild
Manual: verified canvas pan still works with mouse drag
Manual: confirmed pen button no longer triggers unintended pan on stylus hover

Note: clippy reports existing warnings in crates/rnote-ui/src/dialogs/import.rs — not introduced by this change.
Environment / Device

Device: Wacom Intuos S Pad + Pen
OS: Arch Linux x86_64
Kernel: 7.0.3-arch1-2
DE/WM: KDE Plasma 6.6.4 / KWin
Session: Wayland (XDG_SESSION_TYPE=wayland)
Rnote: 0.14.2
GTK: 4.22.4
Libadwaita: 1.9.0


crates/rnote-ui/src/canvaswrapper.rs
```diff
@@ -415,10 +415,20 @@ mod imp {
...
                     obj,
-                    move |_, _, _| {
+                    move |gesture, _, _| {
                         // We don't claim the sequence, because we we want to allow touch zooming.
                         // When the zoom gesture is recognized, it claims it and denies this touch drag gesture.
 
+                        if let Some(event) = gesture.current_event()
+                            && event.device_tool().is_some()
+                        {
+                            let pressure = event.axis(gdk::AxisUse::Pressure).unwrap_or(0.0);
```
Doublonmousse: I wonder if it's better to `unwrap_or(0.5)` so a pen that doesn't have pressure (or where gtk4 can't read it for some reason) still can do the drag gesture even with the button bug.

alberto-is: That's true, I didn't think about the gestures. I already updated it (tested, and the fix for the stylus is still working).

Kneemund: Why do we need `unwrap_or` and a constant default value in the first place? Why not:

```rs
let pressure = event.axis(gdk::AxisUse::Pressure);
if pressure.is_some_and(|p| p <= 0.0) {
    gesture.set_state(EventSequenceState::Denied);
    return;
}
```

alberto-is: That's cleaner, so I've applied the change.

proudmuslim-dev:
Been using this patch for about a week with no issues (albeit on the latest release rather than the main branch)

Device: Wacom Intuos Pro M (2017)
OS: NixOS Unstable
Kernel: 7.0.9 (mainline)
App Version: Rnote v0.14.2
Installation Source: Nixpkgs unstable
GTK: 4.22.4
Libadwaita: 1.9.0
WM: Swaywm (v1.12)
Session: Wayland (XDG_SESSION_TYPE=wayland)

Doublonmousse (approved):

flxzt:
thank you for the contribution!

Issue #1766, opened 2026-05-12
Tab moving touch gestures in terms of multiple windows


der-wasguckstdu-dreier:
**Please submit only one feature request in one issue!**  
If it is UI/UX related, don't post how a single aspect of the UI should be changed, rather in which use-case you felt
the UI is lacking or should be improved.

First of all, I enjoy that there even are tabs, however I think it is kind of finicky to move them to new windows. 

**Is your feature request related to a problem? Please describe.**  
A clear and concise description of what the problem is.

If I want a new window with my desired tab, there is no single button (or gesture like in browsers) to create a new window. There are simply too many clicks necessary for having two already open tabs on both halfs of the screen.

**Describe the solution you'd like**  

- Grabbing a tab and swiping down: new window with the grabbed tab
- Tab view showing tabs from all open windows
- maybe a "split" button where the currently open file can be opened two times inside of one rNote window. This would enhance practicability.

**Describe alternatives you've considered**  

Dont know

**Additional context**  

I am a window hoarder so its sometimes very hard to find the related rNote file to another one. I'd like to have them both in one window so that I don't have to cycle through all open windows just to find this one note.
For this a recent activities tab would be cool also.

Doublonmousse:
> - Grabbing a tab and swiping down: new window with the grabbed tab

I think there's some implementations missing for this to create a new window when dragging.

> - Tab view showing tabs from all open windows

Not sure this is feasible nor something we'd want (that's not standard behavior for tabs)

> - maybe a "split" button where the currently open file can be opened two times inside of one rNote window. This would enhance practicability.

That's #150

PR #1768, opened 2026-05-14
feat: add callback to tabview `connect_create_window` to make detaching tab works

Doublonmousse:
There was a missing implementation for 
```rs
imp.tabview.connect_create_window
```
(see https://docs.rs/libadwaita/latest/libadwaita/struct.TabView.html#method.connect_create_window)
that should create a new window and return its tabview (where the dragged tab will be inserted)

This prevented tabs from detaching from the main window with the error message
```
AdwTabView::create-window handler must not return NULL
```

Now, there is still another error message
```
(rnote:17037): Gtk-CRITICAL **: 15:50:17.281: New application windows must be added after the GApplication::startup signal has been emitted.
```
but dragging tabs out create a new window.

Demo : 

[Screen Recording 2026-05-14 at 16.14.42.webm](https://github.com/user-attachments/assets/f7f92d7a-b93d-4dde-ba1a-8e67e9b51c96)

Though there are still some things to fix
- [x] ~~The current method doesn't really work~~
   - ~I create a new `RnApp` so I'm not 100 % sure that the full setup done in `main()` in addition to the creating of the `RnApp` is propagated to the new child window.~
   - ~This also doesn't fully work as closing the first window closes the second one.  Ideally I'd have an `RnApp` reference accessible from the `imp.tabview.connect_create_window` call~
   - ~Not sure how to call from the appwindow the creating of a new window and also get the tabview back in the callback (with a signal a new window can be created but not sure I can get the tab view back ?)~
- [x] Find why /fix the new error message 
- [x] Maybe it's worth adding a right click `Move to new window` like the adwaita demo 

image

Related : #1766

### DIFF
--- a/crates/rnote-ui/src/app/mod.rs
+++ b/crates/rnote-ui/src/app/mod.rs
@@ -154,7 +154,7 @@ mod imp {
             }
         }
 
-        pub(crate) fn new_appwindow_init_return_tab(&self)->adw::TabView{
+        pub(crate) fn new_appwindow_init_return_tab(&self) -> adw::TabView {
             let appwindow = RnAppWindow::new(self.obj().upcast_ref::<gtk4::Application>());
             appwindow.init(false);
 
@@ -166,9 +166,7 @@ mod imp {
             window_group.add_window(&appwindow);
 
             appwindow.present();
-            // this isn't enough
-            // (rnote:11838): Gtk-CRITICAL **: 14:08:17.296: New application windows must be added after the GApplication::startup signal has been emitted.
-            return appwindow.overlays().tabview()
+            return appwindow.overlays().tabview();
         }
     }
 }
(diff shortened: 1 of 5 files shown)


Doublonmousse:
I've found how to get an instance of the app using https://gtk-rs.org/gtk4-rs/git/docs/gtk4/struct.ApplicationWindow.html#method.application. Now everything works

crates/rnote-ui/src/app/mod.rs
```diff
@@ -153,6 +153,21 @@ mod imp {
...
         }
+
+        pub(crate) fn new_appwindow_init_return_tab(&self) -> adw::TabView {
+            let appwindow = RnAppWindow::new(self.obj().upcast_ref::<gtk4::Application>());
+            appwindow.init(false);
+
+            // create a window group for each app window
+            // to make modals only impact the current app
+            // window.
+            // See issue #1461
```
flxzt: Please add a real link here instead

flxzt: I added this myself in an additional commit

flxzt:
Thanks, looks good!

Issue #1217, opened 2024-09-14
Enormous PDFs Exports


sortedcord:
I have a rnote file that mainly consists of ink and a small picture (5 pages) and when I export it directly as a PDF, rnote created a PDF file that is 12Mb in size, however, if I export it to Xournalpp and then export it to pdf the file size only comes out to be 2Mb in size? Also I wanted to report that the PDF file exported by rnote is really laggy to view on other PDF viewers, especially when scrolling to the next page, there is a significant lag and the pdf viewer (browser) almost freezes up.

**Screenshots**  

**Desktop (please complete the following information):**  
 - OS: Endeavour OS
 - App Version: Rnote v0.11.0
 - Installation Source: Archlinux Community Repo
 - Desktop Environment: KDE Plasma 6.1.4
 - Display Server: Wayland
 - Input Source: Mouse and Keyboard,  External Graphics Tablet

![image](https://github.com/user-attachments/assets/62ea9ff6-ae2f-4bef-9243-d129b267dd0b)

file in question:
[test.zip](https://github.com/user-attachments/files/17002466/test.zip)

Doublonmousse:
Seems like xournal does some sort of compression when creating the output (maybe some quantization of points based on the dpi ?).
Comparing the two pdf and you see that there are visible differences 

![image](https://github.com/user-attachments/assets/8f55e9b0-de5c-406a-aac6-32ad8f4600e8)
(xournalpp on the left rnote on the right)

So that probably accounts for the size difference between the two. Does removing the background in rnote (and exporting a pdf version without the background) helps with the reader ?

There's probably a handful of things that could be improved in rnote regarding pdf size
- Try to see what xournalpp is doing and matching it (lossy compression of the strokes). Imo this should stay optional.
-  See what's possible to do with stroke simplification (find a curve with less points/data needed to construct it that stays close to the actual stroke) : no idea how feasible that is (what algorithm exist today ? for what performance cost ?) but at the very least having a compression that stays "smooth" seems more doable this way

And another part may be possible at the ink modeler stage. Currently the size of the file (and by extension the pdf) is linear in the number of pen events. But that's device dependent (refresh rate for pens may be between 60 and 480 Hz) so writing slowly means a larger file !
Hence the following may be possible
- Have heuristics in place to discard events that are too close to the previous one
- Have the smoothing algorithm serve as a compression step as well (and produce potentially less events than the number of events sent by the pen)

[...]

Issue #269, opened 2022-09-29
Size of exported PDF is very large


amit9838:
- I recently exported my document with 10 pages and the size of exported PDF was 13MBs.
- It would be great if we can choose choose the PDF Quality during export or simply export PDFs with smaller size.

luch0ya:
I would like to second the petition, and include as current work around what is suggested in #334: Export as xopp and then export as PDF from Xournal++ (don't know if for SVG works the same).

PR #1769, opened 2026-05-15
feat: brush stroke simplification

Kneemund:
- Adds simplification to brush strokes, reducing file size and improving performance.
- Addresses #269 and #1217.
- Should be merged after #1775.

The simplification mode (none and this) and quality (low, medium, high) can be chosen in the brush configuration menu. They default to being enabled with medium quality. The quality levels were eyeballed with the following goals (subjective, feel free to test and propose changes):

- High - No visible changes.
- Medium - Very minor changes in very few (especially degenerate) cases if you zoom in.
- Low - Minor changes if you zoom in.

The file size savings naturally depend on the strokes, so they're difficult to quantify. Keeping that in mind, running this on handwritten lecture notes gave the following results:

|Quality|Path Elements|`.rnote`|`.pdf`|
|-|-|-|-|
|None|224177|1.984 MB|9.208 MB|
|High|98958 (44%)|1.011 MB (51%)|4.362 MB (47%)|
|Medium|49301 (22%)|0.535 MB (27%)|2.184 MB (24%)|
|Low|36489 (16%)|0.408 MB (21%)|1.601 MB (17%)|

### Implementation Details

- Simplification is based on the [Ramer–Douglas–Peucker algorithm](https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm), but additionally considers pressure changes (so that e.g. straight line segments can be composed out of multiple points if pressure varies significantly).
- The amount of simplification is controlled through two constants, `geometry_epsilon` and `pressure_epsilon`. They specify how much the position and pressure are allowed to deviate from the original stroke.

### Discussion / TODO

- [ ] `geometry_epsilon` and `pressure_epsilon` are currently eyeballed magic numbers. Should the `geometry_epsilon` scale depending on the camera zoom? Should the `pressure_epsilon` scale depending on the stroke width?
- [ ] We could optimize existing files somehow by adding it as an action somewhere. (Where?)
- [ ] The appearance of textured strokes changes after finishing a stroke (because the simplification removes points). I'm not sure what to do about that.
- [ ] We could add an export option that allows for more lossy simplification.

### Demo (Constant Pressure)
[Bildschirmaufnahme_20260515_160451.webm](https://github.com/user-attachments/assets/9e28ada3-2f01-4f8f-b313-7a01936fc053)

### DIFF
--- a/crates/rnote-engine/src/pens/brush.rs
+++ b/crates/rnote-engine/src/pens/brush.rs
@@ -261,6 +261,7 @@ impl PenBehaviour for Brush {
                         if let Some(Stroke::BrushStroke(brushstroke)) =
                             engine_view.store.get_stroke_mut(*current_stroke_key)
                         {
+                            brushstroke.path.simplify_rdp(0.1);
                             brushstroke.style = engine_view
                                 .config
                                 .pens_config
--- a/crates/rnote-engine/src/pens/brush.rs
+++ b/crates/rnote-engine/src/pens/brush.rs
@@ -261,7 +261,13 @@ impl PenBehaviour for Brush {
                         if let Some(Stroke::BrushStroke(brushstroke)) =
                             engine_view.store.get_stroke_mut(*current_stroke_key)
                         {
-                            brushstroke.path.simplify(0.1, 0.15);
+                            engine_view
+                                .config
+                                .pens_config
+                                .brush_config
+                                .simplification_options
+                                .simplify(&mut brushstroke.path);
+
                             brushstroke.style = engine_view
                                 .config
                                 .pens_config
--- a/crates/rnote-engine/src/pens/brush.rs
+++ b/crates/rnote-engine/src/pens/brush.rs
@@ -261,7 +261,7 @@ impl PenBehaviour for Brush {
                         if let Some(Stroke::BrushStroke(brushstroke)) =
                             engine_view.store.get_stroke_mut(*current_stroke_key)
                         {
-                            brushstroke.path.simplify_rdp(0.1);
+                            brushstroke.path.simplify(0.1, 0.15);
                             brushstroke.style = engine_view
                                 .config
                                 .pens_config
--- a/crates/rnote-compose/src/meson.build
+++ b/crates/rnote-compose/src/meson.build
@@ -27,6 +27,7 @@ rnote_compose_sources = files(
     'penpath/element.rs',
     'penpath/mod.rs',
     'penpath/segment.rs',
+    'penpath/simplification.rs',
     'serialize.rs',
     'shapes/arrow.rs',
     'shapes/cubbez.rs',
--- a/crates/rnote-compose/src/penpath/mod.rs
+++ b/crates/rnote-compose/src/penpath/mod.rs
@@ -301,61 +301,68 @@ pub(crate) fn no_subsegments_for_segment_len(len: f64) -> i32 {
     }
 }
 
+/// Squared distance from point p to the line segment ab.
+fn point_segment_distance_sq(p: na::Vector2<f64>, a: na::Vector2<f64>, b: na::Vector2<f64>) -> f64 {
+    let ab = b - a;
+    let ab_len_sq = ab.norm_squared();
+
+    if ab_len_sq == 0.0 {
+        return (p - a).norm_squared();
+    }
+
+    let t = ((p - a).dot(&ab) / ab_len_sq).clamp(0.0, 1.0);
+    (p - (a + ab * t)).norm_squared()
+}
+
 /// Ramer-Douglas-Peucker simplification for a slice of `Element`.
-/// 
+///
 /// https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm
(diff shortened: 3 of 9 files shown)


crates/rnote-engine/src/pens/brush.rs
```diff
@@ -261,6 +261,7 @@ impl PenBehaviour for Brush {
                         if let Some(Stroke::BrushStroke(brushstroke)) =
                             engine_view.store.get_stroke_mut(*current_stroke_key)
                         {
+                            brushstroke.path.simplify(0.1, 0.15);
```
Doublonmousse: I'd rather have the simplify step be opt-in or dependent on the brush (it's not expected to have simplification when choosing the simple path modeling with `produces line segments from the raw input` imo).

Some people will prefer (and very strongly so for or against) unfiltered raw input or would dislike having post processing that changes the aspect of the stroke after finishing it (the `Modeled` Path modelling mode is for example trying to smooth things without going back on what was smoothed -with the `update`/`predict` API for final/temporary points -).

It _could_ be a (or multiple ?) `Path Modelling` options in the brush configuration toolbar or a separate section (at most 6 for on/off if we do that). The upside is that choosing Simple -> raw will work as expected

Or a new post processing option with a toggle (on/off + maybe values ?). (But it could be confusing if you choose Simple for path modelling and end up with simplification anyway)

Doublonmousse:
I was exploring other algorithms for this recently (also on a post processing basis), like Schneider's (see Schneider, Philip J. "An algorithm for automatically fitting digitized curves." Graphics gems. 1990. 612-626. https://scholar.google.com/scholar?cluster=17652956975514389477&hl=en&as_sdt=0,5).

Not sure which one is better between the two (Schneider outputs splines and it ideally needs some initial preprocessing to precut spline approximation around large angle changes)

Note : xournalpp has an ancient PR on this https://github.com/xournalpp/xournalpp/pull/2729/changes with both post processing and a live approach (the live approach tries to grow a single curve approximation starting from the first point to each new one, and will keep it unless the error is too large in which case the previous best curve is pushed and a new one is started).

On xournal's side the PR is blocked by their file format (no spline) whereas we do have this in our file format.

Note : this is disregarding how to deal with pressure

Doublonmousse:
> We could optimize existing files by introducing a new file format version.

I'm not sure how this relates to simplification. By Applying simplification on previous rnote file ?

crates/rnote-engine/src/pens/brush.rs
Kneemund: Maybe I misunderstand the path modelling variants, but I think that simplification does not really contradict simple/no modelling. Ideally, simplification does not affect the appearance of the stroke. But I do see how it could be confusing given the current description...

I think it should be configurable and opt-out (not opt-in) in the brush settings. Either we expose the parameters directly or add presets.

Kneemund:
Schneider's algorithm also looks interesting. Depending on how it works, it could be a lot harder to deal with pressure. I don't think we have to decide for one over the other, maybe we can implement both and let the user choose (like with path modelling).

Would you say that we need something similar to the live approach? We already do some kind of post processing (or just re-rendering?), so I think it's not really necessary.

> I'm not sure how this relates to simplification. By Applying simplification on previous rnote file ?

Yes, we could apply this to existing strokes if we increment the file format version.

Also, we could add an export option that allows for more lossy simplification.

Doublonmousse:
> Schneider's algorithm also looks interesting. Depending on how it works, it could be a lot harder to deal with pressure. I don't think we have to decide for one over the other, maybe we can implement both and let the user choose (like with path modelling).

Agreed. I could try to do that at some point, not sure what is better here (although it's probably slower than [Ramer–Douglas–Peucker](https://en.wikipedia.org/wiki/Ramer%E2%80%93Douglas%E2%80%93Peucker_algorithm)).

There's also a couple of simplifying functions in kurbo we can also use directly with [simplify_bezpath](https://docs.rs/kurbo/latest/kurbo/simplify/fn.simplify_bezpath.html) calling to [fit_to_bezpath_opt](https://docs.rs/kurbo/latest/kurbo/fn.fit_to_bezpath_opt.html) or [fit_to_bezpath](https://docs.rs/kurbo/latest/kurbo/fn.fit_to_bezpath.html) depending on the settings. They are similar to Schneider (recursive) though with some differences (different fit method and approximation of the error ...).
As we already have kurbo around, maybe we can play with these functions if we want another one to try, I don't expect these to be fundamentally different from applying Schneider's algorithm on them instead (though dealing with pressure maybe isn't easy from this).

> Yes, we could apply this to existing strokes if we increment the file format version.

In the spirit of doing opt-in/out, maybe it makes more sense to expose this somewhere as an option, not a file format increment (cause if you opt out there's no reason to have a difference on the file format)

> Would you say that we need something similar to the live approach? We already do some kind of post processing (or just re-rendering?), so I think it's not really necessary.

No, this would be something to implement as a mode to the ink stroke modeler in addition to the current one (the API surface is already there for it, with an adptation to export splines not points). I'm just curious to try whether this can give better results and feel when writing or not.

> Also, we could add an export option that allows for more lossy simplification.

That seems useful (though it has to come with a preview in that case).

PR #1773, opened 2026-05-16
fixup: fix setters for doc_background_pattern_width/height_unitentry …

### LINKED ISSUE #1258: Prompt to save changes of the the first default blank file created on each startup of rnote
**Describe the bug**  
A prompt to save file or ignore modifications appears when trying to close the default empty and unmodified document that is created at startup of rnote.

The first empty document created at start-up should be considered as empty and unmodified by default. (as it is the case for other empty document created when opening new tabs).

**To Reproduce 1**  
1.Start rnote
2.Try to close rnote
3.A popup appears prompting to save the modifications or ignore modifications of the empty document.

Or equivalently:

**To Reproduce 2**  
1.Start rnote.
2. Create a new empty document in a new tab. 
3. Try to close the first empty document of the initial tab.
3.A popup appears prompting to save the modifications or ignore modifications of the empty document.

**Expected behavior**  
This prompt should not appear as the document is empty and unmodified.

 - OS: Fedroa 40
 - App Version: Rnote v0.11.0]

Doublonmousse:
…on setup

Otherwise the following occurs :
- the unitentry starts with default dpi and value whereas the engine has the correct value
- setting the dpi first makes the unitentry recalculate with the new dpi but default value so a new value in px is applied to the engine through the connect_notify on "value"
- setting the value in px afterwards will recalculate the value (this time with the correct dpi/value/unit combo) so the value in px is correct but this triggers again a modification on the engine because we changed this value just before.

This thus triggers 2*2 updates of the engine upon startup. But changing the order isn't enough (as in that case the first change of the value will return the expected value in px but not the second one if the unit is not px (because of the dpi scaling changing when setting the dpi with a unit != from px.).

To fix this, create a setter that changes both dpi and value in a way where the value property is only set once with the correct value_in_px callback

Fixes #1258. Replaces #1453  

NOTE : the `else` part in `set_dpi_and_value_px` is not technically needed for the fix (I kinda figured it'd be better to have an helper function with all cases covered anyway, and that it's a little clearer this way)

### DIFF
--- a/crates/rnote-ui/src/unitentry.rs
+++ b/crates/rnote-ui/src/unitentry.rs
@@ -317,4 +317,36 @@ impl RnUnitEntry {
         self.set_dpi(dpi);
         self.set_value(value);
     }
+
+    /// sets both dpi and value such that the value property is
+    /// only changed once with the correct self.value_in_px() callback
+    pub(crate) fn set_dpi_and_value_px(&self, dpi: f64, val_px: f64) {
+        if self.unit() == MeasureUnit::Px {
+            // Note : setting value then dpi will work if and only if the unit is is px
+            // Then value_in_px returns self.value() directly
+            // because dpi is not a variable in the convert_measurement call
+
+            self.set_value_in_px(val_px);
+            self.set_dpi(dpi);
+        } else {
+            // NOTE : this is mostly for security
+            // as we default to px for now on startup for the unit
+            // this shouldn't matter too much (at most this will limit duplicated
+            // calls to the engine with erroneous value_px on the first call)
+
+            // silent set on the dpi (because the new dpi value == to the cell one,
+            // does not trigger a calculation on value again)
+            // hence both the property and the cell have the new value
+            self.imp().dpi.replace(dpi);
+            self.set_dpi(dpi);
+
+            self.set_value(MeasureUnit::convert_measurement(
+                val_px,
+                MeasureUnit::Px,
+                dpi,
+                self.unit(),
+                dpi,
+            ));
+        }
+    }
 }
--- a/crates/rnote-ui/src/unitentry.rs
+++ b/crates/rnote-ui/src/unitentry.rs
@@ -322,7 +322,7 @@ impl RnUnitEntry {
         self.set_value(value);
     }
 
-    /// Sets both dpi and value such that no intermittent wrong "value" 
+    /// Sets both dpi and value such that no intermittent wrong "value"
     /// change signals are emitted by the Widget.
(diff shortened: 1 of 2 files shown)


crates/rnote-ui/src/unitentry.rs
```diff
@@ -317,4 +317,36 @@ impl RnUnitEntry {
...
+        } else {
+            // NOTE : this is mostly for security
+            // as we default to px for now on startup for the unit
+            // this shouldn't matter too much (at most this will limit duplicated
+            // calls to the engine with erroneous value_px on the first call)
+
+            // silent set on the dpi (because the new dpi value == to the cell one,
+            // does not trigger a calculation on value again)
+            // hence both the property and the cell have the new value
+            self.imp().dpi.replace(dpi);
```
flxzt, line 340: I don't get why this is necessary, even with the comment.
It assumes some context I don't have, could it be reworded such that it becomes more clear?
Thanks!

Doublonmousse: - I want this set operation to only emit one change of the `value` property (otherwise we emit 2+ updates to the engine with probably a transient value that will probably be wrong)

- If I call `set_dpi`, I'll probably call this if the dpi changes
https://github.com/flxzt/rnote/blob/f3f7efd6d2256d6834c93ee4aa774fe2f8152e1b/crates/rnote-ui/src/unitentry.rs#L169-L177

- But I'll set the `value` property to the wrong value (because the unit is the same for the `value_unit` and `desired_unit`, for all `MeasurementUnit` different from px, this comes out to `self.value().get * value_dpi/desired_dpi` where `value_dpi` is the old dpi and `desired_dpi` the new one). Hence a change of the `value` property

- Then I'll change the value again to enforce the new value `val_px` in pixel units.

By setting the internal `dpi` property, I force the `if dpi != self.dpi.get()` to `false`, and the value property is only changed once at the end here, at which point both the `self.value_in_px()` callback will give the correct value (correct `value/dpi` properties)

flxzt: can you add some version of this as code comment instead?

I am still not sure I understand it completely though.

And what would nice, add code documentation for `set_dpi()` and `set_value()` that points developers to the  newly introduced function `set_dpi_and_value_px()`.

Something like:

```
**Note**: if both "value" and "dpi" should be changed at the same time, use [Self::set_dpi_and_value_px()] instead.
It ensures that no intermittent wrong "value" change signals are emitted by the Widget.
```

Doublonmousse: Done

PR #1774, opened 2026-05-17
fix(ui): use limit-events property for headerbar buttons

### LINKED ISSUE #1770: Replace #1143 by limit-events
#1143 used an event controller in between the button and the headerbar to prevent events not catched by buttons in the titlebar to be sent to the headerbar.

Now that we are on gtk 4.18 or more on all platforms and release, it seems we can use https://docs.gtk.org/gtk4/property.Widget.limit-events.html instead (so set this property on all buttons in the headerbar and/or the gtkbow above should do the same as the event controller).

Seems like a more proper longterm fix.

alberto-is:
Fixes #1770 (Previous #1142).

**Description**
This PR reverts the `EventControllerLegacy` workaround introduced in #1143 and replaces it with the native GTK 4.18+ `limit-events` property. This provides a cleaner, long-term solution to prevent accidental window behaviour (moving, resizing) when interacting with buttons in the headerbar.

**Testing performed**
- Reverted the previous PR and reproduced the event leaking.
- Applied `<property name="limit-events">true</property>`.
- Verified that clicking, double-clicking, and dragging on the headerbar buttons no longer propagates to the window.

https://github.com/user-attachments/assets/b09af605-c851-4db9-b489-45f22d4f6954

### DIFF
--- a/crates/rnote-ui/data/ui/dialogs/dialogs.ui
+++ b/crates/rnote-ui/data/ui/dialogs/dialogs.ui
@@ -71,11 +71,13 @@ Changes which are not saved will be permanently lost.</property>
             <property name="show-start-title-buttons">false</property>
             <child type="start">
               <object class="GtkButton" id="edit_selected_workspace_button_cancel">
+                <property name="limit-events">true</property>
                 <property name="label" translatable="yes">Cancel</property>
               </object>
             </child>
             <child type="end">
               <object class="GtkButton" id="edit_selected_workspace_button_apply">
+                <property name="limit-events">true</property>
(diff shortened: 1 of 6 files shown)


Doublonmousse:
Finally took the time to test with touch/pen input and this works as expected.

I see you also extended this a little bit to export/import dialogs as well.

alberto-is:
Yes, since I was already at it, I checked whether it was needed anywhere else as well

Doublonmousse (approved):

flxzt:
thanks for your contribution!

PR #1775, opened 2026-05-18
fix: use tangents for smooth stroke composition

Kneemund:
- Because of how stroke composition works so far, line segments with highly varying widths had noticeable "dents" in them.
- This was often hidden fairly well by the large amount of path elements, but it becomes much more noticeable with #1769.

|Before|After|
|---|---|
|grafik|grafik|
|grafik|grafik|

### Details
- So far, line segments are constructed as trapezoids with semicircular caps:
   - The trapezoids are constructed to be exactly perpendicular to the stroke segment. 
   - They are only continuous with the line caps if both ends have the same width.
- This PR changes the approach, and it can be interpreted in the following way:
   - A circle with the start width is placed at the start of the line segment.
   - That circle is continuously deformed until it reaches the end width at the end of the line segment.
   - This makes the stroke continuous even at different widths. It is achieved by calculating the direction that is tangential to both circles, and connecting the circles along that direction.

### Debug View Examples
grafik
grafik

(The bottom image shows the degenerate case where one circle is contained in the other. It falls back to the previous behavior.)

### DIFF
--- a/crates/rnote-compose/src/style/smooth/mod.rs
+++ b/crates/rnote-compose/src/style/smooth/mod.rs
@@ -397,6 +397,10 @@ impl Composer<SmoothOptions> for crate::Shape {
     }
 }
 
+/// Appends a circular arc between two points on a circle.
+///
+/// The arc is centered at `center` and connects `start` to `end`.
+/// Between the two possible sweeps, the arc whose midpoint direction best aligns with `direction` is selected.
 fn append_arc_between_points(
     path: &mut kurbo::BezPath,
     center: kurbo::Point,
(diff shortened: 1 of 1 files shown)


crates/rnote-compose/src/style/smooth/mod.rs
```diff
@@ -397,6 +397,113 @@ impl Composer<SmoothOptions> for crate::Shape {
     }
 }
 
+/// Appends a circular arc between two points on a circle.
+///
+/// The arc is centered at `center` and connects `start` to `end`.
+/// Between the two possible sweeps, the arc whose midpoint direction best aligns with `direction` is selected.
+fn append_arc_between_points(
```
Kneemund, line 404: I have a feeling that we could simplify the "direction selection" somehow...

Doublonmousse:
I wonder if we have a systematic pen up event with 0 pressure giving 0 width at the end here from gtk... Maybe we don't want to have this at the end of each stroke ?

Kneemund:
Maybe... But this is a more general issue, it's just that rapid pressure changes most likely occur at the end of strokes.

PR #1776, opened 2026-05-21
refactor: replace use of `nalgebra` with `glam`

flxzt:


crates/rnote-compose/src/builders/coordsystem3dbuilder.rs
```diff
@@ -106,17 +106,21 @@ impl CoordSystem3DBuilder {
...
         };
 
         let backward_axis = Line {
             start: center,
-            end: center + na::vector![tip_x_offset, -tip_x_offset],
+            end: center
+                + Vector2 {
+                    x: tip_x_offset,
+                    y: -tip_x_offset,
+                },
```
Kneemund: Maybe switch to `Vector2::new` for consistency.

crates/rnote-compose/src/ext.rs
```diff
@@ -469,26 +434,42 @@ where
...
+    /// Append rotation with regards to a supplied center
+    fn append_rotation_wrt_center(self, rotation: f64, center: Vector2) -> Self;
+}
+
+impl DPose2Ext for DPose2 {
+    fn append_rotation_wrt_center(self, angle: f64, center: Vector2) -> Self {
+        DPose2::from_rotation(DRot2::from_angle(angle))
+            .prepend_translation(-center)
+            .append_translation(center)
+            * self
```
Kneemund: I think we can replace `DPose2::from_rotation(DRot2::from_angle(angle))` with `DPose2::rotation(angle)`.

In the spirit of this convention, we could add a new function `rotation_wrt_center` that constructs a new `DPose2` using
`DPose2::IDENTITY.append_rotation_wrt_center(angle, center)`. Then we could replace the occurrences of `DPose2::IDENTITY.append_rotation_wrt_center` with that.

crates/rnote-compose/src/serialize.rs
```diff
@@ -17,64 +17,64 @@ pub mod f64_dp3 {
...
     use serde::{Deserializer, Serializer};
 
-    /// Serialize a [`na::Vector2<f64>`] rounded to 3 decimal places
-    pub fn serialize<S: Serializer>(v: &na::Vector2<f64>, s: S) -> Result<S::Ok, S::Error> {
+    /// Serialize a [`Vector2`] rounded to 3 decimal places
+    pub fn serialize<S: Serializer>(v: &Vector2, s: S) -> Result<S::Ok, S::Error> {
         const D: f64 = (10_u32.pow(3)) as f64;
         let mut a = v * D;
-        a = na::vector![a[0].round(), a[1].round()];
+        a = Vector2::new(a.x.round(), a.y.round());
```
Kneemund: Minor nitpick, maybe rename `D` to `SCALE` and use `.round()` directly on the vector.

I.e. `(v * SCALE).round() / SCALE` like below.

crates/rnote-engine/src/engine/rendering.rs
```diff
@@ -61,6 +62,8 @@ impl Engine {
 
             if let Some(image) = &self.origin_indicator_image {
                 // Only create the texture once, it is expensive
+
+                use p2d::math::Vector2;
```
Kneemund: This seems accidental (already imported in outer scope).

crates/rnote-engine/src/engine/visual_debug.rs
```diff
@@ -140,15 +142,17 @@ pub(crate) fn draw_statistics_to_gtk_snapshot(
 
     // A statistics overlay
     {
+        use p2d::math::Vector2;
+
```
Kneemund: This seems accidental (already imported in outer scope).

crates/rnote-engine/src/strokes/bitmapimage.rs
```diff
@@ -117,8 +117,8 @@ impl BitmapImage {
             ),
         };
 
-        let mut transform = Transform::default();
-        transform.append_scale_mut(na::Vector2::new(resize_ratio, resize_ratio));
+        let mut transform = Transform::IDENTITY;
+        transform.append_scale_mut(Vector2::new(resize_ratio, resize_ratio));
```
Kneemund: Minor nitpick; we could use `Vector2::splat` here too.

crates/rnote-ui/src/canvas/canvaslayout.rs
```diff
@@ -10,6 +10,7 @@ use std::cell::Cell;
 
 mod imp {
     use super::*;
+    use p2d::math::Vector2;
```
Kneemund: Should probably be moved to the outer scope imports.

crates/rnote-ui/src/canvas/input.rs
```diff
@@ -341,16 +342,21 @@ fn retrieve_pointer_elements(
...
             .map(|p| {
-                (canvas.engine_ref().camera.transform().inverse()
-                    * na::point![p.x() as f64, p.y() as f64])
-                .coords
+                canvas
+                    .engine_ref()
+                    .camera
+                    .transform()
+                    .inverse()
+                    .transform_point2(Vector2::new(p.x() as f64, p.y() as f64))
```
Kneemund: Can we use `p.to_p2d_vec()` here?

Kneemund (changes requested):
Cool, this makes some parts very clean. It's really nice that there is no distinction between vectors and points anymore, I always thought that was really annoying.

The matrix coefficients being row-major in `nalgebra` (while everything else is column-major) seems to be accounted for everywhere.

Kneemund:
As you've probably already read in the CI logs, this seems to break file compatibility at the moment.

- New file in old version: `Err: invalid length 6, expected a matrix array`
- Old file in new version: `Error: invalid length 9, expected fewer elements in array`

crates/rnote-ui/src/appwindow/actions.rs
```diff
@@ -1095,9 +1097,12 @@ impl RnAppWindow {
 
                 let last_contextmenu_pos = wrapper.last_contextmenu_pos().map(|vec2| {
                     let p = graphene::Point::new(vec2.x as f32, vec2.y as f32);
```
Kneemund: Maybe use `graphene::Point::from_p2d_vec(vec2)` here.

crates/rnote-ui/src/appwindow/actions.rs
```diff
@@ -1095,9 +1097,12 @@ impl RnAppWindow {
...
-                    (canvas.engine_ref().camera.transform().inverse()
-                        * na::point![p.x() as f64, p.y() as f64])
-                    .coords
+                    canvas
+                        .engine_ref()
+                        .camera
+                        .transform()
+                        .inverse()
+                        .transform_point2(Vector2::new(p.x() as f64, p.y() as f64))
```
Kneemund: Maybe use `p.to_p2d_vec()` here.

Kneemund (changes requested):

crates/rnote-compose/src/ext.rs
flxzt: so are you suggesting to introduce a `DPose2Ext` trait supplying this constructor for the `DPose2` type?

crates/rnote-compose/src/builders/coordsystem3dbuilder.rs
flxzt: addressed

crates/rnote-compose/src/serialize.rs
flxzt: addressed

crates/rnote-engine/src/engine/rendering.rs
flxzt: addressed

crates/rnote-engine/src/engine/visual_debug.rs
flxzt: addressed

crates/rnote-engine/src/strokes/bitmapimage.rs
flxzt: addressed

crates/rnote-ui/src/canvas/canvaslayout.rs
flxzt: addressed

crates/rnote-ui/src/canvas/input.rs
flxzt: addressed

crates/rnote-ui/src/appwindow/actions.rs
flxzt: addressed

crates/rnote-ui/src/appwindow/actions.rs
flxzt: addressed

flxzt:
thanks, I would not have expected to get such a detailed review.
I addressed the points and have added a working upgrade path for older files.
I also removed the `Transform` type completely, it's use is replaced by `DAffine2`.

crates/rnote-compose/src/ext.rs
Kneemund: Yes, we already added and use a `DPose2Ext` trait for this current function. 

`DPose2` seems to have a convention where there are constructors and append (/prepend) functions for all operations. We could also do that here by not only adding `append_rotation_wrt_center` but also `rotation_wrt_center` as a constructor.

flxzt: ah, this refactor was quite "mechanic" so I did not internalize what I even added.
Then for consistency I would replace `append_rotation_wrt_center()` with `from_rotation_wrt_center()` entirely
since `append_rotation_wrt_center` is always used in combination with the identity matrix anyway.

Kneemund (approved):

Issue #1777, opened 2026-05-21
Incorrect conversion from Cm or Mm less than 1.00 to Px


ParticularlyPotent:
**Describe the bug**  
In the settings panel, clicking a spin button's unit selection menu to convert (for instance) pattern width from Cm or Mm to Px gets converted incorrectly only when the value in Cm or Mm is less than 1.00. The Px number always gets converted to the same unexpected value (for example, 0.85 Cm and 0.30 Cm both convert to 38 Px for me).

**To Reproduce**  
1. Open up the settings panel
2. Set pattern width to be measured in Cm
3. Set Cm to a value less than 1.00, such as 0.85.
4. Select Px as the unit.
5. See that the value gets incorrectly converted.

**Expected behavior**  
When switching between units, the value should never change. Assuming my DPI, 0.85 Cm (which is 32 px) should convert back to 32 px, but it always becomes 38 px.

**Screenshots**  
Video showcase:
https://github.com/user-attachments/assets/24ffec00-07a7-4b94-85ad-1d5f36277301

**Desktop (please complete the following information):**  
 - OS: Arch Linux
 - App Version: Rnote v0.14.2
 - Installation Source: Flatpak
 - Desktop Environment: KDE Plasma 6.6.5
 - Display Server: Wayland
 - Input Source: Mouse and Keyboard

PR #1778, opened 2026-05-21
Fix spinner lower bound

ParticularlyPotent:
To fix issue #1777

[Cm to Px conversion fix.webm](https://github.com/user-attachments/assets/adc797d0-ae08-41eb-b5d0-15ac0e30920c)

### DIFF
--- a/crates/rnote-ui/src/unitentry.rs
+++ b/crates/rnote-ui/src/unitentry.rs
@@ -140,7 +140,7 @@ mod imp {
         fn set_property(&self, _id: usize, value: &glib::Value, pspec: &glib::ParamSpec) {
             let obj = self.obj();
             // Storing self.value before calling configure_spinner() to avoid issue #1777
-            let saved_value = self.value.get();
+            let copied_value = self.value.get();
 
             match pspec.name() {
                 "value" => {
@@ -157,7 +157,7 @@ mod imp {
                     if unit != self.unit.get() {
                         self.configure_spinner(unit, self.dpi.get());
                         obj.set_value(MeasureUnit::convert_measurement(
-                            saved_value,
+                            copied_value,
                             self.unit.get(),
                             self.dpi.get(),
                             unit,
@@ -171,7 +171,7 @@ mod imp {
                     if dpi != self.dpi.get() {
                         self.configure_spinner(self.unit.get(), dpi);
                         obj.set_value(MeasureUnit::convert_measurement(
-                            saved_value,
+                            copied_value,
                             self.unit.get(),
                             self.dpi.get(),
                             self.unit.get(),
--- a/crates/rnote-ui/src/unitentry.rs
+++ b/crates/rnote-ui/src/unitentry.rs
@@ -187,6 +187,8 @@ mod imp {
 
     impl RnUnitEntry {
         const MIN_VAL_IN_PX: f64 = 1.0;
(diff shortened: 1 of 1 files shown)


Kneemund:
Hi, thanks for reporting the issue and creating this PR!

`MeasureUnit::convert_measurement` already correctly converts the minimal bound for `PX` into other units. Your changes only make the bounds smaller for `MM` and `CM`, that's why the bug seems to disappear. 

The root cause of this bug are the following lines though:
https://github.com/flxzt/rnote/blob/99fd77f6533fd594d6195d69d96ddabed3c70a09/crates/rnote-ui/src/unitentry.rs#L156-L164

- `configure_spinner` updates the widget and applies the new minimum and maximum values, which clamps the `value` to the new minimum.
- `set_value` then converts the value to the new unit, but uses the clamped value of the spinner instead of the original one.

So a proper fix would be to either calculate the new value before calling `configure_spinner`, or storing the original old value in a variable before calling `configure_spinner`. The same issue occurs when setting the DPI, that can be fixed in the same way.

Do you want to make these changes yourself? Or do you want me to take over?

ParticularlyPotent:
Thank you for the response. You're right that it's better to instead use the pre-clamped value and leave configure_spinner alone. Let me post my change in a second...
edit: I am trying to figure out git. I am a beginner. Sorry

ParticularlyPotent:
Sorry, I had a bit of a mix-up with another account. Is this the kind of change you were looking for?

Kneemund:
Yes. Could you add the same fix to the DPI section too (directly below)? In theory, the same bug could occur there.

ParticularlyPotent:
Great catch! Although I do wonder if there should be a comment placed here briefly saying why we are making copies of these floating point numbers before we run configure_spinner. I am not extremely familiar with Rnote's codebase, and I don't have nearly as much Rust experience as y'all, but to me it seems unintuitive that running configure_spinner(), which is meant to set the spinner's properties will clamp values like these and ruin your conversions. Do you think this warrants a comment?

Doublonmousse:
Well, if you feel this warrant comments, feel free to add them (and refer to the github issue number).

The crux of the issue is keeping the UI state (displayed in the UI in the gtk4 element where the spinner bounds apply) and the internal state (the value/unit/dpi properties) coherent when updating. So there's some precautions to take here to update things in a coherent manner (all the more so when updating a property updates more than one like that is the case for unit/dpi).

The reason the update on the value occurs is probably 

https://github.com/flxzt/rnote/blob/99fd77f6533fd594d6195d69d96ddabed3c70a09/crates/rnote-ui/src/unitentry.rs#L80-L85

So setting the bounds on the unitentry will trigger a change of the `self.value` in the case where clamping moves the value to the clamped one

NB : there's other issues with this component, like #1773.

crates/rnote-ui/src/unitentry.rs
Kneemund, line 169: I actually meant this section with "DPI section". We need the same fix here. But I don't think we need to store the DPI (what your latest commit did).

Maybe it would be best to simply calculate the new value altogether before calling the configure function (instead of doing it inline in the function argument)?

Kneemund (changes requested):

ParticularlyPotent:
I hope that's what was intended. Anything else should be done?

PR #1792, opened 2026-07-05
fix: cargo build script issues

flxzt:
Fixes two issues related to the cargo build script:
- `cargo_env` was not passed correctly to it
- replace deprecated calls to `os.system()` with `subprocess.run()`

### DIFF
--- a/build-aux/cargo_build.py
+++ b/build-aux/cargo_build.py
@@ -1,7 +1,8 @@
 #!/usr/bin/env python3
 
+import subprocess
+from subprocess import CalledProcessError
 import sys
-import os
 
 project_build_root = sys.argv[1]
 project_src_root = sys.argv[2]
@@ -11,7 +12,8 @@ cargo_options = sys.argv[5]
 bin_output = sys.argv[6]
 output_file = sys.argv[7]
 
-print(f"""
+print(
+    f"""
 ### executing cargo build script with arguments: ###
     project_build_root: {project_build_root}
     project_src_root: {project_src_root}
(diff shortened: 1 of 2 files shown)


meson.build
```diff
@@ -206,7 +206,7 @@ if build_ui == true
             cargo_build_script,
             meson.project_build_root(),
             meson.project_source_root(),
-            cargo_env,
+            ' '.join(cargo_env),
```
Doublonmousse, line 209: I'm not sure I understand what is changed with this (does the issue occurs with some specific path or config ?)

build-aux/cargo_build.py
```diff
@@ -20,21 +22,23 @@
...
 cargo_call = f"env {cargo_env} {cargo_cmd} build {cargo_options}"
 cp_call = f"cp {bin_output} {output_file}"
 
 print(cargo_call, file=sys.stderr)
-res = os.system(cargo_call)
-if res != 0:
-    print(f"cargo call failed, code {res}", file=sys.stderr)
-    sys.exit(1)
+try:
+    subprocess.run(cargo_call, shell=True, check=True, capture_output=True)
```
Doublonmousse: This seems to suppress output until cargo has finished its compilation so the terminal output can't be used for checking progress (and compile errors aren't shown either in that case, only that cargo returned with an 101 error code)

```suggestion
    subprocess.run(cargo_call, shell=True, check=True, stdout=sys.stdout, stderr=sys.stderr)
```

flxzt: Ah true I don't know why I opted for `capture_output`. Since forwarding to the script stdout, stderr seems to be the default I will just remove this argument.

meson.build
flxzt: This causes that the cargo environment list is passed in correctly by space-separating each item.
Before this only worked by accident. We only supplied a single value in the list so it was not an issue, but if multiple values are given the list would per-default be concaternated without any separator IIRC.

flxzt:
@Doublonmousse approved?

Doublonmousse:
Yes !

Doublonmousse (approved):

Issue #147, opened 2022-04-18
Implement drawing tools


CountOmega:
It would be nice if one could select different tools, e.g. a ruler to draw straight lines (like in Onenote). This would be helpful for users who need technical/mathematical drawings. Also, a protractor could be of use for angles etc.

Kneemund:
I'm currently looking for ideas on how these tools could be controlled. The most intuitive method is replicating OneNote, i.e. using touch input to drag and rotate them.

However, Rnote is supposed to be usable without having touch as a secondary input method, e.g. when using mouse & keyboard or enabling "touch drawing". For those cases we need alternative, intuitive and (if possible) consistent controls.

Separate UI elements (like when scaling/rotating selections) come to mind, but I have no idea where you could position them so that they're not tedious to use.

(The ruler in OneNote can only be used when the device supports touch input - not sure what other programs do.)

![grafik](https://github.com/user-attachments/assets/7f36774f-6fda-4c78-8f31-a542d5156133)

Atreyu-94:
@Kneemund You can add a button for translation and one for rotation like Openboard does that has a lot of math tools.
Some examples: https://youtu.be/G8WKxCiXahY?si=CTU-eeK0x-1WZkhu&t=21

Kneemund:
Thanks. Positioning the buttons seems tricky if you have an "infinite" ruler that can go partially out of view, but maybe it's the only solution that makes it accessible for keyboard and mouse.

I'm also not sure if the ruler should be infinite in the first place. Are there any advantages to a fixed length ruler like the one shown in the video?

TazerXI:
I think it would be included here, but if not please point me in the right direction, but a way to measure distances within the app would be nice. Either doing it similar to measuring distances on Google Maps, where you select 2 points and it gives the distance between them, or using this ruler where instead of/along side drawing the line, it also tells you a precise length of the measurement.

der-wasguckstdu-dreier:
How about using the scroll wheel to rotate the ruler?
 
I'd also really like to be able to set dimensions. I also miss a feature to scale my elements on screen to the real world like the “ruler setting” in xournal.

MisterAnonym1:
You could also look in the ruler implementation of Xournal++. It is only touchscreen compatible but I quite enjoy how it works and it isn't infinite.

gregorianrants:
I think a ruler would be great but i want to also point out an intermediate case between touch and mouse

[...]

PR #1796, opened 2026-07-06
feat: ruler tool

andrinwinzap:
# Add a OneNote-style ruler tool

## Summary

This adds an on-canvas **ruler** (straight-edge) to the brush tool, the first of the helper tools requested in #147. When enabled, a translucent ruler spans the viewport; brush strokes drawn along its edge are snapped to a perfectly straight line, and the ruler can be freely repositioned and rotated with pen, touch, or mouse.

Addresses #147 (ruler portion; the protractor is left for a follow-up).

## Motivation

Rnote had no way to draw a clean straight line freehand. A ruler is the most-requested helper tool in #147 and mirrors a familiar workflow from OneNote, making technical and mathematical sketches practical directly on the canvas.

## What's included

- **Draw against the edge** – starting a brush stroke within the snap zone of a ruler edge locks the stroke to that straight edge for its whole duration. The snap decision is made once at stroke start, so a stroke either follows the ruler or doesn't — it never jumps mid-stroke.
- **Reposition** – dragging the ruler body (pen / one finger / mouse) moves it; the ruler lives in window-relative coordinates, so panning or zooming the canvas leaves it fixed on screen.
- **Rotate** – a two-finger gesture rotates around the projected finger centroid; mouse-wheel / trackpad scrolling over the body also rotates it, with speed-adaptive scaling and a short session so the pivot stays put across successive scrolls.
- **Angle snapping** – optional snap to 0° / ±45° / ±90° with hysteresis (easy to enter a snap, a small nudge to leave it).
- **Tick marks & angle dial** – three-tier tick marks along the edges and an optional dial showing the current angle at the center of rotation.
- **Configuration popover** (in the brush page) exposes: snap distance, angle-snap toggle, show-dial toggle, body width, tick spacing, body opacity, and scroll rotation step. Persisted settings survive across sessions; live ruler position/angle/visibility are session-only.

## Implementation notes

- New `RulerConfig` (`crates/rnote-engine/src/pens/pensconfig/rulerconfig.rs`) holds both the persisted preferences and the in-session runtime state, plus the coordinate conversions and snap math.
- Rendering lives in `crates/rnote-engine/src/pens/ruler.rs`.
- The brush pen (`crates/rnote-engine/src/pens/brush.rs`) gains a `DraggingRuler` state and the per-stroke snap logic.
- UI wiring (toggle, popover, symbolic icon) is in `rnote-ui` (`penssidebar/brushpage.rs` + `.ui`), and the gesture handling (drag / two-finger rotate / scroll-rotate) is in `crates/rnote-ui/src/canvaswrapper.rs`.

## Checks

- `cargo fmt --check` – clean
- `ui-cargo-clippy` / `cli-cargo-clippy` – no warnings
- `cargo-test` (nextest) – all passing

## Screencast

[Screencast_20260706_214039.webm](https://github.com/user-attachments/assets/bac1f0c8-3c2a-4afa-bed2-410ac0ede758)

## AI disclosure

This contribution was developed with the assistance of LLM/GenAI tools.


flxzt:
Commented by @Kneemund that this seems to have the same limitations and has no solution for the issues discussed in https://github.com/flxzt/rnote/issues/147 .
Since this is LLM generated I am closing, the open questions in the issue need to be resolved first.

andrinwinzap:
Thanks for the feedback. And thanks for all the work you are doing with Rnote; it is truly awesome.
I am not quite clear on how to proceed. As far as I can tell, my PR solves the cases for touch input and mouse input. The only potential issue I see there is controlling the ruler on pen tablets/displays with stylus input but no touch and often no scroll wheel as described in @gregorianrants's comment. I'm happy to adjust my PR to handle that if that's the issue.
The other issue I see is the question of finite vs. infinite length and of measurements. The decision to make it infinite was simply a personal preference because I thought it was intuitive, and no definite conclusion emerged from the discussion. But I can also adjust that if needed. Same goes for measurements.
Could you or @Kneemund confirm which of these are the issues you mentioned and how the questions can be resolved?

PR #1801, opened 2026-07-14
Read-only CI

flxzt:
I am not comfortable with a CI that is able to push automatically to the repo anymore.
This removes the pipeline that was doing that and replaces it by documentation documenting what manual steps need to be done for maintenance every once in a while.


Kneemund:
I didn't have time to review this earlier, but both CI PRs look good to me. 

Instead of fully removing the workflow, maybe we can make it read-only instead? I.e. simply fail a job if the translations are outdated?

flxzt:
That is a good suggestion. Let me see if I can implement that.

Issue #1592, opened 2025-11-22
Feature: Auto-generate filenames with timestamps


rafl0:
I moved from Xournal++ to Rnote and it’s a solid upgrade, but I miss the automatic file naming. Xournal++ defaults the filename to a timestamp (e.g., 2025-10-03-Note-21-00) when you hit save.

This is better for organizing files chronologically and allows for a true "one-click save" workflow without having to type a title every time when you make a lot of different notes. It would be great to see a toggle in the settings to enable this behavior in Rnote.

Doublonmousse:
I feel like having the name include the timestamp when no name is provided is a reasonable default to keep it as the only behavior

ZiuBisk:
Hi! I'd like to work on this issue if it's still available.
I'll open a draft PR shortly.

Doublonmousse:
@ZiuBisk Yes, this issue is still available

PR #1802, opened 2026-07-15
fix: ensure new documents / drafts are also autosaved

### LINKED ISSUE #545: Auto Save shall also save drafts or add an option for it please

gi1242:
When opening a new file, rnote won't autosave it until it's saved first. The default name is "New Document", and on tablets it's often a burden to choose a name and save it. So I often just work without saving the file initially. If there's a crash, then everything is lost.

I just implemented a fix:

1. Offer a time/date based name for the file initially. (A user can jst save this in their workspace without having to think twice)
2. Before the file has been officially saved, autosave the draft in ~/.cache/rnote/autosaves
3. If the file is saved, then cleanup any autosaves in ~/.cache/rnote/autosaves
4. On startup, if there are files in ~/.cache/rnote/autosaves, offer to load, delete or ignore them. (If the file is loaded, then it will be treated as a "new file" that is again autosaved as a draft in ~/.cache/rnote/autosaves. The old autosave will have to be deleted by the user (e.g. by choosing delete on the next run, or manually)

Fixes #545

### DIFF
--- a/crates/rnote-ui/src/app/mod.rs
+++ b/crates/rnote-ui/src/app/mod.rs
@@ -14,7 +14,8 @@ use crate::{
     workspacebrowser::RnWorkspacesBar, workspacebrowser::workspacesbar::RnWorkspaceRow,
 };
 use adw::subclass::prelude::AdwApplicationImpl;
-use gtk4::{WindowGroup, gio, glib, glib::clone, prelude::*, subclass::prelude::*};
+use adw::prelude::*;
+use gtk4::{WindowGroup, gio, glib, glib::clone, subclass::prelude::*};
 
 mod imp {
     use super::*;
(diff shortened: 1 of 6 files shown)


crates/rnote-ui/src/app/mod.rs
```diff
@@ -142,6 +143,75 @@ mod imp {
 
             appwindow.present();
 
+            // --- AUTOSAVE RECOVERY CHECK ---
+            let input_file_clone = input_file.clone();
+            glib::spawn_future_local(clone!(#[weak] appwindow, async move {
+                let cache_dir = glib::user_cache_dir();
```
flxzt, line 149: Can you add a function wrapping this invocation to `env.rs`? I want to have all functionality with regards to specific system and user directories collected there.

crates/rnote-ui/src/canvas/mod.rs
```diff
@@ -727,7 +731,7 @@ impl Default for RnCanvas {
 }
 
 pub(crate) static OUTPUT_FILE_NEW_TITLE: once_cell::sync::Lazy<String> =
-    once_cell::sync::Lazy::new(|| gettext("New Document"));
+    once_cell::sync::Lazy::new(|| String::from("%Y%m%d-%H%M%S"));
```
Doublonmousse, line 734: Does this fixes a single date for each rnote session ? I'm not convinced this is run once by canvas/new file (in which case it's better to remove the `static` and move the date logic where it's needed : here and in `default_file_title_for_export`)

Doublonmousse:
Fixes #1592 as well

PR #1809, opened 2026-07-21
wrap `shapeconfig_popover` in a paged popover

### LINKED ISSUE #1504: Shape Tool settings not opening in Windows after building from MSYS2

Doublonmousse:
Follow up on https://github.com/flxzt/rnote/pull/1524 with option 2 : paged popover

image
image

Fixes https://github.com/flxzt/rnote/issues/1504.

Sister PR to #1808 (where option 1 : use a scrolled window was used instead)

### DIFF
--- a/crates/rnote-ui/data/ui/penssidebar/shaperpage.ui
+++ b/crates/rnote-ui/data/ui/penssidebar/shaperpage.ui
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8" ?>
 <interface>
   <template class="RnShaperPage" parent="GtkWidget">
     <property name="layout-manager">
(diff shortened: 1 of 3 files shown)


flxzt:
I personally prefer this over the scrolled solution.
I am wondering if a more fitting UI pattern to use would be the view switcher: https://developer.gnome.org/hig/patterns/nav/view-switchers.html

Doublonmousse:
Yeah, you're right, let's use the view switcher (instead of my custom thing, we'll also get more accessibility/tab support this way as well).

I'll probably not take a look at that very quickly (probably not next week) so if you want to take it on, feel free to do so

anesthetice:
I prefer the other solution (using a scrollable window), as I feel that it is less finicky than having to switch between popover pages. 

For instance in tablet mode, it's easier to simply swipe your finger over the entire popover to scroll down rather than having to press a specific button. Moreover, currently the scroll approach doesn't even require a swipe gesture for nearly all users as the popover only becomes scrollable once its height is larger than the number of pixels available.

Also, this could probably be at least partially fixed by having a set desired width, but currently on my machine at least, when switching back to the first page, the size of the popover "pops back" after a short delay... 

https://github.com/user-attachments/assets/5422e582-dc1f-42a2-91dc-84158f52c80f

Doublonmousse:
Switched to a view switcher where the two pages are kept in sync as far as vertical space is concerned.

For now the name of the pages and icons are uninspired (maybe `Highlighting and constraints` for the second, that does come with a larger width however, maybe a better icon/name combination would be prefered). The resulting popover is also less tall because of reporting the title into the view switcher

image

In debug it does weird things as far as theming goes (had to change the syntax because of https://github.com/flxzt/rnote/pull/1524#issuecomment-3693227791 that I did encounter before)

PR #1811, opened 2026-07-23
feat: support building for Windows on ARM (aarch64) via MSYS2 CLANGARM64

sci42:
Makes the Windows build work in the MSYS2 CLANGARM64 environment, so Rnote runs
natively on Windows on ARM devices, with GPU rendering.

### Approach

Rather than adding aarch64 branches, the Windows recipes now take the
architecture-dependent parts from the MSYS2 environment itself:

- package prefix from `$MINGW_PACKAGE_PREFIX`
- install prefix from `$MSYSTEM_PREFIX` (via `cygpath -m`), replacing the
  hardcoded `C:/msys64/mingw64`
- Inno-Setup `ArchitecturesAllowed` from `host_machine.cpu_family()`
- openssl DLLs by glob, since the filename carries the architecture
  (`libcrypto-3-x64.dll` vs `libcrypto-3-arm64.dll`)
Separately, `inno_build.py` no longer uses `ldd` to find the DLLs to package.
`ldd` resolves imports by loading the binary, and ANGLE's `libGLESv2.dll` blocks
while loading, which made the installer build hang indefinitely. It now reads the
PE import table with `objdump -p` and walks it transitively; the resulting set is
verified to be closed. That commit is independent of ARM and can be taken on its
own.

The existing MINGW64 workarounds are unchanged and now sit behind an explicit
`MSYSTEM == MINGW64` check, so the x86_64 path keeps its exact previous behaviour.

CLANGARM64 needs the gtk/libadwaita/gettext pinning for the same reasons MINGW64
does, but not the `libpthread.dll.a` rename — linking with `lld` works as is.

### New

- `just setup-win-dev` — devel build installing into `$MSYSTEM_PREFIX` instead of
  `/usr`, so the app can be run locally without building the installer.
- `misc/building/rnote-windows-arm64-build.md`

### Tested

On a Surface Pro 11 (Snapdragon X Elite `X1E80100`, Adreno X1-85, Windows 11):

- `rnote-cli test misc/file-tests/*.rnote` passes for all six fixtures
- `GSK_DEBUG=renderer` reports `Using renderer 'GskGLRenderer'` on ANGLE, drawing
  is smooth and the rendering artifacts seen under the cairo fallback are gone
- installer builds, installs, and the installed app is stable across ten minutes
  of ordinary use (drawing, settings, zooming) — the same usage crashed the app
  after roughly two minutes before the `GDK_DISABLE=wgl` change
- opening a document containing a text stroke: 3 of 3 crashes before the
  `PANGOCAIRO_BACKEND=fc` change, 5 of 5 clean after, with the crash reproducing
  identically on the meson-installed build and on the installed app from the
  installer. Forcing `PANGOCAIRO_BACKEND=win32` back on brings the crash back,
  which also confirms the environment still overrides the default

- the **MINGW64/x86_64 build path**: a `release-windows.yml` run on
  `windows-2025` completed every step — `prerequisites-win` under MINGW64,
  `setup-win-installer`, `build`, `build-installer` — and produced an installer
  ([run](https://github.com/sci42/rnote-win-arm64/actions/runs/33399608913)).
  That exercises the recipes this PR rewrites (`$MINGW_PACKAGE_PREFIX` /
  `$MSYSTEM_PREFIX`) and the new `objdump`-based DLL collection on x86_64.

**Not** tested by me:

- **Running** the resulting x86_64 installer. The build path is covered above,
  but I have no x86_64 Windows machine to install it on.
- Whether the DirectWrite font path is equally broken on x86_64 Windows. No
  hardware available, which is why both env defaults are scoped to aarch64
  rather than to Windows in general. If it turns out to be broken there too,
  widening the scope is a one-line change.
- Note also that the `PANGOCAIRO_BACKEND` default lives in `rnote-ui`, so
  `rnote-cli` does not get it. CLI export worked with either backend in my
  tests, but I did not investigate why it is unaffected.

> Disclosure per CONTRIBUTING.md: this was implemented with LLM/GenAI tools.
> No `Co-authored-by:`/`Assisted-by:` trailers were added to the
> commits, per the same document.


sci42:
Small status update, in case it helps triage this:

CI on the current head (`aa25caad`) passes in my fork — all steps green, including tests, file-compatibility and lints:
https://github.com/sci42/rnote-win-arm64/actions/runs/33182838241

The red check on this PR is a stale run from July 29 (`30455971368`) that failed at workflow startup and never produced any jobs, so it says nothing about the current state. Whenever a maintainer gets around to approving the workflow run for this PR, the checks here should reflect that.

No rush from my side — happy to rebase, split this up, or drop it entirely if the added Windows-on-ARM surface isn't something you want to maintain right now.

PR #1814, opened 2026-07-28
feat(ci): notify update translations

flxzt:


flxzt:
This seems to work. Basically only weblate translations updates should trigger the "Check Translations" workflow and it correctly reports when automatic translations updates are possible with instructions how to do it.

flxzt:
note: the `drop:` commits should be dropped before this is merged.

Kneemund:
Hm, the workflow does fail, which is technically correct, but the reason seems to be a repository permission issue: 

```
Run git diff-index --quiet HEAD -- || \
fatal: detected dubious ownership in repository at '/__w/rnote/rnote'
To add an exception for this directory, call:
	git config --global --add safe.directory /__w/rnote/rnote
```

No idea why that happens, I thought the checkout action would take care of this...

Kneemund:
Apparently it's due to running it in the fedora container. We can either run `git config --global --add safe.directory "$GITHUB_WORKSPACE"`, or switch to "native" Ubuntu.

flxzt:
Thanks for pointing it out, let's see if this works now.

.github/workflows/check_translations.yml
```diff
@@ -0,0 +1,37 @@
...
+
+      - name: Update translations
+        run: |
+          just update-translations
+
+      - name: Check File Changes
+        run: |
+          git diff-index --quiet HEAD -- || \
+            echo "::error title=Check Translations::Translations can be updated. Run 'just update-translations' and commit the changes." \
+            && exit 1;
```
Kneemund: I think we need to wrap the echo and exit in parentheses, because && and || have equal precedence for some reason (https://unix.stackexchange.com/questions/88850/precedence-of-the-shell-logical-operators).

Kneemund (changes requested):

.github/workflows/check_translations.yml
flxzt: changed!

Kneemund (approved):
Looks good to me, and should hopefully pass after removing the `drop:` commits :)

PR #1820, opened 2026-08-11
feat: make focus mode toggleable by a pen button

### LINKED ISSUE #1785: Make focus mode toggleable by pen.
My screen is oled so to avoid burning i enable focus mode and set a gtk condition to hide titlebar on maximize. However, Enabling and disabling focus mode is a hastle and i want it to be done by pen button that is it.

alberto-is:
Closes #1785

Adds a new `ShortcutAction` so any pen shortcut key (stylus buttons, drawing pad buttons, touch two-finger long press,keyboard Ctrl+Space, …) can be bound to toggle the app's focus mode from the existing pen shortcut settings, without changing any default shortcuts.

**Changes**

Engine (`rnote-engine`)

* Added a new `ShortcutAction` to toggle **Focus mode**.
* Added a flag to let the UI know when Focus mode needs to be changed.

UI (`rnote-ui`)

* Added **“Focus mode”** to the list of available pen shortcut actions, using the existing icon.
* Renamed a few internal components to reflect that the list now contains both pen actions and Focus mode.

**Testing**

- `cargo fmt --check`
- `cargo clippy --workspace --all-targets -- -D warnings` (no warnings)
- `cargo test --workspace`
- `meson compile ui-cargo-clippy -C _mesonbuild`
- `meson compile cli-cargo-clippy -C _mesonbuild`
- `meson test -C _mesonbuild` (3/3 OK)
- Manual: bound a pen shortcut to "Focus mode" and verified that focus mode
  toggles and the shortcut remains configured after reopening the settings.

https://github.com/user-attachments/assets/ee4f0662-3495-4d5b-974d-37c9b50d7b3c

**Disclosure**
This implementation was created with the assistance of an LLM.

### DIFF
--- a/crates/rnote-ui/src/settingspanel/penshortcutmodels.rs
+++ b/crates/rnote-ui/src/settingspanel/penshortcutmodels.rs
@@ -14,9 +14,9 @@ use std::str::FromStr;
 pub(crate) const FOCUS_MODE_ENTRY: &str = "focus_mode";
 
 #[derive(Debug, Clone)]
-pub(crate) struct ChangePenStyleListModel(StringList);
+pub(crate) struct PenShortcutActionListModel(StringList);
 
-impl Default for ChangePenStyleListModel {
+impl Default for PenShortcutActionListModel {
     fn default() -> Self {
         Self(StringList::new(&[
             &PenStyle::Brush.to_string(),
@@ -30,7 +30,7 @@ impl Default for ChangePenStyleListModel {
     }
 }
 
-impl Deref for ChangePenStyleListModel {
+impl Deref for PenShortcutActionListModel {
     type Target = StringList;
 
     fn deref(&self) -> &Self::Target {
@@ -38,16 +38,16 @@ impl Deref for ChangePenStyleListModel {
     }
 }
 
-impl DerefMut for ChangePenStyleListModel {
+impl DerefMut for PenShortcutActionListModel {
     fn deref_mut(&mut self) -> &mut Self::Target {
         &mut self.0
     }
 }
 
 #[derive(Debug, Clone)]
-pub(crate) struct ChangePenStyleListFactory(SignalListItemFactory);
+pub(crate) struct PenShortcutActionListFactory(SignalListItemFactory);
 
-impl Default for ChangePenStyleListFactory {
+impl Default for PenShortcutActionListFactory {
     fn default() -> Self {
         let factory = SignalListItemFactory::new();
         factory.connect_setup(move |_factory, list_item| {
@@ -112,7 +112,7 @@ impl Default for ChangePenStyleListFactory {
     }
 }
 
-impl Deref for ChangePenStyleListFactory {
+impl Deref for PenShortcutActionListFactory {
     type Target = SignalListItemFactory;
 
     fn deref(&self) -> &Self::Target {
@@ -120,16 +120,16 @@ impl Deref for ChangePenStyleListFactory {
     }
 }
 
-impl DerefMut for ChangePenStyleListFactory {
+impl DerefMut for PenShortcutActionListFactory {
     fn deref_mut(&mut self) -> &mut Self::Target {
         &mut self.0
     }
 }
 
 #[derive(Debug, Clone)]
-pub(crate) struct ChangePenStyleIconFactory(SignalListItemFactory);
+pub(crate) struct PenShortcutActionIconFactory(SignalListItemFactory);
 
-impl Default for ChangePenStyleIconFactory {
+impl Default for PenShortcutActionIconFactory {
     fn default() -> Self {
         let factory = SignalListItemFactory::new();
         factory.connect_setup(move |_factory, list_item| {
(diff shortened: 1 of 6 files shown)


crates/rnote-ui/src/settingspanel/penshortcutrow.rs
```diff
@@ -85,6 +96,7 @@ mod imp {
                         ShortcutAction::ChangePenStyle { mode, .. } => {
                             *mode = penshortcutrow.shortcut_mode();
                         }
+                        _ => {}
```
muditchaudhari: These two `match`es on `ShortcutAction` are no longer exhaustive — this one and the `_ => ShortcutMode::Temporary` on line 80.

Before this PR the match here was exhaustive over the single variant, so adding a variant would have been a compile error pointing right at the places that need updating. With `_` a future `ShortcutAction` silently falls through and does nothing.

Spelling out `ShortcutAction::ToggleFocusMode => {}` instead would keep that safety net, at no cost here.

crates/rnote-ui/src/settingspanel/penshortcutrow.rs
```diff
@@ -150,8 +162,20 @@ impl RnPenShortcutRow {
...
+            .string(self.selected())
+            .is_some_and(|string| string == FOCUS_MODE_ENTRY)
+    }
+
+    fn focus_mode_index(&self) -> Option<u32> {
+        let index = self.imp().shortcut_actions_model.find(FOCUS_MODE_ENTRY);
+        (index != u32::MAX).then_some(index)
+    }
+
+    pub(crate) fn selected_pen_style(&self) -> Option<PenStyle> {
```
muditchaudhari, line 178: Small thing about the selection ↔ action mapping: it now uses two different mechanisms. This function and `set_pen_style` below (`self.set_selected(style.to_u32().unwrap())`) go through the *index*, while the focus mode entry is detected by *string* comparison.

The index coupling already existed, but it was harmless while the model contained nothing but pen styles. Now it is load-bearing: it silently assumes the pen styles occupy positions 0..=5 and that the focus entry is appended after them. Reordering the model, or inserting another non-pen entry later, would break the mapping with no compiler complaint — and `row.selected_pen_style().unwrap()` on line 83 would start panicking rather than failing gracefully.

Deriving both directions from the string (the way `selected_is_focus_mode` and `focus_mode_index` already do) would remove the assumption entirely. Possibly out of scope here, in which case a short comment on `set_pen_style` noting the ordering requirement would probably be enough.

crates/rnote-ui/src/settingspanel/penshortcutmodels.rs
```diff
@@ -63,28 +68,32 @@ impl Default for ChangePenStyleListFactory {
...
+            let item_string = list_item
+                .item()
+                .unwrap()
+                .downcast::<StringObject>()
+                .unwrap()
+                .string();
             let item_box = list_item.child().unwrap().downcast::<gtk4::Box>().unwrap();
 
+            let (label, icon_name) = if item_string == FOCUS_MODE_ENTRY {
+                (gettext("Focus mode"), String::from("focus-mode-symbolic"))
```
muditchaudhari: Minor: the entry string got a constant (`FOCUS_MODE_ENTRY`), but the icon name is written literally here and again on line 152 in `PenShortcutActionIconFactory`. A second constant next to the first would keep the two from drifting.

muditchaudhari (commented):
Nice feature — reusing the existing pen shortcut infrastructure rather than adding a separate binding mechanism seems like the right call, and the engine side follows the existing `WidgetFlags` pattern closely (default, `BitOrAssign`, handled in `handle_widget_flags`). Adding `#[serde(rename = "toggle_focus_mode")]` so persisted shortcut configs stay stable is a nice touch.

A few things I noticed while reading through it, left inline. Not a maintainer, just a fellow contributor — take or leave as you see fit.

I also checked the two things that looked most likely to cause trouble, and both are fine: `focus-mode-symbolic` is present in `crates/rnote-ui/data/icons/scalable/actions/` and registered in the gresource, and `StringList::find()` is gated behind gtk4-rs's `v4_18` feature, which matches the project's `gtk4 >= 4.18` requirement — so no minimum version bump is implied.

*Disclosure: I used an AI assistant while reviewing this.*

crates/rnote-ui/src/settingspanel/penshortcutrow.rs
alberto-is: Thx, I usually code in C++, so a few Rust details slip past me

crates/rnote-ui/src/settingspanel/penshortcutmodels.rs
alberto-is: Good catch, updated.

crates/rnote-ui/src/settingspanel/penshortcutrow.rs
alberto-is: Honestly, I did think about this when implementing the feature, which is why the new focus mode entry uses find instead of an index. I wanted to keep the changes minimal, so I left the existing pen style helpers as they were. But you're right, having two different approaches is a bit messy and could cause issues if new actions are added later. I've now updated them to use the same string based approach as focus mode: selected_pen_style parses the selected entry's string, and set_pen_style finds the index using find

alberto-is: (updated too, thanks for the review in general)

PR #1821, opened 2026-08-13
fix: use the native Command modifier for shortcuts on MacOS

### LINKED ISSUE #663: Keyboard shortcut for MacOS - replace Ctrl with Cmnd
Hi! As a MacOS user I would be very grateful for change all the Ctrl+... shortcut into Cmnd+.... shortcuts, since all the other apps use Cmnd instead of Ctrl. Maybe it also would be possible to add the feature for users to change the innate shortcuts?

muditchaudhari:
Fixes #663.

## Problem

On MacOS all application shortcuts are triggered with Control instead of the native
Command modifier, so Rnote does not behave like other Mac applications. This is also
inconsistent with what the app itself advertises: the MacOS menu bar shows the standard
Command accelerators (e.g. `⌘Z` for Undo), while the shortcut that actually works is
`Ctrl+Z`. As noted in the issue, this makes the shortcuts look simply broken until you
find out that Control is the modifier in use.

## Cause

Gtk4 has no platform-dependent primary accelerator modifier. `<Primary>` is merely an
alias for `<Ctrl>` (unlike in Gtk3, where it mapped to Command on MacOS), so
`setup_action_accels()` registered Control accelerators on every platform. On MacOS the
native modifier is Command, which Gdk reports as `<Meta>`.

## Fix

Accelerators are now declared with a `<Primary>` placeholder that is substituted with
`<Meta>` on MacOS and with `<Ctrl>` on all other platforms. The substitution happens in
one place in `crates/rnote-ui/src/utils.rs`; `cfg!(target_os = "macos")` is used to match
the existing platform handling in `env.rs` and `dialogs/export.rs`.

The keyboard shortcuts dialog hardcodes its accelerators in `shortcuts.ui` and does not
derive them from the registered ones, so it performs the same substitution when it is
built. Without this the dialog would keep displaying Control on MacOS while the app
reacts to Command.

To answer the question in the issue: yes, a plain remap works for all of the registered
accelerators. Everything they cover is standard on MacOS (`⌘S`, `⌘O`, `⌘T`, `⌘Z`, `⌘⇧Z`,
`⌘C`/`⌘X`/`⌘V`, `⌘B`/`⌘I`/`⌘U`, `⌘Q`, `⌘N`), including zoom (`⌘+`, `⌘-`, `⌘0`), which
follows the same convention as Safari and Preview. Accelerators without a modifier
(`F9`, `F10`, `F11`, the bare digits for the color setters) are untouched.

## What is intentionally left unchanged

Shortcuts that are not registered as application accelerators, but are handled in the
engine through `ModifierKey::KeyboardCtrl`, keep using Control on all platforms:

- zooming by scrolling (`Ctrl+ScrollUp` / `Ctrl+ScrollDown`)
- duplicating a selection (`Ctrl+D`)
- text editing in the typewriter (`Ctrl+A`, `Ctrl+←`/`→`, `Ctrl+Backspace`, `Ctrl+Home`/`End`)
- holding Control to toggle shaper constraints, and the selector / nudge modifiers

These three entries in `shortcuts.ui` therefore still declare `<Ctrl>` directly, and the
existing "Hold Ctrl ..." labels in the settings and shaper pages stay correct.

Mapping these to Command as well would mean remapping `ModifierKey::KeyboardCtrl` in
`canvas/input.rs`, which also changes the drawing-tool modifiers (shape constraints,
nudge distance, add-to-selection) and is not obviously the native behaviour in every case
— MacOS uses Option, not Command, for word-wise navigation, for instance. That felt like
a separate decision rather than something to fold into this change. Happy to follow up
with it if you would like it, either in this PR or a later one.

## Testing

Built and run on MacOS (Apple Silicon, MacOS 26, Gtk 4.22.4, libadwaita 1.9.3,
`-Dprofile=devel`), verified by hand in the running application:

Every registered accelerator was exercised and works with Command:

- `⌘C` / `⌘X` / `⌘V` copy, cut, paste
- `⌘Z` / `⌘⇧Z` undo, redo
- `⌘S` / `⌘⇧S` / `⌘O` save, save as, open
- `⌘T` / `⌘W` / `⌘N` / `⌘Q` new tab, close tab, new window, quit
- `⌘P` print, `⌘L` clear document, `⌘⇧I` import
- `⌘⇧A` / `⌘⇧R` add and remove page, `⌘⇧P` snap positions, `⌘⇧O` tabs overview
- `⌘+` / `⌘-` / `⌘0` zoom in, out, reset
- `⌘1` … `⌘6` pen styles
- `⌘B` / `⌘I` / `⌘U` text formatting
- `⌘?` keyboard shortcuts dialog, `⌘⇧V` visual debug (devel profile)

The corresponding Control shortcuts (`Ctrl+C`, `Ctrl+V`, `Ctrl+Z`, `Ctrl+S`, `Ctrl+O`,
`Ctrl+T`) no longer trigger these actions.

The engine-level shortcuts still work with Control as before: `Ctrl+ScrollUp` /
`Ctrl+ScrollDown` zooming, `Ctrl+D` duplicate, `Ctrl+A` select-all inside a text box,
`Ctrl+←` / `Ctrl+→` / `Ctrl+Backspace` word navigation and deletion, and holding Control
to toggle shaper constraints.

The keyboard shortcuts dialog displays Command for the registered accelerators and
Control for the engine-level ones.

No regressions found: typing inside text boxes, `F9` / `F10` / `F11`, the unmodified
digit keys for the color setters and `Ctrl+Space` all behave as before.

`cargo fmt --check` and `cargo clippy` are clean, and `cargo test` passes, including two
added unit tests covering the placeholder substitution on both platform branches.

The non-MacOS path was not exercised at runtime: the substitution produces exactly the
same `<Ctrl>` accelerator strings that were hardcoded before, so Linux and Windows
behaviour is unchanged by construction.

## Disclosure

An AI assistant was used while researching and preparing this change.

### DIFF
--- a/crates/rnote-compose/src/penevent.rs
+++ b/crates/rnote-compose/src/penevent.rs
@@ -85,6 +85,10 @@ pub enum KeyboardKey {
     CtrlLeft,
     /// Ctrl right.
     CtrlRight,
+    /// Meta left. This is the Command key on MacOS.
+    MetaLeft,
+    /// Meta right. This is the Command key on MacOS.
+    MetaRight,
     /// Home.
     Home,
     /// End.
--- a/crates/rnote-engine/src/pens/typewriter/penevents.rs
+++ b/crates/rnote-engine/src/pens/typewriter/penevents.rs
@@ -1057,6 +1057,8 @@ impl Typewriter {
                                 }
                                 KeyboardKey::CtrlLeft
                                 | KeyboardKey::CtrlRight
+                                | KeyboardKey::MetaLeft
+                                | KeyboardKey::MetaRight
                                 | KeyboardKey::ShiftLeft
                                 | KeyboardKey::ShiftRight => EventResult {
                                     handled: false,
--- a/crates/rnote-ui/data/ui/shortcuts.ui
+++ b/crates/rnote-ui/data/ui/shortcuts.ui
@@ -12,37 +12,37 @@
             <child>
               <object class="GtkShortcutsShortcut">
                 <property name="title" translatable="yes">Show Keyboard Shortcuts</property>
-                <property name="accelerator">&lt;ctrl&gt;question</property>
+                <property name="accelerator">&lt;Primary&gt;question</property>
               </object>
             </child>
             <child>
               <object class="GtkShortcutsShortcut">
                 <property name="title" translatable="yes">New Window</property>
-                <property name="accelerator">&lt;ctrl&gt;n</property>
+                <property name="accelerator">&lt;Primary&gt;n</property>
               </object>
             </child>
             <child>
               <object class="GtkShortcutsShortcut">
                 <property name="title" translatable="yes">New Tab</property>
-                <property name="accelerator">&lt;ctrl&gt;t</property>
+                <property name="accelerator">&lt;Primary&gt;t</property>
               </object>
             </child>
             <child>
               <object class="GtkShortcutsShortcut">
                 <property name="title" translatable="yes">Close the Active Tab</property>
-                <property name="accelerator">&lt;ctrl&gt;w</property>
+                <property name="accelerator">&lt;Primary&gt;w</property>
               </object>
             </child>
             <child>
               <object class="GtkShortcutsShortcut">
                 <property name="title" translatable="yes">Quit the Application</property>
-                <property name="accelerator">&lt;ctrl&gt;q</property>
+                <property name="accelerator">&lt;Primary&gt;q</property>
               </object>
             </child>
             <child>
               <object class="GtkShortcutsShortcut">
                 <property name="title" translatable="yes">Toggle Tabs Overview</property>
-                <property name="accelerator">&lt;ctrl&gt;&lt;shift&gt;o</property>
+                <property name="accelerator">&lt;Primary&gt;&lt;shift&gt;o</property>
               </object>
             </child>
             <child>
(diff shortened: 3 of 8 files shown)


muditchaudhari (commented):
fix: use the native Command modifier for shortcuts on MacOS

flxzt:
Did you actually test it as described or let the AI write this?

muditchaudhari:
Yes, I built it and tested it by hand on my Mac. And yes, I used an AI assistant for the implementation and to draft the description, which is why I disclosed it at the bottom.
The testing was mine: a local `-Dprofile=devel` build (Apple Silicon, Gtk 4.22.4, libadwaita 1.9.3), going through the accelerators by hand.
Two things I hit while testing that aren't visible from the diff, in case they're useful:

* Before this change, the MacOS menu bar already displays `⌘Z` for Undo while the shortcut that actually works is `Ctrl+Z` — which matches what @fyrkrans described in #663.
* After this change `⌘A` still does not select all inside a text box, because the typewriter handles select-all through `ModifierKey::KeyboardCtrl` in the engine rather than as an accelerator — same for `Ctrl+←`/`→`, `Ctrl+Backspace` and `Ctrl+D`. That is why I scoped those out and left them on Control.

Happy to record a screen capture of the shortcuts working if that would help.

muditchaudhari:
Recorded a capture of the shortcuts working on MacOS, with the on-screen keyboard visible so the modifiers are legible.



What the capture shows, in order:

- The **Edit** menu in the MacOS menu bar, which already advertises `⌘Z` / `⌘X` / `⌘C` / `⌘V` — the mismatch described in #663
- `⌘Z` / `⌘⇧Z` undo and redo
- `Ctrl+Z` and `Ctrl+C` no longer triggering those actions
- `⌘C` / `⌘V` inside a text stroke — the `Hello World Hello World` line is a copy/paste round trip
- `⌘A` *not* selecting all inside a text stroke, which is the intentional boundary described below

Testing also turned up a bug in my first commit, now fixed in `9eca8755`:

The typewriter cancels an active text selection for every key that is not explicitly allowed through, and only Ctrl and Shift were on that list. `KeyboardKey` had no variant for the Meta key, so `retrieve_keyboard_key()` mapped Command to `KeyboardKey::Unsupported` — pressing it collapsed the selection before the clipboard accelerator could run, which made `<Meta>c` and `<Meta>x` unusable inside a text stroke. Mapping `Meta_L` / `Meta_R` to new `KeyboardKey` variants and letting them pass through like the other modifier keys fixes it. That commit is 8 added lines across `rnote-compose`, `rnote-engine` and `rnote-ui`.

`ModifierKey` is deliberately left unchanged, so everything routed through `ModifierKey::KeyboardCtrl` still uses Control on MacOS — selecting all text, word-wise navigation and deletion, the shaper constraint modifier, and `Ctrl+D`. That is the `⌘A` behaviour visible at the end of the video. Happy to map Command there as well if you would prefer it, but it also changes the drawing-tool modifiers, so it seemed like your call rather than mine.


https://github.com/user-attachments/assets/d3a85401-2472-42b4-ae20-915610170762

muditchaudhari:
@flxzt can you please review it?

muditchaudhari:
@flxzt no rush on the review itself, I know the backlog is long and this is volunteer time.

But if this isn't something the project wants, I would rather know than leave it sitting open. Is it the AI-assisted implementation, the scope of the change, or just the queue? Happy to close it myself if it is the first, and happy to rework it if it is the second.

A one line answer either way is all I am after.

Issue #1440, opened 2025-05-07
Sometimes pointer jumps from stylus position  to  last position of mouse pointer


Aldijonas:
**Describe the bug**  
Eraser bound to stylus sometimes button briefly jumps to last position of mouse pointer, which causes unwanted erasing of strokes. Sometimes random lines from stylus position to last position mouse pointer appear. It is frustrating for random elements to disappear when erasing. 

**To Reproduce**  
1. Bind stylus button to eraser
2. Place mouse pointer above strokes
3. Hover stylus in different place than mouse pointer
4. Hold stylus button
5. Pointer will sometimes briefly jump to last position of mouse pointer, erasing strokes underneath mouse pointer and then those under stylus nib.

**Expected behavior**  
Pointer should not briefly jump to last position of mouse pointer and then come back to stylus pointer position. It should not erase unwanted items to create lines from stylus position to last position of mouse pointer.

**Console Output**  
RUST_LOG=rnote=debug RUST_BACKTRACE=1 rnote
   0.002741132s DEBUG rnote: .. tracing subscriber initialized.

(rnote:4722): Gtk-WARNING **: 14:50:00.632: Unknown key gtk-modules in /home/aldijonas/.config/gtk-4.0/settings.ini

(rnote:4722): Gtk-WARNING **: 14:50:00.700: Theme directory 32/emblems of theme Eleven-Dark has no size field

(rnote:4722): Gtk-WARNING **: 14:50:00.703: Theme directory 32@2x/emblems of theme Eleven-Dark has no size field

(rnote:4722): Gtk-WARNING **: 14:50:00.705: Theme directory 48/emblems of theme Eleven-Dark has no size field

(rnote:4722): Gtk-WARNING **: 14:50:00.707: Theme directory 48@2x/emblems of theme Eleven-Dark has no size field

(rnote:4722): Adwaita-WARNING **: 14:50:00.886: Using GtkSettings:gtk-application-prefer-dark-theme with libadwaita is unsupported. Please use AdwStyleManager:color-scheme instead.

(rnote:4722): Gtk-WARNING **: 14:50:01.272: Theme directory 32/emblems of theme Eleven-Dark has no size field

(rnote:4722): Gtk-WARNING **: 14:50:01.275: Theme directory 32@2x/emblems of theme Eleven-Dark has no size field

(rnote:4722): Gtk-WARNING **: 14:50:01.278: Theme directory 48/emblems of theme Eleven-Dark has no size field

[...]

PR #1824, opened 2026-08-18
Fix stylus pointer jump 

alberto-is:
Fixes #1440

Related to #665 and #534

GTK can deliver mouse motion events to the canvas while a stylus is hovering or drawing. Those events were processed using the current stylus state, which could move the pointer to the mouse position and cause unwanted strokes, erasures, pans, or selections, zoom, etc.

Regarding the related issues: Both issues appear to be caused by the same underlying problem, so this PR should resolve them as well. I believe this is the case because keeping the cursor outside the canvas prevents the issue, and the lines shown in their screenshots appear to originate from the same location, which seems to be the mouse pointer position.

**Changes**
- Prevent mouse coordinates from being processed with the stylus state, which caused unwanted strokes, erasures, pans, and selections.
    - Track whether a stylus input session is active in the canvas pointer controller.
    - Ignore non-stylus pointer events while that session is active.
    
- Recover mouse input when GTK does not report the stylus leaving proximity. Without this recovery, the stylus remains active and a mouse click can move the canvas instead of drawing with the mouse.
    - Allow a deliberate mouse click to end a stale stylus session and process the click normally..

The following videos demonstrate the behavior before and after the fix:

- Before the fix:

https://github.com/user-attachments/assets/1dd3e47c-d834-4c0f-8947-428aa7d19533

- After the fix:

https://github.com/user-attachments/assets/df0512b9-74c0-4778-89ca-db154f37bb4b

PR #1826, opened 2026-08-20
fix: ignore emulated pointer events while stylus in proximity

samu4199:
Fixes the stylus pointer jump (#1440) and frozen pen panning in the margin.
On Wayland the compositor keeps the mouse position in sync with the pen, delivering mouse motion events while a stylus is hovering or drawing. These were processed using the stylus state, causing the pen position to jump to the mouse position, stray strokes/erasures, and pen panning in the margin freezing after a few
millimeters.

Changes:
- Track stylus proximity in the canvas pointer controller
- Stop non-stylus motion events from reaching the canvas and the pan
  gestures while a stylus is in proximity
- A deliberate mouse click hands control back to the mouse, even if the
  compositor never reports the stylus leaving proximity

### DIFF
--- a/crates/rnote-ui/src/canvas/input.rs
+++ b/crates/rnote-ui/src/canvas/input.rs
@@ -13,12 +13,13 @@ use std::collections::HashSet;
 use std::time::{Duration, Instant};
 use tracing::trace;
 
-// Returns whether the event should be inhibited from propagating, and the new pen state
+// Returns whether the event should be inhibited from propagating, the new pen state and whether a stylus is in proximity
 pub(crate) fn handle_pointer_controller_event(
     canvas: &RnCanvas,
     event: &gdk::Event,
     mut pen_state: PenState,
-) -> (glib::Propagation, PenState) {
+    mut stylus_in_proximity: bool,
+) -> (glib::Propagation, PenState, bool) {
     let now = Instant::now();
     let mut widget_flags = WidgetFlags::default();
     let touch_drawing = canvas.touch_drawing();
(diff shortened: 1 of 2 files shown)


alberto-is:
Hi! Just noticed that this PR seems to address the same underlying issue as my PR #1824, which was opened a few days ago.

The approaches also look quite similar, especially regarding ignoring non-stylus/emulated pointer events while the stylus is active and handing control back to the mouse on click.

Tagging this here so the maintainers are aware of both implementations and can decide which approach they prefer.

crates/rnote-ui/src/canvas/input.rs
```diff
@@ -13,26 +13,61 @@ use std::collections::HashSet;
...
+            // while the stylus is in proximity. Otherwise mouse motion events (or a
+            // tablet's emulated core pointer) are fed into the current pen with the
+            // mouse position, making the eraser/pen flicker between the two positions
+            // or draw stray straight lines (e.g. on Wayland where the compositor keeps
+            // the mouse position in sync with the pen). Stopping them also prevents a
+            // buttonless mouse motion from resetting the scroller's pan gesture mid-drag
+            // (gtkgesturesingle.c:208-213), which would freeze pen panning in the margin.
+            gdk::EventType::MotionNotify => {
+                return (glib::Propagation::Stop, pen_state, stylus_in_proximity);
+            }
```
alberto-is, line 68: Could you provide a short video or GIF showing the issue you're describing? I’m having trouble reproducing the behavior that should occur without this `MotionNotify` handling, so it would help me understand exactly what you’re seeing

flxzt:
since https://github.com/flxzt/rnote/pull/1824 is now merged and addressed the same issue, I think this can be closed

Issue #327, opened 2022-10-27
Text/Math Recognition


OmegaRogue:
**Is your feature request related to a problem? Please describe.**
Taking handwritten notes is useful for quickly writing down things with sketches or math without the need for a keyboard, as a replacement for taking notes on Paper, but having digital Text is useful for searching through it and other use cases.
**Describe the solution you'd like**
Having the option to select handwritten text and convert it to a text box, and the option to select formulas other math to convert it to a formatted (maybe LaTeX) math element, like Xournal++ supports with LaTeX integration.

**Additional context**
OneNote supports OCR for mathematical formulas and equations

PhilDevProg:
For text recognition, [Tesseract](https://github.com/tesseract-ocr/tesseract) could be used. Here are the Rust bindings: https://github.com/antimatter15/tesseract-rs
I also found this: https://pyimagesearch.com/2020/08/24/ocr-handwriting-recognition-with-opencv-keras-and-tensorflow/ which describes how to do handwritten text recognition by using OpenCV, Keras & TensorFlow using Python but the OCR tool would have to be trained by the user (but this would propably result in better recognition).

ashrude:
The process described in there would technically work, but since we'd have access to the stroke information, it could be implemented much better. Having stroke information allows for higher accuracy.

Doublonmousse:
For the math conversion, this looks more like Task 1 of the CROHME competition https://crohme2023.ltu-ai.dev/data-tools/
This uses inkml as the file format for the stroke data and uses the stroke information ((x,y) position only).

Ideally, this would work in an incremental fashion and update the prediction for each new stroke rather than (or in addition to) only be accessible as a tool that convert a stroke selection to the math equivalent.

Conversion of a selection is only good the conversion is very good. If not you end up trying it a couple time, see that it fails to convert moderately complex equations and end up not using it (all the more so if there is no easy way to modify the converted equation afterward).

 Though #788 would have to be completed before this can be tackled

PellelNitram:
Hi! I’m working on [Xournal++ HTR](https://github.com/PellelNitram/xournalpp_htr), a local [handwritten text recognition](https://en.wikipedia.org/wiki/Handwriting_recognition) plugin for Xournal++. I figured commenting here might help with this issue.

[...]

Issue #224, opened 2022-07-17
[Feature request]: Textsearch Function (with highlighting)


TeiVDK:
**Is your feature request related to a problem? Please describe.**
With the new added text feature in the BETA, Rnote became ever-closer to completely replacing all my note taking apps. However, as is usual when one takes keyboard-written notes, the need to quickly search for a word or sentence often comes up. The app, in it's current form, does not offer such an option. 

**Describe the solution you'd like**
A search function like in the new Gnome Text Editor or Gedit that allows to quickly find a word while highlighting it.

I mention highlighting because some apps have a poor implementation of the highlighting that makes it not stand out as much when a lot of text is on the screen.

A "simple" search would suffice :) (I know implementing any feature can be complicated and I thank anyone that uses his/her free time to implement it.) 

**Describe alternatives you've considered**
The only "alternative" to a search function would be to manually look for a word or sentence.

mityax:
This would play nicely together with text recognition (e.g. #327) – it would be pretty awesome to have all your notes searchable and the feature combination would make this app definitely stand out amongst the other options.

Text-reognition doesn't necessarily mean transforming the hand-written text into machine text in my eyes. It could also just mean storing the recognized text as meta data or so and then making it search- (and highlight-) able.

PellelNitram:
Hi! This might be relevant here as well: I'm working on [a plugin](https://github.com/PellelNitram/xournalpp_htr) for local [HTR (Handwritten Text Recognition)](https://en.wikipedia.org/wiki/Handwriting_recognition) that will allow features like the search asked for here.

I wrote up the details and posted a [demo video](https://www.youtube.com/watch?v=rwHbacjjAS8) in [this issue](https://github.com/flxzt/rnote/issues/327#issuecomment-3560771585). We are looking into extending the plugin to support `rnote` files, so please let us know if you'd be interested in that integration!

Cheers! :-)

CreativeCoder0:
[...]

PR #1828, opened 2026-08-22
WIP implementation to get to get handwriting recognition

Verantor:
I am working to get handwriting recognition to work with rnote. Its a custom CNN + Conformer approach. Right now it uses greedy decode so it has no language understanding, but I will try to get an N-gram language model to further improve recognition. Right now I just insert text over the handwriting.

This is the reason for this Pull Request because I was wondering on how you expect Handwriting recognition to work for a notes app like searching and maby spelling correction.

If you have suggestion on how to implement those features or what i did wrong feel free to complain and help!


alberto-is:
From your message, it seems that your objective is to implement spelling corrections. That’s a bit more complicated. However, this is related to #224 and, above all, to #327.

The former was about search, and if you implement this, it could also be used for search. However, it is more closely related to the latter. In any case, you will find useful information in the comments, and you can also check this repository: [xournalpp_htr](https://github.com/PellelNitram/xournalpp_htr), which is the one mentioned in those issues.

Also, maybe it would be better to create a subproject for this? It seems that several iterations will be necessary, and it should probably be optional due to the hardware requirements.

Verantor:
So thanks for showing me the issues and the xournalpp implementation but right now I believe in my implementation of hw recognition. I don't believe that the hardware requirements are too big, but I guess I can hide it behind a feature flag or a setting. 

What do you mean by subproject? Another create?

alberto-is:
By “subproject,” I was actually thinking of a separate repository included as a Git submodule, rather than another crate inside the main project. Since I assume this feature may require several iterations and experiments as it evolves, so having it developed somewhat independently could make that process easier. But if you think it fits well directly into the main project, I don't have a strong opinion about it.

Regarding the complexity, I was also thinking about language support. From what I understand, the current model has a fixed vocabulary and mainly supports English and German. For example, it does not appear to support characters such as ñ, accented characters like á or é, or non-Latin scripts such as Chinese.

So, if the goal is to add spelling correction, I think it would also be important to define which languages are supposed to be supported and how that should interact with the recognition model. That was mostly why I said that the correction part seemed more complicated to me.

Verantor:
So as a quick update on how it looks right now

[How it looks right now](https://github.com/user-attachments/assets/679f89d6-ea9f-45ad-8b5b-8c74316501a4)

Also if someone wants to contribute sentences for the model to learn on tell me.

PR #1831, opened 2026-08-24
feat(installer): allow a per-user install mode on Windows

gdegrange:
Hello ! 

The Windows installer currently requires administrative privileges, which makes Rnote impossible to install on managed machines. PrivilegesRequired defaults to admin and no override is allowed, so /CURRENTUSER has no effect either.

This PR adds one directive, PrivilegesRequiredOverridesAllowed=dialog, which makes Setup show the standard "Select Setup Install Mode" dialog and enables the /ALLUSERS and /CURRENTUSER parameters. PrivilegesRequired is left at its default of admin, so "Install for all users" stays the default and nothing changes for existing users.

No other change is needed: the script already uses {autopf}, {autofonts}, {autoprograms}, {autodesktop} and Root: HKA, which resolve to their per-user equivalents in non administrative install mode. The rnote-cli PATH code already writes to HKEY_CURRENT_USER.

One caveat: per-user FontInstall requires Windows 10 1803 or newer, so the four bundled fonts would be skipped on older versions in per-user mode.

### DIFF
--- a/build-aux/rnote_inno.iss.in
+++ b/build-aux/rnote_inno.iss.in
@@ -33,8 +33,10 @@ OutputDir={#meson_build_root}
 OutputBaseFilename=@INSTALLER_NAME@
 SetupIconFile={#meson_source_root}\crates\rnote-ui\data\icons\rnote.ico
 UninstallDisplayIcon = {app}\bin\{#MyAppExeName}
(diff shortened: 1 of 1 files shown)


flxzt:
Seems like a good change. I'll give it a try and if it works without problems it will be merged.

flxzt:
works well, thanks!

PR #1835, opened 2026-08-26
fix: scale each imported pdf page independently to fit the document format

### LINKED ISSUE #1834: Importing pdf with different page formats breaks "adjust document" option
**Describe the bug**  
If you have a pdf file with different page formats, when importing you'll get misalignment of pages.

**To Reproduce**  
1. Get a pdf with Vertical and Horizontal pages (scanning from an app on your phone for example).
2. Import with "Adjust document" enabled (i'm using a spanish-translated version).

**Expected behavior**  
The document should adjust format "page-per-page" so that each page gets the same size/format from the original.

**Screenshots**  
 
Image
Here i've used CamScanner for some university stuff. Page borders are in dark blue. You can see how the second scanned-page is horizontal, and the third scanned-page is starting in the middle of the second document-page, cropping it at the middle when trying to export afterwards.

 - OS: Windows 11 Home 25H2
 - App Version: Rnote v0.14.2
 - Installation Source: GitHub repo
 - Desktop Environment: *.
 - Display Server: *.
 - Input Source: Mouse and Keyboard.

*: I have no idea. I'll edit if you help me get this information.

**Additional context**  
If there's currently a way to configure the import so that my expected behaviour is actually executed, then i apologise and ask for you to explain this. In that case, i'll come up with another issue explaining how UNintuitive the import options are, since i've tried many option combinations already and this seems to be the closest to what i want.

alberto-is:
Fixes #1834

When importing a PDF containing pages with different formats, such as a mix of portrait and landscape pages, the pages could become misaligned when **Adjust document** was enabled. 

This happened because all pages were scaled using the dimensions of the first page. Pages with a different format therefore did not fit correctly in the document layout.

**Changes**

- Scale each imported PDF page independently when **Adjust document** is enabled.
- Keep each page's aspect ratio while fitting it within the document format.
- Preserve the existing import behavior when **Adjust document** is disabled.
- Apply the fix to both bitmap and vector PDF imports.

**Testing**

- `cargo fmt --all -- --check`
- `cargo clippy --workspace --all-targets -- -D warnings`
- `cargo test --workspace`

The following video shows the result before and after the fix when importing a PDF with mixed page formats:

https://github.com/user-attachments/assets/9871debf-82d4-4c5b-8cc7-26482244bc84

### DIFF
--- a/crates/rnote-engine/src/strokes/bitmapimage.rs
+++ b/crates/rnote-engine/src/strokes/bitmapimage.rs
@@ -152,12 +152,24 @@ impl BitmapImage {
             format.width() * (pdf_import_prefs.page_width_perc / 100.0)
         };
 
-        // calculate the page zoom based on the width of the first page.
-        let page_zoom = if let Some(first_page) = pages.first() {
-            page_width / first_page.render_dimensions().0 as f64
-        } else {
+        // Uniform zoom based on the widest page, keeps relative page sizes.
+        let max_intrinsic_width = page_range
+            .clone()
+            .filter_map(|page_i| pages.get(page_i))
+            .map(|page| page.render_dimensions().0 as f64)
+            .fold(0.0f64, f64::max);
+        if max_intrinsic_width <= 0.0 {
             return Ok(vec![]);
-        };
+        }
+        let page_zoom_fit = page_width / max_intrinsic_width;
+
+        // Stride for OnePerDocumentPage: start each page on a document page boundary.
+        let max_rendered_height = page_range
+            .clone()
+            .filter_map(|page_i| pages.get(page_i))
+            .map(|page| page.render_dimensions().1 as f64 * page_zoom_fit)
+            .fold(0.0f64, f64::max);
+
         let x = insert_pos[0];
         let mut y = insert_pos[1];
 
@@ -171,11 +183,11 @@ impl BitmapImage {
                     let dimensions = page.render_dimensions();
                     (dimensions.0 as f64, dimensions.1 as f64)
                 };
-                let width = intrinsic_width * page_zoom;
-                let height = intrinsic_height * page_zoom;
+                let width = intrinsic_width * page_zoom_fit;
+                let height = intrinsic_height * page_zoom_fit;
                 let render_settings = hayro::RenderSettings {
-                    x_scale: (pdf_import_prefs.bitmap_scalefactor * page_zoom) as f32,
-                    y_scale: (pdf_import_prefs.bitmap_scalefactor * page_zoom) as f32,
+                    x_scale: (pdf_import_prefs.bitmap_scalefactor * page_zoom_fit) as f32,
+                    y_scale: (pdf_import_prefs.bitmap_scalefactor * page_zoom_fit) as f32,
                     width: Some((pdf_import_prefs.bitmap_scalefactor * width).ceil() as u16),
                     height: Some((pdf_import_prefs.bitmap_scalefactor * height).ceil() as u16),
                     bg_color: vello_cpu::color::AlphaColor::WHITE,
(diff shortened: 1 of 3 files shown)


alberto-is:
This is a design decision I'd like to flag before merging.

The min() in page_zoom_fit gives the page a "contain" behavior: each page is scaled to fit entirely within the document page format while preserving its aspect ratio. The trade-off is that when a page's aspect ratio doesn't match the format, there can be empty space on one side. For example, a portrait page imported into a landscape-oriented format may leave blank space to the right.

An alternative would be to scale by width only (page_width / intrinsic_width), which would always fill the full width. The height would then match the page's natural size, and since the document is already expanded to fit the content via set_height, nothing would be clipped.

I kept the min() approach because it ensures the page is never distorted or cropped. However, if it's  preferred to always fill the full width, it's a one-line change and I'm happy to update it.

Any preference on which behavior we should go with?

Doublonmousse:
> The min() in page_zoom_fit gives the page a "contain" behavior: each page is scaled to fit entirely within the document page format while preserving its aspect ratio. The trade-off is that when a page's aspect ratio doesn't match the format, there can be empty space on one side. For example, a portrait page imported into a landscape-oriented format may leave blank space to the right.

This way you guarantee that a pdf page is always inside a rnote page on export. Wouldn't you cut your page (portrait page into a landscape pdf document) into two with the second option?

Your video shows the second option if I'm not mistaken

alberto-is:
Hi, the video shows the first option, but in this case the format matches ( I found the problem because I tried it later with other PDFs). The thing is, the main problem is divided into two: the min to contain sometimes leaves the white space, and the scale only takes the first page into consideration. I corrected the latter because the former is more like a "feature result", since the pages are only cut when we merge formats (but that was usual too before this PR):

image

If we want to avoid cutting the pages, we would need to check whether each page fits in the remaining space of the current layout, taking into consideration the pages that have already been imported. The downside is that this could leave some white space between pages. For example, in this case, with the first option, the page gets cut:

image
image

We have a horizontal page that only occupies half of the layout, so the next page goes into the other half. Since it doesn't fit in the remaining space, it gets cut. So, maybe having some white space between pages would be better?

As I said, this happens because we calculate whether the page fits in the format, but we don't take into account that the format has already been partially occupied. The thing is that each case is more like a "personal preference", right? For me, for example, I prefer the pages to be cut rather than having white spaces between them, and I don't really mind a "small" gap on the right. But I can implement the second solution, or even add the calculations to leave white space when needed to avoid cutting the pages (I don't think both at the same time are possible because, in any case, one page could occupy several pages, with one page and a bit being the most common case)  if it is more aligned with your vision.

flxzt:
I wonder if it is even desirable to always fit the document pages uniformly into the Rnote document width.

If a PDF has multiple size pages (e.g. A4 and A3), should that not be reflected in the Rnote document as well?

Another idea for a solution would be to take the maximum width of all PDF pages and scale them to this value accordingly.

alberto-is:
I can update the PR to use a single zoom level for all pages, based on the maximum intrinsic width, so that the widest PDF page fits `page_width` while the others keep their relative size. I assume that in this case, we don't need to consider it a problem if a very tall page exceeds `format.height()` and gets split on export, since that was already the previous behavior.

Btw, with this approach, narrower pages will leave empty space on the right if left-aligned (as is the current behavior). Do you prefer to keep them left-aligned, or center them?

flxzt:
Good question, I'll leave it up to you.
Btw. is this approach what makes sense to you as well or would you prefer your initial version?

alberto-is:
So, I went with the max-width approach;  it’s much better than my initial idea. I updated the PR to use a single zoom level based on the widest page and dropped the per-page fit from the initial version.

Since I personally prefer keeping all the pages together because I don’t use the export that much, I also enabled “Page Spacing” with “Adjust Document”, so “One per Document Page” can avoid splitting pages on export for those who use it. I also centered the narrower pages, but I can left-align them instead if you prefer.

Btw (x3), I applied the zoom changes to both modes (Adjust Document on/off) because it didn’t make much sense to me to keep the zoom level based on the first page.

crates/rnote-engine/src/strokes/bitmapimage.rs
```diff
@@ -152,12 +152,24 @@ impl BitmapImage {
...
-        let page_zoom = if let Some(first_page) = pages.first() {
-            page_width / first_page.render_dimensions().0 as f64
-        } else {
+        // Uniform zoom based on the widest page, keeps relative page sizes.
+        let max_intrinsic_width = page_range
+            .clone()
+            .filter_map(|page_i| pages.get(page_i))
+            .map(|page| page.render_dimensions().0 as f64)
+            .fold(0.0f64, f64::max);
+        if max_intrinsic_width <= 0.0 {
```
flxzt, line 161: Should an error be reported in that (unlikely) case?
Because the condition is slightly different than before:
- if no pages are present
- new: if pages are present, but have zero width.

crates/rnote-engine/src/strokes/vectorimage.rs
```diff
@@ -234,12 +234,24 @@ impl VectorImage {
...
-            page_width / first_page.render_dimensions().0 as f64
-        } else {
+        // Uniform zoom based on the widest page, keeps relative page sizes.
+        let max_intrinsic_width = page_range
+            .clone()
+            .filter_map(|page_i| pages.get(page_i))
+            .map(|page| page.render_dimensions().0 as f64)
+            .fold(0.0f64, f64::max);
+        if max_intrinsic_width <= 0.0 {
             return Ok(vec![]);
```
flxzt, line 244: see: https://github.com/flxzt/rnote/pull/1835/changes#r4064349123

PR #1845, opened 2026-09-07
fix: premultiply alpha when loading raster images

sci42:
Fixes the transparency issue with imported images reported in #1843, where PNGs containing transparent areas were rendered as a solid-colored block.

## The Bug

A PNG can still store color data in pixels that are fully transparent. This is common in images where the background has been removed; the alpha channel is set to 0, but the underlying color remains. While other programs hide this data, Rnote displayed it as a solid-colored area.

The cause is an incorrect label. Data from the `image` crate uses "straight alpha" but is marked as "premultiplied." Consequently, the renderer assumes the color has already been blended with the alpha and skips that step, causing the hidden color to appear unfiltered in the image.

## The Solution

Upon loading, the alpha is now factored in once so that the label is correct. During export, `into_imgbuf` performs the reverse operation—reverting the calculation—because the `image` crate expects straight alpha. This addresses images from Cairo and SVG sources that are actually premultiplied, where soft edges previously appeared too dark upon export.

This maintains a single internal format, and no changes are made to the `.rnote` file format.

## Limitation

Documents already saved contain the buffer with the incorrect label and will continue to display the solid-colored area. Only newly imported images will be correct; re-importing the image fixes the issue in an affected document.

*Disclosure: I used AI tooling to analyse the problem and to draft the implementation and its tests. I verified the diagnosis against the reported file, ran the tests myself, and wrote this description and my replies on my own.*

### DIFF
--- a/crates/rnote-engine/src/image.rs
+++ b/crates/rnote-engine/src/image.rs
@@ -120,7 +120,11 @@ impl From<image::DynamicImage> for Image {
         let pixel_width = dynamic_image.width();
         let pixel_height = dynamic_image.height();
         let memory_format = ImageMemoryFormat::R8g8b8a8Premultiplied;
-        let data = glib::Bytes::from_owned(dynamic_image.into_rgba8().to_vec());
+        // the image crate decodes to straight alpha, so it needs to be premultiplied
+        // to match the memory format.
+        let data = glib::Bytes::from_owned(convert_image_straight_to_premultiplied(
+            dynamic_image.into_rgba8().into_vec(),
+        ));
         let bounds = Aabb::new(
             Vector2::ZERO,
             Vector2::new(pixel_width as f64, pixel_height as f64),
@@ -224,10 +228,11 @@ impl Image {
         self.assert_valid()?;
 
         match self.memory_format {
+            // the image crate expects straight alpha, so the premultiplication needs to be undone.
             ImageMemoryFormat::R8g8b8a8Premultiplied => image::RgbaImage::from_vec(
                 self.pixel_width,
                 self.pixel_height,
-                self.data.to_vec(),
+                convert_image_premultiplied_to_straight(self.data.to_vec()),
             )
             .ok_or_else(|| {
                 anyhow::anyhow!(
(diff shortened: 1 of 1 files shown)


flxzt:
Maybe you can ask AI to check whether this conversion is possible inside `DynamicImage`/with the image crate. I am skeptical about fiddling with the bytes ourselves.

sci42:
Fair point — I went looking instead of guessing. Summary of what I checked and what that leaves us with.

**The `image` crate can't do it.** In 0.25.10 there are three occurrences of "premultipl" in the entire source: two doc comments and a private f32 step inside `Blend::blend`. No public API on `DynamicImage`, `RgbaImage` or `Rgba`.

**But the loading side doesn't need any conversion at all.** Both backends already accept straight alpha:

- `piet::ImageFormat::RgbaSeparate` — piet-cairo premultiplies itself in `make_image`
- `gdk::MemoryFormat::R8g8b8a8`

So an `ImageMemoryFormat::R8g8b8a8` variant would let decoded images stay straight and simply be labelled correctly, and `into_imgbuf()` could then pass those bytes to the `image` crate untouched. That removes the byte handling for imported raster images entirely — which is the path the reported bug is on.

One consequence worth deciding deliberately: `memory_format` is serialised into the `.rnote` file, so a new variant makes new files unreadable for older versions.

**The other direction can't be delegated.** Cairo genuinely produces premultiplied data (`ARgb32` is its only option), so exporting cairo/SVG-sourced images through the `image` crate needs un-premultiplication somewhere. What I found:

- `cairo-rs`: no such API (zero mentions)
- `GdkTextureDownloader` can convert, but it is behind the `ui` feature and `rnote-cli` exports too
- `palette` (already a workspace dependency) has `Premultiply`/`unpremultiply`, and `color` (transitive, via vello) has `PremulColor::un_premultiply()` — both are float-based, so u8 → f32 → u8 per pixel, with different rounding

For that one spot it is therefore either a float round trip or integer arithmetic. For what it's worth, piet-cairo does the same thing by hand — `fn premul` inside `make_image`.

**Options, your call:**

1. Straight-alpha variant, and keep a small integer helper only for the cairo/SVG export path. Least code, no new dependency, but the file-format consequence above.
2. Straight-alpha variant plus `palette` for the export path. No hand-written arithmetic, at the cost of a float round trip per pixel and `palette` added to `rnote-engine`.
3. This PR as it stands: both conversions explicit, nothing about the stored format changes.

I am happy to rebuild it either way — 1 would be my preference, 3 is the smallest step for the user-visible bug. Just say which and I'll push it.

flxzt:
I refuse to review when my feedback is only forwarded to a prompt.

I am not against this change per-se, but I want to be able to discuss with a human.

If wanted, re-open with human generated description and responses where I know someone understands the change that is proposed.

sci42:
Hi Felix Zwettler,
I am sorry for sending you an AI-generated response. I initially analyzed the problem myself, but when you asked for a much more detailed investigation, I simply presented what I considered to be the best solution. I realize it might look as though I just passed everything along, but unfortunately, I couldn't fully grasp the issue in such a short time; since you had suggested I ask my AI, I assumed that was acceptable.

Your messages were never used directly as a prompt, and I certainly didn't intend to make a bad impression by presenting the results of my best breakdown. I apologize for my actions and will try from now on to use AI less and strictly as an analysis tool.

sci42:
I’ve taken another close look at the issue and the surrounding code. You introduced a standardized image format; the enum is marked `#[non_exhaustive]`, and `TryFrom<gdk::MemoryFormat>` explicitly rejects any other format.

The problem is that, during loading, the data is tagged incorrectly. Data from the `image-crate` uses straight alpha but is marked as premultiplied. Consequently, the renderer assumes the color has already been combined with the alpha and skips that step. This causes the color underlying fully transparent pixels to appear unfiltered in the final image.

I also considered what adding a second format would entail. That would create more problems because the `memory_format` field is written to the `.rnote` file. Older versions of Rnote wouldn't be able to open new files—without a clear error message, since the version-handling logic routes everything from version 0.15.0 onwards into the same code path. Furthermore, it would only defer the problem, as Cairo requires premultiplied values; this would necessitate a conversion every time—specifically, during every render pass.

That’s why I’d prefer to keep the approach described in my PR: convert once upon loading and convert back upon export. This preserves the internal format and leaves the `.rnote` file unchanged.

Does that work for you?

PR #1852, opened 2026-09-12
feat: Add searchpanel 

Verantor:
### Description

This PR implements a search feature within the sidebar. It uses fuzzy search to find text objects, and the results are displayed directly in the sidebar.

### Features

- Search Panel: A new dedicated search panel located in the sidebar.
- Shortcuts: Added the `Ctrl + F` shortcut to quickly open the search.
- Camera Navigation: Clicking on a search result automatically moves the camera to focus on it.


Verantor:
I just noticed that at certain sizes the panel selector bar thingy disappears. I guess I f'ed something up, I try to fix it tomorrow.

Edit:
This only happens when using the minimize and maximize buttons. I can't think of a proper fix.

